PR #2, opened 2021-08-02
feat: include the new setupProxy.js

leamendoza97:
# Whats?
Add the setupProxy.js to the package and fix the context that rollup build.

# Why?
The setupProxy.js is a proxy loaded by the CLI which allows mapping different endpoints.
See more https://github.com/decentraland/sdk/issues/12
https://github.com/decentraland/cli/pull/569

### DIFF
--- a/packages/decentraland-ecs/src/setupProxy.js
+++ b/packages/decentraland-ecs/src/setupProxy.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 module.exports = function (dcl, app, express) {
     const dclKernelPath = path.dirname(require.resolve('decentraland-kernel/package.json', {paths: [dcl.getWorkingDir()]}));
     const dclKernelDefaultProfilePath = path.resolve(dclKernelPath, 'default-profile');
+    const dclKernelImagesDecentralandConnect = path.resolve(dclKernelPath, 'images', 'decentraland-connect');
     const dclUnityRenderer = path.resolve(dclKernelPath, 'unity-renderer');
 
     const routeMappingPath = {
@@ -25,7 +26,8 @@ module.exports = function (dcl, app, express) {
         });
     }
     
-    createStaticRoutes(app, '/@/artifacts/unity-renderer/*', dclUnityRenderer)
+    createStaticRoutes(app, '/@/artifacts/unity-renderer/*', dclUnityRenderer);
+    createStaticRoutes(app, '/images/decentraland-connect/*', dclKernelImagesDecentralandConnect);
 
     app.use('/default-profile/', express.static(dclKernelDefaultProfilePath));
 };
--- a/packages/decentraland-ecs/src/setupProxy.js
+++ b/packages/decentraland-ecs/src/setupProxy.js
@@ -2,35 +2,40 @@ const path = require('path')
 const fs = require('fs')
 
 module.exports = function (dcl, app, express) {
-    const dclKernelPath = path.dirname(require.resolve('decentraland-kernel/package.json', {paths: [dcl.getWorkingDir()]}));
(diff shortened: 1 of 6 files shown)


packages/decentraland-ecs/src/setupProxy.js
```diff
@@ -0,0 +1,45 @@
...
+            type: 'text/plain'
+        },
+        '/@/artifacts/unity-renderer/unity.framework.js.unityweb':{
+            path: path.resolve(dclUnityRenderer, 'unity.framework.js.unityweb'), 
+            type: 'text/javascript'
+        },
+        '/@/artifacts/unity-renderer/unity.wasm.unityweb':{
+            path: path.resolve(dclUnityRenderer, 'unity.wasm.unityweb'), 
+            type: 'application/wasm'
+        },
```
menduz, line 33: May it make sense to use wildcards? `/@/artifacts/unity-renderer/*` files change across versions, being too specific may not be a wise future-proof desicion.

Example code: https://github.com/decentraland/explorer-website/blob/main/src/setupProxy.js#L13

leamendoza97: Yes, I'll change it. In the previous commit there wasn't 1-to-1 mapping but now makes sense.

packages/decentraland-ecs/src/setupProxy.js
```diff
@@ -0,0 +1,45 @@
+const path = require('path')
+const fs = require('fs')
+
+module.exports = function (dcl, app, express) {
+    const dclKernelPath = path.resolve(dcl.getWorkingDir(), 'node_modules', 'decentraland-kernel');
```
menduz, line 5: Use `require.resolve` to recursively find the path using node_modules resolutor embedded in Node.js
```suggestion
    const dclKernelPath = require.resolve('decentraland-kernel', {paths: [dcl.getWorkingDir()]})
```

packages/decentraland-ecs/src/setupProxy.js
```diff
@@ -0,0 +1,45 @@
...
+    const dclKernelDefaultProfilePath = path.resolve(dclKernelPath, 'default-profile');
+    const dclUnityRenderer = path.resolve(dclKernelPath, 'unity-renderer');
```
menduz, line 7: Same as `decentraland-kernel`

menduz (changes requested):
Awesome, I have some suggestions in the comments

packages/decentraland-ecs/src/setupProxy.js
leamendoza97: That doesn't resolve the directory instead it resolves the module file if exists (not the decentraland-kernel case). Maybe, `const dclKernelPath = path.dirname(require.resolve('decentraland-kernel/package.json', {paths: [dcl.getWorkingDir()]}))` ?

menduz: RIGHT, your suggestion is correct :shipit:

packages/decentraland-ecs/src/setupProxy.js
leamendoza97: I only changed the dclKernelPath because the others are relative to it. If decentraland-kernel is resolved, can we assume the others exist?

menduz: I think this one should be assumed to exist also, resolved the same way for consistency
```suggestion
    const dclUnityRenderer = path.dirname(require.resolve('@dcl/unity-renderer/package.json', {paths: [dcl.getWorkingDir()]}));
```

menduz (approved):

PR #8, opened 2021-08-17
chore: change s3 secrets and comment an installable package to test

leamendoza97:
# Whats ?
Add a step to comment the installable package with an URL, also change the S3 secrets for CI.

# Why?
Make PR and branches testing easier

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,12 +63,12 @@ jobs:
           deterministic-snapshot: true
 
           ## publish every package to s3
-          s3-bucket: ${{ secrets.CDN_BUCKET }}
+          s3-bucket: ${{ secrets.SDK_TEAM_S3_BUCKET }}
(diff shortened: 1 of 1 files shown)


menduz:
Deterministic snapshots are required for the non-s3 deployment, can we get it working without them?

leamendoza97:
> Deterministic snapshots are required for the non-s3 deployment, can we get it working without them?

Sure, but **runId** is lower number than generated date, so for **next** version we'll publish in **@ci** instead of **@next**.
This is because https://github.com/menduz/oddish-action/blob/master/oddish.ts#L414
See https://github.com/decentraland/js-sdk-toolchain/runs/3352015861#step:17:193

menduz:
aren't we breaking something bigger then? I prefer to use `ci` for a couple of days, the deterministic versions are important to install from NPM because the publish has a different version than the one present in the package.json. And the final versions of the ecs/package.json won't even exist

menduz (approved):

PR #13, opened 2021-08-31
fix: recursive export, now it ignores the files specified in .dclignore

leamendoza97:
# Whats ?
Now it reads the .dclignore file to exclude the paths specified there from `dcl export` and `dcl start`.
Also, it changes the base64 hash to sha1 in export to no expose the path where the export is done.

# Why?
As of the preview and export were added, the .dclignore hasn't been considered to this jobs, so the export can be recursively self-added and in the same way preview include export dir. This causes that the export directory non stop increasement

### DIFF
--- a/packages/decentraland-ecs/src/setupUtils.ts
+++ b/packages/decentraland-ecs/src/setupUtils.ts
@@ -106,6 +106,8 @@ export function getSceneJson({
     let ignoreFileContent = ''
     if (fs.existsSync(dclIgnorePath)) {
       ignoreFileContent = fs.readFileSync(path.resolve(folder, '.dclignore'), 'utf-8')
+    } else {
+      ignoreFileContent = defaultDclIgnore()
     }
 
     return entityV3FromFolder({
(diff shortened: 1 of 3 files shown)


packages/decentraland-ecs/src/setupUtils.ts
```diff
@@ -42,33 +55,66 @@ export function entityV3FromFolder({ folder, addOriginalPath }: { folder: string
         }
         const _folder = folder.replace(/\\/gi, '/')
         const key = file.replace(_folder, '').replace(/^\/+/, '')
+        return key
+      })
+      .filter(($) => !!$) as string[]
+
+    const ig = ignore().add(ignorePattern || '')
```
pablitar, line 62: Shouldn't this also include node_modules and .git? Or that should come from .dclignore?

pablitar: Maybe consider that as default if no .dclignore file is provided or is empty?

pablitar (approved):
LGTM!

packages/decentraland-ecs/src/setupUtils.ts
leamendoza97: Yep, it's a good idea. **.dclignore** is generated when you create a new project (https://github.dev/decentraland/cli/blob/a90a99e0060f82b864b14b283e4ba9552509488b/src/lib/Project.ts) but it can be deleted accidentally

PR #16, opened 2021-09-06
feat: add video event observable

leamendoza97:
# What?
Add video event observable and update posix, kernel and unity-renderer dependencies version.

# Why?
https://github.com/decentraland/sdk/issues/11

# How to test it?
1. Clone the example repository https://github.com/decentraland-scenes/video-streaming
2. Install PR auto-generated package
3. Add these lines to the end of **src/game.ts**
```
let flag = true
const testEntity = new Entity()
testEntity.addComponent(
    new utils.Interval(5000, () => {
        flag = !flag
        if (flag){
            videoTexture.play()
        }else{
            videoTexture.pause()
        }
        log(flag)
    })
)

engine.addEntity(testEntity)
onVideoEvent.add((data) => {
    log("VideoEvent :", videoTexture)
    log("onVideoEvent::data :", data)
})  
```


packages/decentraland-ecs/src/decentraland/Events.ts
```diff
@@ -121,6 +126,10 @@ export function _initEventObservables(dcl: DecentralandInterface) {
           onPlayerExpressionObservable.notifyObservers(event.data as IEvents['playerExpression'])
           return
         }
+        case 'videoEvent': {
+          onVideoEvent.notifyObservers(event.data as IEvents['videoEvent'])
+          return
+        }
```
menduz, line 132: I think we should look for the component by Id and reflect the changes in the component itself. We may keep the event tho.

menduz: That way the code may become:

```ts
const video = new VideoTexture()
entity.setComponent(video)
entity.setComponent(new PlaneShape)

class MySystem {
  update(){
    if (video.position > 10) {
      // do something
    }
  }
}

```

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -1084,6 +1094,15 @@ export class VideoTexture extends ObservableComponent {
   @ObservableComponent.field
   seek: number = -1
 
+  @ObservableComponent.readonly
```
menduz, line 1097:
```suggestion
```

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -1084,6 +1094,15 @@ export class VideoTexture extends ObservableComponent {
...
+  position: number = -1
+
+  @ObservableComponent.readonly
```
menduz, line 1100:
```suggestion
```

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -1084,6 +1094,15 @@ export class VideoTexture extends ObservableComponent {
...
+  videoLength: number = -1
+
+  @ObservableComponent.readonly
```
menduz, line 1103:
```suggestion
```

menduz (changes requested):
Non-sinchronizable fields should have no decorator. That is to prevent an endless loop of updates between the scene and the renderer.

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -1135,4 +1149,12 @@ export class VideoTexture extends ObservableComponent {
 
     return super.toJSON()
   }
+
+  update(videoEvent: IEvents['videoEvent']) {
+    if (videoEvent.videoClipId == this.videoClipId) {
+      this.data['status'] = (videoEvent.videoStatus as VideoStatus) || VideoStatus.NONE
+      this.data['videoLength'] = videoEvent.totalVideoLength
+      this.data['position'] = videoEvent.currentOffset
```
menduz, line 1157:
```suggestion
      this.status = (videoEvent.videoStatus as VideoStatus) || VideoStatus.NONE
      this.videoLength = videoEvent.totalVideoLength
      this.position = videoEvent.currentOffset
```

menduz: Sorry for missing this in the previous review, but `data` only works with the decorators

menduz: Something you can do is using getters like this:

```ts
class MyComponent {
  private _status: number = DEFAULT_STATUS

  get status(): number {
    return this._status
  }

  // use @internal to strip this declaration from the final .d.ts
  // @internal
  update({ status }) {
    this._status = status
  }
}

```

menduz (approved):

PR #32, opened 2021-10-05
fix: add mapping .br extension to the same path without it

leanmendoza:
# What ?
..

# Why?
When you access to preview mode in HTTPS mode, the unity-renderer fetch fails.

### DIFF
--- a/packages/decentraland-ecs/src/setupProxy.ts
+++ b/packages/decentraland-ecs/src/setupProxy.ts
@@ -61,12 +61,19 @@ const setupProxy = (dcl: any, app: express.Application) => {
   }
 
   createStaticRoutes(app, '/images/decentraland-connect/*', dclKernelImagesDecentralandConnect)
-  createStaticRoutes(app, '/@/artifacts/unity-renderer/*', dclUnityRenderer)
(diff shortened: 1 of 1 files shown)


packages/decentraland-ecs/src/setupProxy.ts
```diff
@@ -61,12 +61,19 @@ const setupProxy = (dcl: any, app: express.Application) => {
...
   createStaticRoutes(app, '/@/artifacts/loader/*', dclKernelLoaderPath)
   createStaticRoutes(app, '/default-profile/*', dclKernelDefaultProfilePath)
 }
 
-const createStaticRoutes = (app: express.Application, route: string, localFolder: string) => {
+const createStaticRoutes = (
+  app: express.Application,
+  route: string,
+  localFolder: string,
+  mapFile: ((filePath: string) => string) | null = null
```
gonpombo8, line 75: u can avoid that null default value making that arg optional with the `?` operator
```ts
mapFile?: (filePath: string) => string,
```

gonpombo8 (approved):

packages/decentraland-ecs/src/setupProxy.ts
leanmendoza: smart, ty

PR #35, opened 2021-10-14
feat: change self to globalThis

leanmendoza:


leanmendoza:
This breaks the preview mode with current kernel and maybe production too, so, I closed this for now.

PR #40, opened 2021-10-27
fix: UIInputText value update

pbosio:
fix the odd behavior where `onTextSubmit` and `onChanged` callbacks were triggered but `value` never changed accordingly

### DIFF
--- a/packages/decentraland-ecs/src/decentraland/UIShapes.ts
+++ b/packages/decentraland-ecs/src/decentraland/UIShapes.ts
@@ -356,7 +356,7 @@ export class UIInputText extends UIShape {
 
   // @internal
   @OnUUIDEvent.uuidEvent
-  onTextChanged: OnChanged
+  protected readonly onTextChanged: OnChanged
 
   @OnUUIDEvent.uuidEvent
   onFocus: OnFocus | null = null
--- a/packages/decentraland-ecs/src/decentraland/UIShapes.ts
+++ b/packages/decentraland-ecs/src/decentraland/UIShapes.ts
@@ -369,14 +369,19 @@ export class UIInputText extends UIShape {
     this.onTextChanged = new OnChanged((e) => {
       const { value, isSubmit } = e.value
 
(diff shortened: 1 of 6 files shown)


packages/decentraland-ecs/src/decentraland/UIShapes.ts
```diff
@@ -350,17 +350,36 @@ export class UIInputText extends UIShape {
...
+    super(parent)
+    this.onTextChanged = new OnChanged((e) => {
+      const { value, isSubmit } = e.value
+
+      const isDirty = this.dirty
+      this.value = value
+      this.dirty = isDirty
+
+      if (isSubmit && this.onTextSubmit) {
+        this.onTextSubmit.callback({ text: value } as IEvents['onTextSubmit'])
```
gonpombo8, line 377: instead of using the casting `as`, its a good practice to declare the const with the strict types.

```ts
const value: IEvents['onTextSubmit'] = { text: value }
```
This way you cant add extra or invalid props to the event

packages/decentraland-ecs/src/decentraland/UIShapes.ts
```diff
@@ -350,17 +350,36 @@ export class UIInputText extends UIShape {
...
   onBlur: OnBlur | null = null
+
+  constructor(parent: UIShape | null) {
+    super(parent)
+    this.onTextChanged = new OnChanged((e) => {
+      const { value, isSubmit } = e.value
+
+      const isDirty = this.dirty
+      this.value = value
+      this.dirty = isDirty
```
gonpombo8, line 374: i don't understand this dirty thing.
Why assign to a new prop and then reassign it to this.dirty ?

pbosio: modifying `this.value` will set component as `dirty`.
`dirty` component will send a message with it values to the renderer.
since `this.value` here is being set by a message from the renderer itself, we don't want to send a message back, unless something else than `this.value` have changed.

gonpombo8 (approved):

PR #41, opened 2021-11-05
[WIP] Export scene namespace

gonpombo8:


scripts/build.spec.ts
```diff
@@ -78,9 +78,16 @@ function fixTypes() {
       let content = readFileSync(dtsFile).toString()
 
       content = content.replace(/^export declare/gm, 'declare')
-
       content = content.replace(/^export \{([\s\n\r]*)\}/gm, '')
 
+      /* Lets explain this amazing regex 🧙‍♂️
```
gonpombo8, line 83: Why this ?
Well, we try with @pbosio to add namespaces, but the rollout types generates a namespace with export fields.
Ok, and what's wrong with this?
We have a regex that checks that the types generated doesn't contain any export word.
```ts
if (content.match(/\bexport\b/)) {
  throw new Error(`The file ${dtsFile} contains exports`)
}
```

### types before regex 
```ts
declare namespace Scene {
  */
  * @ public
  * /
  export type OnEnterLeaveOptions = {
    ...
  }

}
```

packages/decentraland-ecs/src/decentraland/Scene.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  export type OnEnterLeaveOptions = {
+    onlyCurrentUser: boolean
+  }
+
+  /**
+   * TThese events are triggered after a character enters the scene.
+   * @public
+   */
+  export function OnEnterScene(callback: (event: IEvents['onEnterScene']) => void, options?: OnEnterLeaveOptions) {
```
menduz, line 18: 1. For consistency sake, this should be an observable also
2. Functions are camelCase not PascalCase

packages/decentraland-ecs/src/decentraland/Scene.ts
```diff
@@ -0,0 +1,57 @@
...
+    ;async () => {
```
menduz, line 19: What is this free function supposed to do?

menduz (changes requested):

PR #44, opened 2021-11-10
fix: missing modules declarations

pbosio:


packages/decentraland-ecs/types/@decentraland/Identity/index.d.ts
```diff
@@ -1,9 +1,29 @@
...
+    hairColor: ColorString
+    eyeColor: ColorString
+    wearables: WearableId[]
+    snapshots: Snapshots
+  }
+
   export type UserData = {
     displayName: string
-    publicKey: string
+    publicKey: string | null
```
menduz, line 22: Was this always null?

pbosio: At least since july 2020 (https://github.com/decentraland/explorer/blob/6b83254c330cc284352d79166da1df17df392054/kernel/packages/shared/apis/UserIdentity.ts)
`publicKey` can be `null`

menduz (approved):

PR #59, opened 2021-12-15
feat: add AttachToAvatar component

pbosio:
add component to be able to attach entities to avatar's bones


packages/@dcl/legacy-ecs/src/decentraland/Components.ts
```diff
@@ -175,6 +176,38 @@ export class Transform extends ObservableComponent {
...
+/** @public */
+export type AttachToAvatarConstructorArgs = {
+  avatarId?: string
+  anchorPointId?: AttachToAvatarAnchorPointId
+}
+
+/**
+ * @public
+ */
+@Component('engine.transform', CLASS_ID.AVATAR_ATTACH)
```
menduz, line 195: Using `engine.transform` means that an ECS entity cannot have both `Transform` and `AttachToAvatar` components. Just like an entity cannot have both `PlaneShape` and `BoxShape`. Is that intended?

packages/@dcl/legacy-ecs/src/decentraland/Components.ts
```diff
@@ -175,6 +176,38 @@ export class Transform extends ObservableComponent {
   }
 }
 
+/** @public */
+export enum AttachToAvatarAnchorPointId {
+  NameTag = 0,
```
menduz, line 181: Should 0 (default) be the AvatarRoot instead?

packages/@dcl/legacy-ecs/src/decentraland/Components.ts
```diff
@@ -175,6 +176,38 @@ export class Transform extends ObservableComponent {
   }
 }
 
+/** @public */
+export enum AttachToAvatarAnchorPointId {
```
menduz, line 180: I'd mark this `@public` as `@alpha` instead and move `AttachToAvatarAnchorPointId` to posix

packages/@dcl/legacy-ecs/src/decentraland/Components.ts
pbosio: Yeap, it's intended. I think it was actually your idea when we talk about this a while ago :P
It makes sense cause it will be always easier to set something as a child of the attached entity and transform that, making transform local to the attached bone

packages/@dcl/legacy-ecs/src/decentraland/Components.ts
pbosio: we talk about this with Alvaro and he decided not to add `AvatarRoot` as an option in this iteration, just the place where the name tag is and the hands

packages/@dcl/legacy-ecs/src/decentraland/Components.ts
menduz: 👍

menduz (approved):

PR #62, opened 2021-12-17
feat: Added six new emotes to PredefinedEmotes

Landeplage:
This change coincides with these two PRs:
https://github.com/decentraland/kernel/pull/138
https://github.com/decentraland/unity-renderer/pull/1648


leanmendoza (approved):

menduz (changes requested):
We don't need this kind of changes if those are not going to be permanent.

Will we have this `SNOWBALL_HIT` forever?

Workaround: use `"snowballthrow" as any` instead of `PredefinedEmote.SNOWBALL_THROW`

nearnshaw:
Agreed, that workaround should work fine, no need to add new values to the enum, as those need to be maintained forever

PR #74, opened 2022-01-05
fix unity-renderer check path

gonpombo8:


menduz (approved):
let's give it a try. Worst case scenario we add a `npm i @dcl/unity-renderer@latest` to the building process to make sure it will _always_ use the latest

PR #86, opened 2022-01-13
fix: plaform export from declaration

leanmendoza:
The export enum wasn't exported correctly.
Fix the error
Screenshot_20220113_122143

### DIFF
--- a/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
+++ b/packages/decentraland-ecs/types/@decentraland/EnvironmentAPI/index.d.ts
@@ -13,7 +13,7 @@ declare module '@decentraland/EnvironmentAPI' {
     configurations: Record<string, string | number | boolean>
   }
 
-  export enum Platform {
+  export const enum Platform {
     DESKTOP = 'desktop',
     BROWSER = 'browser'
   }


menduz (changes requested):
We can't use enums in .d.ts files. An explanation is available in posix/README, it is the same thing.

Alternative is to use `export type platforms = "a" | "b"`

leanmendoza:
> We can't use enums in .d.ts files. An explanation is available in posix/README, it is the same thing.
> 
> Alternative is to use `export type platforms = "a" | "b"`

Yes, I remembered that. In fact my first approach was a type
https://github.com/decentraland/kernel/pull/167/commits/b8ef0b28498c5559fb181785471158a0eb944ffb#diff-5ccab96ae94a1e1188eb82a74f43f59dc746b67b66d1394ae74d2935b3343f0aR26

But I thought an enum was going to be more easy to use for developers (even VS Code suggesting the options).

In the other hand, we already have an enum and here my doubts start.
https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/decentraland-ecs/types/%40decentraland/RestrictedActions/index.d.ts#L23

menduz:
That enum shouldn't be working. @nearnshaw @pbosio are you aware of somebody using it?

@leanmendoza sadly declarations can only hold types, and enums are types AND values. Which won't work for the intended use case

nearnshaw:
The `PredefinedEmote` is used for the TriggerEmote() function. I think pretty much every scene that uses TriggerEmote uses this enum

https://docs.decentraland.org/development-guide/trigger-emotes/

The function expects a  value from the `PredefinedEmote` enum, and Typescript raises an error if you pass a string (although passing a string does work)

gonpombo8:
in fact that PredefinedEmote is working. I think the `const enum` keyword treats the enum as values.
```ts
log(PredefinedEmote.CLAP)
console > 0,0:  clap
.js file compiled > log('Clap', \"clap\" /* CLAP */)
```
I dont know if this has other problems, but if we are not sure we can switch to string literal type

menduz:
Omg, you are right [typescript playground link](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBRQITiCeDuALQeDMAAUOH0qACFWY7EE9U4AEpgiocM0PGQvIh4H0GMD2DQvm4zJyGMxIPxYNB1IkpfA9CIvAAiACiqwYiAW+j1SWxevADlQAHJhAA5SASPVBFQAeS+ZHeghVWkgR1QjgUrmkeFATnFOmGWH0KAQiGhwkucokEuiIgMHGxCyYrA4PCCJWCJW4kDVNzRVJoKPYiG8rNAAG9FCV25VOdyMqAALygO2YGkMO1t0AAXyAA)

menduz (approved):

PR #107, opened 2022-02-04
Allow `types` field in package.json as well as `typings`

bensalilijames:
The TypeScript docs seem to prefer `types` as opposed to `typings`:

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

And there are many packages that use `types` instead of `typings`. So this PR allows the toolchain to support both! 💯

### DIFF
--- a/packages/@dcl/build-ecs/index.ts
+++ b/packages/@dcl/build-ecs/index.ts
@@ -10,6 +10,7 @@ type PackageJson = {
 
   // only package.json
   typings?: string
+  types?: string
 
   bundledDependencies: string[]
   bundleDependencies: string[]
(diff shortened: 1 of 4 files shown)


packages/@dcl/build-ecs/index.ts
```diff
@@ -824,14 +828,16 @@ function validatePackageJsonForLibrary(
...
   }
 
-  if (!packageJson.typings) {
-    throw new Error(`field "typings" in package.json is missing.`)
+  if (!packageJson.typings && !packageJson.types) {
+    throw new Error(`field "typings" or "types" in package.json is missing.`)
   } else {
-    const typingsFile = ts.sys.resolvePath(packageJson.typings)
+    const typingsFile = ts.sys.resolvePath(
+      (packageJson.typings || packageJson.types) as string
```
gonpombo8, line 835: what if u have both keys ?

gonpombo8 (changes requested):

packages/@dcl/build-ecs/index.ts
bensalilijames: If you have both, it prefers `typings` to `types`, this is only a problem if they're different, and if they are, then I would say that's a problem with the published package!

PR #140, opened 2022-06-01
add ecs7 and copy build to decentraland-ecs

gonpombo8:


test/build-ecs/fixtures/ecs7-scene/scene.json
```diff
@@ -0,0 +1,47 @@
+{
+  "ecs7": true,
```
gonpombo8, line 2: this would do the trick. We need to add this prop to @dcl/schemas

PR #143, opened 2022-06-07
Update setup-export.spec.ts

menduz:


leanmendoza:
I think `export` is no longer working because of the changes introduced last and the no maintenance of the feature. Actually is not supported in the last version of CLI:
Screenshot_20220607_094043

PR #158, opened 2022-06-22
fix ecs7 env types

gonpombo8:
Remove legacy es5 env types.
Target to es2018 just in case some browser doesn't support all funcionalities. (??, .?, etc)
Add es2020 lib so we can use new features.
Fix js-runtime types, and copy them to `types/env/index.d.ts` on build

If we dont specify the lib, then we only have es2018 & dom features. That's why we add `lib: ["es2020"]`, so it only includes the es2020.d.ts lib


packages/@dcl/ecs/Makefile
```diff
@@ -25,6 +25,7 @@ test-watch:
 
 build:
 	rm -rf dist/
+	rm -rf src/components/generated
```
leanmendoza, line 28: ❤️

leanmendoza (approved):

PR #159, opened 2022-06-23
feat: add pointer events

Suduck:
This adds pointer-events to the new ECS


packages/@dcl/ecs/src/components/definitions/OnPointerResult.proto
```diff
@@ -0,0 +1,17 @@
+﻿syntax = "proto3";
+
+import "common/id.proto";
+option (ecs_component_id) = 1065;
+import "common/Vector3.proto";
+
+message PBOnPointerResult {
+  string identifier = 1;
+  int64 entity_id = 2;
```
gonpombo8, line 9: why entity_id ?
This could be on the wireMessage, like all the other components.
```ts
const result = PointerResult.getFrom(entity)
```

gonpombo8 (commented):
Can we create an "integration" test where you put a onPointerDown and in the next tick you receive an OnPointerResult ?
Similar to crdt.spec.ts or engine.spec.ts

packages/@dcl/ecs/src/components/definitions/OnPointerResult.proto
Suduck: Removed!

packages/@dcl/ecs/src/components/definitions/OnPointerUp.proto
```diff
@@ -0,0 +1,12 @@
+﻿syntax = "proto3";
+
+import "common/id.proto";
+option (ecs_component_id) = 1061;
+
+message PBOnPointerUp {
+  string identifier = 2;
```
menduz, line 7: Wouldn't be better to use a number? The impact of numeric entity ids was huge

packages/@dcl/ecs/test/components/OnPointerDown.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      showFeedback: true
+    })
+
+    // wait a tick to receive a response
+    newEngine.update(1 / 30)
+
+    // We receive an OnPointerResult
+    const onPointerResult = OnPointerResult.create(entity, {
+      identifier: '98352439-fdsfds',
+      button: 'PRIMARY',
```
menduz, line 52: Same here, an int or enum would be faster and cheaper for the serialization

menduz (changes requested):

packages/@dcl/ecs/src/components/definitions/OnPointerUp.proto
Suduck: done!

packages/@dcl/ecs/test/components/OnPointerDown.spec.ts
Suduck: ANY

leanmendoza (approved):

packages/@dcl/ecs/test/components/OnPointerDown.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  it('should receive OnPointerResult', () => {
+    const newEngine = Engine()
+    const { OnPointerDown, OnPointerDownResult } = newEngine.baseComponents
+    const entity = newEngine.addEntity()
+    OnPointerDown.create(newEngine.addEntity())
+
+    // We create an onPointerDownEvent
+    const onPointerDown = OnPointerDown.create(entity, {
```
gonpombo8, line 36:
```suggestion
OnPointerDown.create(entity, {
```

gonpombo8 (approved):

PR #160, opened 2022-06-23
Refactor renderer transport

gonpombo8:


packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -149,14 +149,11 @@ export type PreEngine = ReturnType<typeof preEngine>
 /**
  * @public
  */
-export function Engine({
-  transports
-}: { transports?: Transport[] } = {}): IEngine {
+export function Engine(
+  { transports }: { transports: Transport[] } = { transports: [] }
```
menduz, line 153: Can this be extracted to a named type?

gonpombo8: sure

PR #176, opened 2022-07-07
fix: send to transport a copied array instead of a reference array. 

leanmendoza:
# What?
Add the method `toBinaryCopied()` to ByteBuffer and use it in the transport sending.

# Why?
If we have more than a `transport`, the data will be changed before being sent.

### DIFF
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -118,7 +118,7 @@ export function crdtSceneSystem({
       }
 
       if (buffer.size()) {
-        transport.send(buffer.toBinary())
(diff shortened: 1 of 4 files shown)


packages/@dcl/ecs/src/systems/crdt/index.ts
```diff
@@ -118,7 +118,7 @@ export function crdtSceneSystem({
       }
 
       if (buffer.size()) {
-        transport.send(buffer.toBinary())
+        transport.send(buffer.toCopiedBinary())
```
menduz, line 121: Shouldn't this be part of the transport _itself_ instead of the specific ecs library?

I'm thinking that all WebWorkerTransports will face the same issue

menduz (approved):

PR #184, opened 2022-07-15
feat: add system priority and a low priority default system that remove result components

leanmendoza:


packages/@dcl/ecs/src/initialization.ts
```diff
@@ -37,3 +37,17 @@ if (dcl) {
       .finally(() => engine.update(dt))
   })
 }
+
```
gonpombo8, line 40: can we move this to the `systems` folder ?
such like we have the crdt system inside the engine

packages/@dcl/ecs/src/engine/systems.ts
```diff
@@ -0,0 +1,59 @@
...
+    systems.push({
+      fn,
+      priority,
+      id
+    })
+    sort()
+    return id
+  }
+
+  function remove(id: SystemId) {
```
gonpombo8, line 40: could this be the fn instead of the id ?

so we can do something like
```ts
function mySystem(dt: number) {
  // some code
}

engine.addSystem(mySystem)

engine.removeSystem(mySystem)
```

packages/@dcl/ecs/src/engine/systems.ts
```diff
@@ -0,0 +1,59 @@
...
+  const systems: SystemItem[] = []
+
+  function sort() {
+    systems.sort((a, b) => b.priority - a.priority)
+  }
+
+  function add(fn: Update, priority: number): SystemId {
+    const id = autoIncrementId++
+
+    if (systems.findIndex((item) => item.fn === fn) !== -1) {
```
gonpombo8, line 27:
```suggestion
    if (systems.find((item) => item.fn === fn)) {|
    }
```

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -30,13 +36,14 @@ function preEngine() {
...
   >()
-  const systems = new Set<Update>()
+  const systems = SystemContainer()
 
-  function addSystem(fn: Update) {
-    if (systems.has(fn)) {
-      throw new Error('System already added')
-    }
-    systems.add(fn)
+  function addSystem(fn: Update, priority: number = SYSTEMS_REGULAR_PRIORITY) {
```
gonpombo8, line 41:
```suggestion
  function addSystem(fn: Update, priority: number = SYSTEMS_REGULAR_PRIORITY, name?: string) {
```

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -155,10 +167,14 @@ export function Engine({ transports }: IEngineParams = {}): IEngine {
   function update(dt: number) {
     crdtSystem.receiveMessages()
 
-    for (const system of engine.systems) {
-      system(dt)
+    // oneshot-components
```
gonpombo8, line 170: remove

packages/@dcl/ecs/src/initialization.ts
gonpombo8: also, call this fn in the `engine.update()` fn instead of adding to the systems container.
So its explicit that we are killing all the results after all the update ticks

PR #196, opened 2022-07-25
feat: re-add  camera mode area component

Suduck:
This readd the camera mode area component to the ECS 7

### DIFF
--- a/packages/@dcl/ecs/tools/protocol-buffer-generation/index.ts
+++ b/packages/@dcl/ecs/tools/protocol-buffer-generation/index.ts
@@ -15,7 +15,6 @@ const NON_EXPOSED_LIST = [
   1090, // Billboard
   1050, // UiTransform
   1070, // AvatarModifierArea,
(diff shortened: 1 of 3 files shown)


packages/@dcl/ecs/tools/protocol-buffer-generation/index.ts
```diff
@@ -15,7 +15,6 @@ const NON_EXPOSED_LIST = [
   1090, // Billboard
   1050, // UiTransform
   1070, // AvatarModifierArea,
-  1071, // CameraModeArea
```
gonpombo8, line 18: run make build. We have a bug that doesnt check if the `.md` changes on build.
So by now we have to always run `make build` so the latests changes gets commited

Suduck: Done!

gonpombo8 (approved):

PR #214, opened 2022-08-04
chore: remove express dependency in the setupProxy.ts (it only uses the type)

leanmendoza:


packages/@dcl/sdk/src/setupProxy.ts
```diff
@@ -78,8 +78,6 @@ const setupProxy = (dcl: any, app: express.Application) => {
     }
   }
 
-  app.use(express.json())
```
menduz, line 81: Isn't this needed?

menduz (approved):

packages/@dcl/sdk/src/setupProxy.ts
gonpombo8: i think we are already calling this method at the cli

gonpombo8 (approved):

PR #220, opened 2022-08-10
fix: preview scenes fetch

pbosio:
added `post` endpoint `/content/entities/active`


kuruk-mm (approved):
tested and it's working!

PR #222, opened 2022-08-17
feat: renaming and docs methods 

leanmendoza:
# What?
## Log and Error
- Expose globally log and error
```ts
// Now
log('Hello world!')

error('Can not fetch the wearable')

```

## Component definitions
The order of the params is now first the definition, then the ID.
To define a component from a generic schema (using `ISchema<T>`), it's available the function `engine.defineComponentFromSchema`.
```ts
// Before
const COMPONENT_ID = 3000 // a value that doesn't matter actually
const CountableId = engine.defineComponent(COMPONENT_ID, Schemas.Map({
  id: Schemas.String,
  count: Schemas.Int
}))

// Now
const CountableId = engine.defineComponent({
  id: Schemas.String,
  count: Schemas.Int
}, COMPONENT_ID)
```
```ts
// Before
const VisibleComponent = engine.defineComponent(2311, Schemas.Bool)

// Now
const VisibleComponent = engine.defineComponentFromSchema(Schemas.Bool, 2311)
```

## Renaming some methods
- Change `ComponentDefinition.getFrom(entity)` to `ComponentDefinition.get(entity)`
- Change `ComponentDefinition.mutable(entity)` to `ComponentDefinition.getMutable(entity)`
- Add `ComponentDefinition.getMutableOrNull(entity)`
- Change `engine.groupOf(ComponentDefinition)` to `engine.getEntitiesWith(ComponentDefinition)`
- Remove `mutableGroupOf`, because it's an opened door to non-performance practices. 
```ts
// Before
for (const [entity, transform] of engine.mutableGroupOf(Transform)) {
  transform.position.x += 1
}

// Now
for (const [entity] of engine.getEntitiesWith(Transform)) {
  Transform.getModifiable(entity).position.x += 1
}

```

  

# Why?
https://github.com/decentraland/sdk/issues/347

### DIFF
--- a/packages/@dcl/ecs/etc/ecs.api.md
+++ b/packages/@dcl/ecs/etc/ecs.api.md
@@ -59,8 +59,8 @@ export type ComponentDefinition<T extends ISchema = ISchema<any>> = {
     create(entity: Entity, val?: ComponentType<T>): ComponentType<T>;
     createOrReplace(entity: Entity, val?: ComponentType<T>): ComponentType<T>;
     deleteFrom(entity: Entity): ComponentType<T> | null;
-    getModifiable(entity: Entity): ComponentType<T>;
-    getModifiableOrNull(entity: Entity): ComponentType<T> | null;
+    getMutable(entity: Entity): ComponentType<T>;
+    getMutableOrNull(entity: Entity): ComponentType<T> | null;
     upsertFromBinary(entity: Entity, data: ByteBuffer): ComponentType<T> | null;
     updateFromBinary(entity: Entity, data: ByteBuffer): ComponentType<T> | null;
     toBinary(entity: Entity): ByteBuffer;
@@ -157,7 +157,7 @@ export namespace Components {
 
 // @public (undocumented)
 export type ComponentSchema<T extends [ComponentDefinition, ...ComponentDefinition[]]> = {
-    [K in keyof T]: T[K] extends ComponentDefinition ? ReturnType<T[K]['getModifiable']> : never;
+    [K in keyof T]: T[K] extends ComponentDefinition ? ReturnType<T[K]['getMutable']> : never;
 };
 
 // Warning: (ae-forgotten-export) The symbol "EcsResult" needs to be exported by the entry point index.d.ts
--- a/packages/@dcl/ecs/etc/ecs.api.md
+++ b/packages/@dcl/ecs/etc/ecs.api.md
@@ -54,12 +54,13 @@ export const CameraModeArea: ComponentDefinition<ISchema<PBCameraModeArea>>;
 export type ComponentDefinition<T extends ISchema = ISchema<any>> = {
     _id: number;
     has(entity: Entity): boolean;
-    getFrom(entity: Entity): DeepReadonly<ComponentType<T>>;
+    get(entity: Entity): DeepReadonly<ComponentType<T>>;
     getOrNull(entity: Entity): DeepReadonly<ComponentType<T>> | null;
     create(entity: Entity, val?: ComponentType<T>): ComponentType<T>;
-    mutable(entity: Entity): ComponentType<T>;
     createOrReplace(entity: Entity, val?: ComponentType<T>): ComponentType<T>;
     deleteFrom(entity: Entity): ComponentType<T> | null;
+    getModifiable(entity: Entity): ComponentType<T>;
+    getModifiableOrNull(entity: Entity): ComponentType<T> | null;
     upsertFromBinary(entity: Entity, data: ByteBuffer): ComponentType<T> | null;
     updateFromBinary(entity: Entity, data: ByteBuffer): ComponentType<T> | null;
     toBinary(entity: Entity): ByteBuffer;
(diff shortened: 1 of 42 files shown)


menduz:
- What is the rationale to choose modifiable instead of mutable?
- Component ID is actually very important to store information of the components in a file.. having that defined by autonumeric values will break the loading/saving of the scenes depending on the evaluation order of the .defineComponent and those components will not be synchronizable for the same reason

leanmendoza:
> * What is the rationale to choose modifiable instead of mutable? https://stackoverflow.com/questions/652678/is-mutable-the-same-as-the-word-modifiable-in-relation-to-strings
> * Component ID is actually very important to store information of the components in a file.. having that defined by autonumeric values will break the loading/saving of the scenes depending on the evaluation order of the .defineComponent and those components will not be synchronizable for the same reason

I kept the rationale of the issue https://github.com/decentraland/sdk/issues/347 for both items. IMO I like `mutable` instead of `modifiable` because I feel it is more accurate but it's true that it needs a level up of understanding when you read it the first time.  

About the component id, it's a good point. Should I revert to the `optional`? Thinking about this point spawns my other thoughts about serialization/deserialization problems (like scene versions). But after deep thinking, this change can introduce big avoidable compatibilities (like in hot reloads).

menduz:
I think it is safer to make the component id required for the time being, we can defer this conversation to revisit this again a near future.

And for mutable, I think that when it comes to programming it is good to talk about mutability/immutability vs modifiability vs inmodifiability.. I wouldn't change it, it may be dangerous to introduce new concepts

packages/@dcl/ecs/etc/ecs.api.md
```diff
@@ -430,9 +437,11 @@ export namespace Vector3 {
...
 // Warnings were encountered during analysis:
 //
-// dist/engine/types.d.ts:25:5 - (ae-forgotten-export) The symbol "Update" needs to be exported by the entry point index.d.ts
-// dist/engine/types.d.ts:32:5 - (ae-forgotten-export) The symbol "SdkComponents" needs to be exported by the entry point index.d.ts
-// dist/engine/types.d.ts:38:5 - (ae-forgotten-export) The symbol "Transport" needs to be exported by the entry point index.d.ts
+// dist/engine/types.d.ts:26:5 - (ae-forgotten-export) The symbol "Update" needs to be exported by the entry point index.d.ts
+// dist/engine/types.d.ts:28:5 - (ae-forgotten-export) The symbol "Spec" needs to be exported by the entry point index.d.ts
+// dist/engine/types.d.ts:28:5 - (ae-forgotten-export) The symbol "Result" needs to be exported by the entry point index.d.ts
+// dist/engine/types.d.ts:33:5 - (ae-forgotten-export) The symbol "SdkComponents" needs to be exported by the entry point index.d.ts
+// dist/engine/types.d.ts:39:5 - (ae-forgotten-export) The symbol "Transport" needs to be exported by the entry point index.d.ts
```
menduz, line 444: Please export these types

gonpombo8 (dismissed):

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -59,10 +61,23 @@ function preEngine() {
...
+        i++
+      }
+      return i
+    }
+    return componentId
+  }
+
+  function defineComponentFromSchema<T extends ISchema>(
+    spec: T,
+    componentIdDefined?: number
```
gonpombo8, line 78: im not sure about this. The component ID is an important thing to be created randomly.

I prefer to be a mandatory field and expose a method that is getAvailableComponentID or something like that, so the creator knows how this works and whats the pro/cons of doing this

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -83,22 +105,11 @@ function preEngine() {
     return component
   }
 
-  function* mutableGroupOf<
```
gonpombo8, line 86: thanku

gonpombo8 (commented):
What about the documentation/comments for the defualt values ? 
Can we add it here too ?

packages/@dcl/ecs/src/initialization.ts
```diff
@@ -37,3 +37,6 @@ if (dcl) {
       .finally(() => engine.update(dt))
   })
 }
+
+export const log = dcl.log
+export const error = dcl.error
```
gonpombo8, line 42: If we expose this methods maybe we should remove the `dcl` types for the scene ?

leanmendoza: yep, that's the idea :D

packages/@dcl/ecs/src/engine/component.ts
```diff
@@ -20,33 +20,174 @@ export type ComponentType<T extends ISchema> = EcsResult<T>
  */
 export type ComponentDefinition<T extends ISchema = ISchema<any>> = {
   _id: number
+
+  /**
+   * Return the default value of the current component
```
menduz, line 25:
```suggestion
   * Returns the default value of the current component
```

menduz (approved):

PR #226, opened 2022-08-19
feat: transform with optionals

leanmendoza:
# What?
Create a wrapper for the Transform definition. Now, the fields are optional and you can pass only the position if you wish.

```ts
// now
Transform.create(entity) // ok => create the default transform
Transform.create(entity, { position: { x: 1, y: 2, z: 3} }) // wrong, all fields are mandatory (even if the value that I'd put them are the defaults)
 
// now
Transform.create(entity, { position: { x: 1, y: 2, z: 3} }) => ok, the other fields are filled within default ones

```

### DIFF
--- a/packages/@dcl/ecs/etc/ecs.api.md
+++ b/packages/@dcl/ecs/etc/ecs.api.md
@@ -8,32 +8,32 @@
 // Warning: (ae-forgotten-export) The symbol "PBAnimator" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export const Animator: ComponentDefinition<ISchema<PBAnimator>>;
+export const Animator: ComponentDefinition<ISchema<PBAnimator>, PBAnimator>;
 
 // Warning: (ae-forgotten-export) The symbol "PBAudioSource" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export const AudioSource: ComponentDefinition<ISchema<PBAudioSource>>;
+export const AudioSource: ComponentDefinition<ISchema<PBAudioSource>, PBAudioSource>;
 
 // Warning: (ae-forgotten-export) The symbol "PBAvatarAttach" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export const AvatarAttach: ComponentDefinition<ISchema<PBAvatarAttach>>;
+export const AvatarAttach: ComponentDefinition<ISchema<PBAvatarAttach>, PBAvatarAttach>;
 
 // Warning: (ae-forgotten-export) The symbol "PBAvatarShape" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export const AvatarShape: ComponentDefinition<ISchema<PBAvatarShape>>;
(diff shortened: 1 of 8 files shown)


packages/@dcl/ecs/etc/ecs.api.md
```diff
@@ -68,25 +68,476 @@ export type ComponentDefinition<T extends ISchema = ISchema<any>> = {
...
-    x: number;
-    y: number;
-    z: number;
-    w: number;
-    };
-    scale: {
-    x: number;
-    y: number;
-    z: number;
+    Transform: {
```
gonpombo8, line 71: this types are weird, maybe the spread operator ?

packages/@dcl/ecs/tools/protocol-buffer-generation/generateIndex.ts
```diff
@@ -11,6 +13,9 @@ function importComponent(component: Component) {
 }
 
 function defineComponent(component: Component) {
+  if (component.componentId === TransformComponent.componentId) {
```
gonpombo8, line 16: what if instead of doing this wrapper, we create a new param to `defineComponent` with defaultValues.
We can use this for custom components also

```ts
const PlayerType = MapType({
  velocity: Vector3,
  position: Vector3,
  name: EcsString,
  etc: MapType({ ... })
})

const defaultValues = {
  velocity: { x: 0, y: 0, z: 0 },
  position: { x: 0, y: 0, z: 0 },
}

const PlayerComponent = engine.defineComponent(PlayerType, 1000, defaultValues)
PlayerComponent.create(entity, { name: 'myName' })
```

test/build-ecs/fixtures/ecs7-scene/game.ts
```diff
@@ -18,9 +18,7 @@ function createCube(x: number, y: number, z: number) {
   const myEntity = engine.addEntity()
 
   Transform.create(myEntity, {
-    position: { x, y, z },
-    scale: { x: 1, y: 1, z: 1 },
-    rotation: { x: 0, y: 0, z: 0, w: 1 }
+    position: { x, y, z }
```
gonpombo8, line 21: can we add some test to the ecs library for the transform?

packages/@dcl/ecs/etc/ecs.api.md
menduz: or nominal types

packages/@dcl/ecs/tools/protocol-buffer-generation/generateIndex.ts
menduz: It may kill the performance?

leanmendoza: https://github.com/decentraland/js-sdk-toolchain/pull/230

leanmendoza: This is a WIP approach; it extends the functionality to all the Schemas and may be helpful. Schemas always serialize their fields so that the performance won't be affected.

PR #230, opened 2022-08-24
feat: default constructors

leanmendoza:
A general approach to https://github.com/decentraland/js-sdk-toolchain/pull/226

# What?
- This adds the possibility of determining a default value (constructor) when you define a component.
- This default value is serialized at definition time, and then it will be deserialized every time is created. This is necessary to delegate the responsibility of a deep clone to `deserialize` and `create` methods from the schema. Otherwise, the object reference will be cloned and this can create mistakes.
- Wrap `Transform` and `MeshRenderer`
- Add `removeComponentDefinition` to remove already defined components, (to create a wrapper). It's not exposed to the creator and it only should be used at the scene-creation time.

## Default constructor
Enable to set a default value when the component is created, it's really useful for internal use creating wrappers. But also, they can extend the functionality to create components with Schemas and constructors:
If we are defining a new component, and our schema look as this:
```ts
const NpcComponent = engine.defineComponent(
{
    name: Schemas.String,
    shouldBeProccessed: Schemas.Bool, 
}, 1111)
``` 

When we assign to an entity: we can pass without a value, or pass one with all fields filled.
```ts
NpcComponent.create(entity)
// or
NpcComponent.create(entity, { name: 'cool name', shouldBeProccessed: true })

// wrong
NpcComponent.create(entity, { name: 'cool name' })
```

With this new feature, we can pass a default constructor:
```ts
const NpcSchema = {
    name: Schemas.String,
    shouldBeProccessed: Schemas.Bool, 
}
const NpcComponent = engine.defineComponent(NpcSchema, 1111,  { shouldBeProccesd: true })

// and now yes
NpcComponent.create(entity, { name: 'cool name' })
```

## Deprecate support for primitive components
When we get a particular value of a component with a primitive Schema (Bool, Int, Number, etc), the returned value is a copy, not a reference, so the real value won't actually be modified. 
```ts
const MyFlagComponent = engine.defineComponent(Schemas.Bool, 1111)
for (const [entity, myFlag] of engine.getEntitiesWith(MyFlagComponent)){
   if (myFlag) {
      // do something
   }
} 
```
Because of this, the only allowed type for component definition is a `Record<string, unknown>` 

# Why?
To create component wrappers.


packages/@dcl/ecs/src/engine/component.ts
```diff
@@ -191,18 +194,46 @@ export type ComponentDefinition<T extends ISchema = ISchema<any>> = {
...
+  function getDefaultValue() {
+    if (constructorDefault) {
+      return spec.deserialize(
+        createByteBuffer({
+          writing: {
+            buffer: defaultBuffer.buffer(),
+            currentOffset: defaultBuffer.currentWriteOffset()
+          }
+        })
+      )
```
menduz, line 223: I'd add a comment here to evaluate the performance of this method in the future. It seems to me that generating the defaults using this method instead of a deep-clone may be a lot of heavylifting. After feature parity we should revisit this and measure the performance

menduz (approved):

packages/@dcl/ecs/src/engine/component.ts
leanmendoza: https://github.com/decentraland/sdk/issues/430

PR #239, opened 2022-08-29
feat: add cyclic parenting checker

leanmendoza:
# What?
Add a cyclic parenting checker.
Test copied from `unity-renderer` implementation

# Why? 
https://github.com/decentraland/sdk/issues/399


packages/@dcl/ecs/src/engine/component.ts
```diff
@@ -62,7 +62,7 @@ export type ComponentDefinition<
    * Transform.get(otherEntity) // throw an error!!
    * ```
    */
-  get(entity: Entity): DeepReadonly<ComponentType<T>>
+  get(entity: DeepReadonly<Entity> | Entity): DeepReadonly<ComponentType<T>>
```
menduz, line 65: Why is this necessary? Entity should be a opaque identifier

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -196,6 +196,20 @@ export function Engine({ transports }: IEngineParams = {}): IEngine {
...
+      let transform = baseComponents.Transform.getOrNull(entity)
+      while (transform && transform.parent) {
+        if (transform.parent === entity) {
+          dcl.error(`There is a cyclic parent with entity ${entity}`)
+          break
+        } else {
+          transform = baseComponents.Transform.getOrNull(transform.parent)
+        }
+      }
+    }
```
menduz, line 211: I think there will be a tremendous penalty cost associated with this checks, can seting a "getter/setter" to the transform help reduce that number? should we do it only in dev mode instead? and not run the validation in production code?

leanmendoza: This `for iteration` is a system, and even it might be an optional system for debugger purposes.

packages/@dcl/ecs/src/engine/component.ts
leanmendoza: I can take a look to the DeepReadonly type, but without this, it happens the next:
```ts
const myTransform = Transform.get(entityA)
// typeof myTransform.parent => DeepReadonly<Entity>
const myParentComponent = MeshCollider.get(myTransform.parent) // mismatch types between Entity and DeepReadonly<Entity>
```

menduz: Does it make sense to change `DeepReadonly<Entity> -> Entity` in the transform definition then?

menduz (approved):

PR #243, opened 2022-08-31
fix: deep readonly

leanmendoza:
# What? 
Fix the deep read-only generic type for component data.


packages/@dcl/ecs/etc/ecs.api.md
```diff
@@ -318,8 +327,10 @@ export namespace Quaternion {
         w: number;
     };
     export function normalize(q: ReadonlyQuaternion): MutableQuaternion;
+    // Warning: (ae-forgotten-export) The symbol "DeepReadonly" needs to be exported by the entry point index.d.ts
+    //
     // (undocumented)
-    export type ReadonlyQuaternion = DeepReadonly<MutableQuaternion>;
+    export type ReadonlyQuaternion = DeepReadonly_2<MutableQuaternion>;
```
gonpombo8, line 333: we need to export DeepReadOnly.
Also why this have the `_2`. Are we still exporting the old one ? Or maybe another interface with the same name?

gonpombo8 (approved):

packages/@dcl/ecs/etc/ecs.api.md
leanmendoza: I couldn't remove the _2, I don't know why that is happening. The old one is provided by the @dcl/ecs-math but it seems it's automatically exported because of the dependency in Vector3, etc. The explicit reference to the deep read only wes removed

PR #252, opened 2022-09-12
feat: remove old observable and add message bus from ecs6

leanmendoza:


packages/@dcl/ecs/src/runtime/MessageBus.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (!_messageObserver) {
+    _messageObserver = new Observable<IEvents['comms']>()
+  }
+  return _messageObserver
+}
+
+function ensureCommunicationsController() {
+  if (!communicationsControllerPromise) {
+    communicationsControllerPromise = dcl.loadModule(
+      '@decentraland/CommunicationsController',
```
menduz, line 21: ~system?

leanmendoza: thx!

PR #280, opened 2022-10-10
feat: add input states

leanmendoza:
Refactor the `wasEntityClicked` and `isPointerActive`:
- Group all functions in an `Input` object
- The entity parameter is moved to the end of the list, and it's optional, to handle global events.
- Names change to `wasJustClicked` and `wasPointerJustActive` respectively

Add `isActionDown` to query if an action is being pressed.
Add `getInputCommand` and `getClick` to retrieve info about the event (raycast, origin, direction, etc)


kuruk-mm (approved):
Code looks great!

PR #296, opened 2022-10-19
add execute task system. Reorganize engine intiialization fn

gonpombo8:
Add support for async functions
```tsx
// engine scene stuff


async function fetchData() {
  const data = await fetch('some-api')
  return data
}

async function postSomething() {

}

executeTask(() => {
  const data = await fetchData()
  await postSomething(data.profile)
})
```


packages/@dcl/ecs/src/runtime/initialization/dcl.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (response.data?.length) {
+        if (rendererTransport.onmessage) {
+          for (const byteArray of response.data) {
+            rendererTransport.onmessage(byteArray)
+          }
+        }
+      }
+    }
+
+    dcl.onUpdate((dt: number) => {
```
menduz, line 27: onUpdate should now accept an async function and should block the next update until the promise is completed

menduz (commented):
yes

packages/@dcl/ecs/src/runtime/initialization/dcl.ts
leanmendoza: pullRendererMessages should be awaited, but tasks can be deferred through the event loop and not in the system's pipeline

leanmendoza (approved):

PR #299, opened 2022-10-20
feat: add math schemas to global namespace Schemas

leanmendoza:
# What?
Add `Vector3`, `Quaternion`, `Color3` and `Color4` schemas to namespace `Schemas`


packages/@dcl/ecs/src/schemas/index.ts
```diff
@@ -28,6 +32,11 @@ export namespace Schemas {
 
   export const Number = Float32
 
+  export const Vector3 = Vector3Schema
+  export const Quaternion = QuaternionSchema
+  export const Color3 = Color3Schema
+  export const Color4 = Color4Schema
```
menduz, line 38: Won't this conflict with ecs-math if exported globally?

leanmendoza: Nop, this is only accessible with `Schemas.Vector3`, `Schemas.Color3`, etc.

menduz (approved):

PR #300, opened 2022-10-20
feat: add remove entity with children

leanmendoza:
# What? 
Add `engine.removeEntityWithChildren` that scans all the tree children and calls `engine.removeEntity` for each one.


packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -202,10 +202,38 @@ export function Engine({ transports }: IEngineParams = {}): IEngine {
...
+
+    // This avoid infinite loop when there is a cyclic parenting
+    if (proccesedEntities.find((value) => firstEntity === value)) return []
+    proccesedEntities.push(firstEntity)
+
+    for (const [entity, value] of engine.getEntitiesWith(
+      baseComponents.Transform
+    )) {
+      if (value.parent === firstEntity) {
+        const childrenOfChildren = getTreeEntityArray(entity, proccesedEntities)
```
menduz, line 219:
```suggestion
        yield* getTreeEntityArray(entity, proccesedEntities)
```

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -202,10 +202,38 @@ export function Engine({ transports }: IEngineParams = {}): IEngine {
     }
   }
 
+  function getTreeEntityArray(
```
menduz, line 205:
```suggestion
  function* getTreeEntityArray(
```

packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -202,10 +202,38 @@ export function Engine({ transports }: IEngineParams = {}): IEngine {
...
+    for (const [entity, value] of engine.getEntitiesWith(
+      baseComponents.Transform
+    )) {
+      if (value.parent === firstEntity) {
+        const childrenOfChildren = getTreeEntityArray(entity, proccesedEntities)
+        children.push(...childrenOfChildren)
+      }
+    }
+
+    return [firstEntity, ...children]
```
menduz, line 224:
```suggestion
    yield firstEntity
```

menduz (approved):
You can use generators to prevent array iterations. But taking into account the amount of times this is going to be called it doesn't make much sense TBH

PR #315, opened 2022-11-01
refactor Input events. Use EventsManager on ReactEcs

gonpombo8:
# What ?? 
Input.isPressed for global input state
Input.isTriggered for specific entity action events
Mark Input.isClicked as internal for the moment
Mark EventsManager.click as internal
Remove React UI Events internal Map, and use EventsManager.onPointerDown instead (Lets the engine handle all this logic)
Add 100% Coverage, and fail if it is not covered 🎉


jest.config.js
```diff
@@ -20,7 +26,7 @@ module.exports = {
   coveragePathIgnorePatterns: [
     "scripts",
     "test",
-    "packages/@dcl/react-ecs"
+    // "packages/@dcl/react-ecs"
```
leanmendoza, line 29: comment

packages/@dcl/ecs/src/components/index.ts
```diff
@@ -40,6 +40,6 @@ export function defineSdkComponents(engine: PreEngine) {
  * })
  * ```
  */
-export function makeCollisionMask(...layers: ColliderLayer[]): number {
-  return layers.reduce((item, currentValue) => item | currentValue)
-}
+// export function makeCollisionMask(...layers: ColliderLayer[]): number {
```
leanmendoza, line 43: comment

packages/@dcl/ecs/src/index.ts
```diff
@@ -14,6 +14,6 @@ export * from './runtime/types'
 
 export * from './runtime/observables'
 export * from './runtime/temp-fp/Observable'
-export * from './runtime/MessageBus'
+export * from './runtime/messageBus'
 
-export { makeCollisionMask } from './components'
+// export { makeCollisionMask } from './components'
```
leanmendoza, line 19: comment

nearnshaw:
@gonpombo8 
The name of `Input.isActive` can potentially be confused with `Input.IsActionDown`

The first triggers only on the tick right after the button went down. It's an event that is already over.
The second triggers also on every tick that the button remains down. It's an ongoing event.

How about `Input.wasActive` or `Input.wasActivated`
or `Input.wasTriggered`?

leanmendoza (approved):
LGTM 😄 I marked the dead code to remove it

leanmendoza:
> @gonpombo8 The name of `Input.isActive` can potentially be confused with `Input.IsActionDown`
> 
> The first triggers only on the tick right after the button went down. It's an event that is already over. The second triggers also on every tick that the button remains down. It's an ongoing event.
> 
> How about `Input.wasActive` or `Input.wasActivated` or `Input.wasTriggered`?

I'm more towards keeping the `Input.isActive`. The event is known in the current tick, so you ask for something that is happening and known in the current tick.

About `Input.isActionDown`, it remains in continuity. So it's valid to say that every tick is true the event is happening. We can orientate it to `Input.isActionPressing` or something that remarks continuity.

PR #316, opened 2022-11-02
feat: add animator helper and Schemas.Entity

leanmendoza:
# What?
Add helper functions extension to the component definition of `Animator`. 
- playSingleAnimation
- stopAllAnimations
- getClip/getClipOrNull

Also, add the `Schemas.Entity`


packages/@dcl/ecs/src/components/extended/Animator.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export interface AnimatorComponentDefinition extends ComponentDefinition {
+  /**
+   * @public
+   *
+   * [Helper] Get a `mutable` version of animator clip
+   * @param entity
+   * @param name
+   */
+  getClip(entity: Entity, name: string): PBAnimationState | null
```
gonpombo8, line 16: Maybe we need to follow our pattern of getSomethingOrNull ?  `getClipOrNull`.
Not sure, im not a big fan of Component.getOrNull. But thats another discussion

packages/@dcl/ecs/src/components/extended/Animator.ts
```diff
@@ -0,0 +1,94 @@
...
+      entity: Entity,
+      name: string,
+      shouldReset: boolean = true
+    ): boolean {
+      // Get the mutable to modifying
+      const animator = Animator.getMutableOrNull(entity)
+      if (!animator) return false
+
+      const animItem = animator.states.find((item) => item.name === name)
+      if (!animItem) return false
```
gonpombo8, line 66: can we reuse the getClip fn here ?

packages/@dcl/ecs/src/components/extended/Animator.ts
```diff
@@ -0,0 +1,94 @@
...
+    getClip(entity: Entity, name: string): PBAnimationState | null {
+      const anim = Animator.getMutableOrNull(entity)
+      if (!anim) return null
+
+      const state = anim.states.find((item) => item.name === name)
+      if (!state) return null
+
+      return state
+    },
+    playSingleAnim(
```
gonpombo8, line 56:
```suggestion
    playSingleAnimation(
```
or

```suggestion
    play(
```

packages/@dcl/ecs/src/components/extended/Animator.ts
```diff
@@ -0,0 +1,94 @@
...
+        state.shouldReset = true
+      }
+
+      animItem.playing = true
+      animItem.shouldReset = shouldReset
+
+      return true
+    },
+
+    stopAnims(entity: Entity, resetCursor: boolean = true): boolean {
```
gonpombo8, line 80:
```suggestion
    stop(entity: Entity, resetCursor: boolean = true): boolean {
```

packages/@dcl/ecs/src/components/index.ts
```diff
@@ -2,9 +2,11 @@ import { PreEngine } from '../engine'
 
 import { defineLibraryComponents } from './generated/index.gen'
 import { ColliderLayer } from './generated/pb/decentraland/sdk/components/mesh_collider.gen'
-
 import { defineTransformComponent } from './legacy/Transform'
 
+import * as AnimatorSchema from './generated/Animator.gen'
```
gonpombo8, line 7: remove ?

packages/@dcl/ecs/src/schemas/custom/Entity.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export const EntitySchema: ISchema<Entity> = {
+  serialize(value: Entity, builder: ByteBuffer): void {
+    builder.writeInt32(value)
+  },
+  deserialize(reader: ByteBuffer): Entity {
+    return reader.readInt32() as Entity
+  },
+  create() {
+    return 0 as Entity
```
gonpombo8, line 16: can we create a test for this schema ? 😄

packages/@dcl/ecs/src/components/extended/Animator.ts
gonpombo8: I think we can overload this fn and depend on the params you play a single animation or not

packages/@dcl/ecs/etc/ecs.api.md
```diff
@@ -6,10 +6,10 @@
 
 /// <reference types="@dcl/posix" />
 
-// Warning: (ae-forgotten-export) The symbol "PBAnimator" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "AnimatorComponentDefinition" needs to be exported by the entry point index.d.ts
```
menduz, line 9: missing export!

packages/@dcl/ecs/src/components/extended/Animator.ts
leanmendoza: both, animItem and animtor are required in the function. getClip only returns animitem

packages/@dcl/ecs/src/components/extended/Animator.ts
leanmendoza: I think the `orNull` should be deprecated. It always exists the possibility that something is undefined, so if you are really sure that it doesn't, you can use `!`, and the animation can be a good use-case

packages/@dcl/ecs/src/components/extended/Animator.ts
leanmendoza: the play is very generic at this point, I chose the verbose one. If it doesn't fit in the future, we can deprecate it and add the generic one.

packages/@dcl/ecs/src/components/extended/Animator.ts
leanmendoza: Same thinking as comment with `play`

packages/@dcl/ecs/src/components/extended/Animator.ts
menduz: The convention so far was that we never use `!` due to linter and error prone of the code, and then orNull can return something or null (this is very functional and monadic)

And if something was `getSomething(x): Something`, it can fail if Something doesn't exist. On the contrary, `getSomethingOrNull(x): Something | null` enforces good type checking and doesn't throw

leanmendoza: Thanks, that point had gone from me. In that case, I think we can write both functions `getClip` and `getClipOrNull`. 
What about the `play` and `stop` functions? They return `true | false`, true if they find the clip and change the info, false otherwise.

menduz: sounds reasonable to return a boolean and defaulting to a value in that case

gonpombo8 (approved):

PR #317, opened 2022-11-02
feat: bundle api declarations

leanmendoza:
# What?
Bundle APIs from @dcl/protocol to a single declaration file and reference it in the `@dcl/js-runtime` types.


menduz:
WAIT! I've been waiting for this moment for more than a year

Look, there is one **veeery good** way to do this: creating a package, and referencing it in the "types" section of the tsconfig.json.

The name of the originally intended package was `@dcl/js-runtime`, if that is installed, every project with `"types": ["@dcl/js-runtime"]` will have complete types for Decentraland.

Without minor modifications, you should be able to add this change to the @dcl/js-runtime/index.d.ts (maybe using `/// <reference file="modules.d.ts" />`) and that way, removing the awful module resoution hacks

gonpombo8 (approved):

packages/@dcl/js-runtime/index.d.ts
```diff
@@ -1,4 +1,5 @@
 /// --- FETCH ---
+/// <reference types="./apis" />
```
menduz, line 2:
```suggestion
/// <reference path="./apis.d.ts" />
```

packages/@dcl/sdk/types/tsconfig.ecs7.json
```diff
@@ -22,23 +22,11 @@
...
-      "rpc-modules/ParcelIdentity",
-      "rpc-modules/Permissions",
-      "rpc-modules/Players",
-      "rpc-modules/PortableExperiences",
-      "rpc-modules/SignedFetch",
-      "rpc-modules/SocialController",
-      "rpc-modules/UserActionModule",
-      "rpc-modules/UserIdentity"
+      "@dcl/js-runtime",
+      "ecs7"
```
menduz, line 26: what is ecs7?

leanmendoza: ecs7 is the lib to use the ecs7 (and the runtime initialization) 
`js-runtime` only has minimal exposure by the `scene-runtime`. You could use `@dcl/js-runtime` and compile some custom ecs or lib.

menduz: OH perfect

menduz:
From the generated files all non-functions should be trimmed, like this one `export const protobufPackage = "decentraland.common.sdk"`

Also, there is a cosmetic change to remove a couple comments that may also be good

```
  // ########### BLOCK

  // importtype { GetUserPublicKeyRequest, GetUserPublicKeyResponse, GetUserDataRequest, GetUserDataResponse } from "proto/decentraland/kernel/apis/user_identity.gen";
```

menduz (approved):

PR #334, opened 2022-11-23
fix: receive crdt

leanmendoza:
# What?
Call the sendCrdt even if there is no new data to send.
Also, add runtime tests.

# Why?
The response is the update from the renderer, so it's important to make always the call


packages/@dcl/ecs/src/runtime/initialization/dcl.ts
```diff
@@ -1,12 +1,20 @@
...
+  engine: IEngine,
+  rendererTransport: RendererTranport
+) {
   if (typeof dcl !== 'undefined') {
     dcl.loadModule('~system/EngineApi', {}).catch(dcl.error)
 
     dcl.onUpdate((dt: number) => {
       engine.update(dt)
+      if (!rendererTransport.wasCalled()) {
+        rendererTransport.send(new Uint8Array([]))
```
gonpombo8, line 15: what if we call `.send` at the crdt system even if the buffer.size is empty instead of adding all this new logic?

gonpombo8: maybe we can add some prop to the transport interface `shouldSendOnEmtpy: boolean`.
So if in the future we need to another transport with the same logic we can also reuse this, and all the logic is contained in the crdt system

leanmendoza: I think it's to assign responsibility to the CRDT system only because of this. The transport definition until now is ok for me. If there is no message, the send isn't invoked.

If we want to implement the functionality of tick-pinging, I think we should well-describe the message and its purpose.

🤔

PR #336, opened 2022-11-26
Rollup

menduz:
Implement rollup to:
- treeshake unused code (working with components)
- regular typescript libraries (like ethers.js)
- remove AMD dependency, and reduce overhead of maintenance, learning-curve and daily use

## Other PR counterparts:
- https://github.com/decentraland/adr/pull/141 Documentation
- https://github.com/decentraland/scene-runtime/pull/50 Separated runtimes for SDK6 & SDK7
- https://github.com/decentraland/kernel/pull/733 Kernel implementation

# Breaking changes:

As per ADR-133 Scenes now need to export an `onUpdate` function. This is done in scenes via `export * from '@dcl/sdk'`. This is not a definitive solution, since the ergonomics are not sound. But it is way better than implementing a wrong abstraction from the ground up. 

Playground doesn't work anymore because AMD was removed.


leanmendoza:
>  Maybe remove components from the engine.baseComponents in favor of idempotent Component.forEngine(engine).getFrom(entity) to help type checker, tree shaking, compute and bundle size

To do this, the component definition should be defined on the fly. Otherwise, all definitions are always previously invoked so the code will be in the bundle.

I think we can auto-generate every file with the `@public` exposed of something like:
```
import { engine } from './../runtime'
import { defineMeshRenderer } from './extended/MeshRenderer.ts'

// @public
export const MeshRenderer = defineMeshRendererExtended(engine)
```

In this way, the syntax would be similar to the current and we could remove the `baseComponents`.

menduz:
There are two things motivating this change:
- There can be more than an engine
- Not all components need to be registered, because not all of them will be used
  - And iterating over all components every engine.update should only work if they are used

Thus the solution, lazily create the components when they are needed. For a specific engine.

menduz:
> > Maybe remove components from the engine.baseComponents in favor of idempotent Component.forEngine(engine).getFrom(entity) to help type checker, tree shaking, compute and bundle size
> 
> To do this, the component definition should be defined on the fly. Otherwise, all definitions are always previously invoked so the code will be in the bundle.
> 
> I think we can auto-generate every file with the `@public` exposed of something like:
> 
> ```
> import { engine } from './../runtime'
> import { defineMeshRenderer } from './extended/MeshRenderer.ts'
> 
> // @public
> export const MeshRenderer = defineMeshRendererExtended(engine)
> ```
> 
> In this way, the syntax would be similar to the current and we could remove the `baseComponents`.

Ended up doing something like that. Check this commit! https://github.com/decentraland/js-sdk-toolchain/pull/336/commits/9c615f3e4f0c49c6b490bef3fbc665c6a99b1393

packages/@dcl/react-ecs/src/system.ts
```diff
@@ -15,7 +15,7 @@ export function createReactBasedUiSystem(
   pointerSystem: PointerEventsSystem
 ): ReactBasedUiSystem {
   const renderer = createReconciler(engine, pointerSystem)
-  let uiComponent: UiComponent = () => ({})
+  let uiComponent: UiComponent
 
   function ReactBasedUiSystem() {
     renderer.update(uiComponent())
```
menduz, line 21: oops? `cannot call function uiComponent is undefined`

leanmendoza: what about https://github.com/decentraland/js-sdk-toolchain/pull/336/commits/9f1223c08f02b9e881125d3fb1f460844db33199

packages/@dcl/ecs/src/ecs.md
```diff
@@ -1,6 +1,6 @@
 ```ts
 /**
- * This is te first approach we discuss with @mendez
+ * This is te first approach we discuss with @menduz
```
leanmendoza, line 3: 😆

gonpombo8 (approved):

PR #348, opened 2022-12-02
feat add async fn for transports

gonpombo8:
Add async send transport.

Now on every update tick we wait for the crdt to process all the messages, then run the system fns, and then wait for the crdt system to process & send all the messages.
```ts
async function update(dt: number) {
  await crdt.receiveMessages()
  systems.fn()
  await crdt.sendMessages()
}
```


packages/@dcl/ecs/src/systems/crdt/index.ts
```diff
@@ -164,32 +164,53 @@ export function crdtSceneSystem(engine: Pick<IEngine, 'getComponentOrNull'>) {
...
   }
 
+  /**
+   * @public
+   * Add a transport to the crdt system
+   */
   function addTransport(transport: Transport) {
+    if (transports.find((t) => transport.type === t.type)) {
+      throw new Error('Transport already added')
+    }
```
menduz, line 198: What is the rationale behind this if? It seems like the addTransport should not care about anything related to the type of the transport. i.e. we sould be able to add 15 ws transports if desired

menduz (approved):
Amazing job. Only left one comment 🚀

menduz (changes requested):
This impl is not working as intended. Will push tests soon to another PR

gonpombo8:
Closed via https://github.com/decentraland/js-sdk-toolchain/pull/352

PR #350, opened 2022-12-06
async transports tests

menduz:


packages/@dcl/ecs/src/systems/crdt/index.ts
```diff
@@ -105,7 +104,7 @@ export function crdtSceneSystem(
         })
       } else {
         // Add message to transport queue to be processed by others transports
-        broadcastMessages.push(message)
+        // broadcastMessages.push(message)
```
menduz, line 107: This one has no tests. and increments _a lot_ the amount of CRDT messages in comparision with main branch

PR #357, opened 2022-12-13
UiInput - UiDropdown - UiBackground - PointerEvents 

gonpombo8:
https://github.com/decentraland/protocol/pull/93

## Why we need this refactor ?
We want to reuse the same PointerEvents components that we are currently using with the 3D objects.
And for the UI we need to explicit which entity is going to have PointerEvents, so the naming pointerEventFeedback is not more valid for this cases.
Also it's more intuitive that a Entity needs a PointerEvent component to be able to have a PointerEventResult, even if its not the case for unity renderer because of the raycast calculating this for all the entities.


packages/@dcl/ecs/package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@dcl/crdt": "file:../crdt",
     "@dcl/js-runtime": "file:../js-runtime",
-    "@dcl/protocol": "^1.0.0-3603890942.commit-44633cf"
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-3686058926.commit-d398053.tgz"
```
leanmendoza, line 32: to change before merge

packages/@dcl/react-ecs/src/reconciler/index.ts
```diff
@@ -116,18 +96,23 @@ export function createReconciler(
         update.props as EventSystemCallback,
         {
           button: InputAction.IA_POINTER,
-          hoverText: ''
+          hoverText: '@dcl/react-ecs/onPointerDown'
```
leanmendoza, line 99: is this string necessary?

leanmendoza (approved):
LGTM!

packages/@dcl/react-ecs/src/reconciler/index.ts
gonpombo8: yep, this creates the PointerEvent component in the background (PointerHoverFeedback).

menduz: why are we serializing and sending that string though?

gonpombo8: https://github.com/decentraland/js-sdk-toolchain/pull/357/commits/addd2f1d416bf0556909e57f53bb0d7f97d01e7b#diff-cb93e4e4437dc40ff0f226b8a5587deaa008fe1926da8b250fec3935ceaf0a53R108-R110

menduz (approved):

PR #360, opened 2022-12-21
refactor: entity as number with versions

leanmendoza:
### Also, this is the first stage for the migration to an Entity with versions.
- [ ] **Stage 1: Generate and reuse entities at engine level**
- [ ] Stage 2: Integrate the mask with CRDT and add G-set
- [ ] Stage 3: Optimize the component sets to not leak memory
- [ ] Stage 4: Idem as 3) but in the renderer

Discussion in https://github.com/decentraland/sdk/issues/500


gonpombo8 (approved):

test/react-ecs/tree.spec.tsx
```diff
@@ -541,9 +542,12 @@ describe('RectEcs UI ✨', () => {
     // Add an element at the beginning of the array
     uiEntities.unshift({ id: 8, value: 8 })
     await engine.update(1)
-    const newEntity = ((entityD as number) + 1) as Entity
+    const newEntity = getUi(entityA).rightOf
```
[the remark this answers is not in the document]
leanmendoza: Yes, it's reusing an old entity. The assertion also checks the match of other properties, at `rightOf` property level you're right, it isn't testing anything. Should I remove it?

packages/@dcl/playground-assets/etc/playground-assets.api.md
```diff
@@ -467,8 +472,17 @@ export type EngineEvent<T extends IEventNames = IEventNames, V = IEvents[T]> = {
...
 };
 
-// @public (undocumented)
-export type Entity = unknown;
+// Warning: (tsdoc-escape-greater-than) The ">" character should be escaped using a backslash to avoid confusion with an HTML tag
+// Warning: (tsdoc-escape-greater-than) The ">" character should be escaped using a backslash to avoid confusion with an HTML tag
+// Warning: (tsdoc-escape-greater-than) The ">" character should be escaped using a backslash to avoid confusion with an HTML tag
+// Warning: (tsdoc-escape-greater-than) The ">" character should be escaped using a backslash to avoid confusion with an HTML tag
+// Warning: (tsdoc-malformed-html-name) Invalid HTML element: Expecting an HTML name
+// Warning: (tsdoc-malformed-html-name) Invalid HTML element: A space is not allowed here
```
kuruk-mm, line 480: Should we mute these warning?

packages/@dcl/ecs/src/engine/entity.ts
```diff
@@ -1,77 +1,109 @@
-import { EntityUtils } from './entity-utils'
+/**
+ * @public It only defines the type explicitly, no effects.
+ */
+export type uint32 = number
+export const MAX_U16 = 0xffff
+export const MASK_UPPER_16_ON_32 = 0xffff0000
+
+export const AMOUNT_VERSION_AVAILABLE = MAX_U16 + 1
```
kuruk-mm, line 8: Can we add /* @public */ to each line?

test/ecs/entity.spec.ts
```diff
@@ -48,27 +40,35 @@ describe('Entity container', () => {
...
       entityContainer.generateEntity()
     }
-
     expect(() => {
       entityContainer.generateEntity()
     }).toThrowError()
 
-    changeRange(realValue)
+    const randomEntityNumber = (32e3 +
+      Math.round(Math.random() * 32e3)) as Entity
```
kuruk-mm, line 53: I don't like to have random tests. In my opinion; the tests should be determistic. If you set a fixed seed for the Math.random to repeat the same pattern should be ok.

But this is just a suggestion. Not required.

kuruk-mm (approved):
Great work!

packages/@dcl/playground-assets/etc/playground-assets.api.md
menduz: We should fix the underlying cause, not mute

PR #362, opened 2022-12-27
feat: crdt with gset

leanmendoza:
## Summary
1. `@dcl/crdt`: This adds the GSet in the CRDT library and state resolution of components with deleted entities
2. `@dcl/ecs`: (Serialization) Add a new message `DELETE_ENTITY` with payload the `entityId` to be deleted.
3. WireMessage was changed to `CrdtMessage`, the all `CrdtProtocol` was moved to `serialization/crdt`.

## @dcl/crdt
As the Entity is defined by its `EntityId`, we refer to it with the pair `[entityNumber, entityVersion]` without setting how the compound number is made.

### Grow-only set
The Grow-only Set implemented is done with a specific optimization for this case. The premise is: `as entity-version is an incremental positive number for each entity number, its existence implies that all lower versions were previously deleted.`

In an ordinary Grow-only set, we could add `[1, 2]`, `[1, 3]` with `[entityNumber, entityVersion]`. In the implemented, we call the function `addTo`, for example, addTo(`[1, 6]`) which means `[1,0]`, `[1,1]`, ....., `[1,6]` will be in the set.

Another consequence of this version: the function `add` isn't defined.

### Conflict-resolution
The introduced grow-only set functionality allows us to optimize the memory of the state. All deleted entities, clean up the state that previously filled. If we receive a message with a component update of deleted entities, we ignore it.

The function `processMessage` is modified from returning a message to returning a result enum. The resulting enum is a more descriptive value, and we can rebuild the message that we need in the function of the result. Each result value is well-defined and documented with a reason and whether a change was impacted in the state.


packages/@dcl/crdt/src/index.ts
```diff
@@ -1,38 +1,54 @@
...
-    }
-
-    for (let i = 0; i < a.byteLength; i++) {
-      if (a[i] !== b[i]) {
-        return false
+    const n = a.byteLength > b.byteLength ? b.byteLength : a.byteLength
+    for (let i = 0; i < n; i++) {
+      const res = a[i] - b[i]
+      if (res !== 0) {
+        return Math.sign(res)
```
menduz, line 33: For speed, prevent calling functions
```suggestion
        return res > 1 ? 1 : -1
```

menduz (commented):
Overall looks solid and promising, will review again once test coverage is reached (it may change impl). I'd like to see a snapshot test printing the delete command in the .crdt file

packages/@dcl/crdt/src/index.ts
```diff
@@ -1,38 +1,54 @@
...
     if (a instanceof globalBuffer && b instanceof globalBuffer) {
       /* istanbul ignore next */
-      return (a as any).equals(b)
+      return (a as any).compare(b)
     }
   }
 
-  return false
+  if (typeof a === 'string') {
+    return a.localeCompare(b as string)
```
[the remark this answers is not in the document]
leanmendoza: null checks are at the top of the function

test/ecs/customComponentWithSystems.spec.ts
```diff
@@ -294,37 +302,123 @@ describe('test CRDT flow E2E', () => {
...
+
+      await Promise.all([engineA.update(0), engineB.update(0)])
+      expect(env.connection.interceptedMessages).toMatchObject([])
+
+      expect(int8A.getOrNull(entityA)).toBe(88)
+      expect(int8B.getOrNull(entityA)).toBe(88)
+    })
+  })
+  // describe('conflict resolution case 4', () => {
+  //   it('same as case 3) but with roles inverted', async () => {
```
menduz, line 335: Is there a reason to be commented instead of .skip?

test/ecs/customComponentWithSystems.spec.ts
```diff
@@ -294,37 +302,123 @@ describe('test CRDT flow E2E', () => {
...
           timestamp: 7
         }
       ])
-      env.interceptedMessages.length = 0
-      await Promise.all([engineA.update(0), engineB.update(0)])
-      expect(env.interceptedMessages).toMatchObject([])
+      env.connection.interceptedMessages = []
+    })
+
+    it('now we are receiving the updates from engineA and finally the entity is removed and sync', async () => {
```
menduz, line 385: Thanks, this was super clear

test/ecs/ecs.crdt.spec.ts
```diff
@@ -260,20 +291,188 @@ describe('CRDT tests', () => {
...
-      1,
-      cusutomComponent,
-      buffer
-    )
+    PutComponentOperation.write(entity, 1, cusutomComponent, buffer)
     serverTransport.onmessage!(buffer.toBinary())
     await serverEngine.update(1)
     const crdtState = serverEngine.getCrdtState()
-    const component = crdtState.get(entity as number)?.get(cusutomComponent._id)
+    const component = crdtState.components
```
menduz, line 298: Thanks for changing the order of the internal state

test/ecs/ecs.crdt.spec.ts
```diff
@@ -260,20 +291,188 @@ describe('CRDT tests', () => {
...
+    const {
+      clients: [clientA, clientB, clientC],
+      getCrdtStates
+    } = SandBox.createEngines({ length: 3 })
+
+    const entityA = clientA.engine.addEntity()
+    clientA.Transform.create(entityA, { position: Vector3.One() })
+
+    // before the update, the crdt state is out-to-date
+    expect(checkCrdtStateWithEngine(clientA.engine).freeConflicts).toBe(false)
```
menduz, line 317: What does .freeConflicts mean? Is it "not pending messages to be processed"?

test/ecs/utils.ts
```diff
@@ -34,6 +140,133 @@ export namespace SandBox {
...
+        const N: number = Math.min(
+          length || msgsOutgoing.length,
+          msgsOutgoing.length
+        )
+        if (N === 0) return
+
+        const buffer = createByteBuffer()
+        for (let i = 0; i < N; i++) {
+          const msg = msgsOutgoing.pop()!
+          const offset = buffer.incrementWriteOffset(msg.length)
```
menduz, line 191: buffer.writeBuffer(msg.data, false)

menduz (approved):
Awesome job! I left some ignorable comments

test/ecs/customComponentWithSystems.spec.ts
leanmendoza: nope, thanks for the catch, I'm restoring that lines

PR #366, opened 2022-12-30
Add initial CLI implementation for sdk

nicoecheza:
### How to use it?
* Run `npx sdk init` | `npx sdk help init` | `npx sdk help`

### Developer notes:
**Why there is a new `tsconfig.cli.json`?**
Because for using the **cli** files from the terminal, we need:
* `Common JS`  modules
* the `node` typings imported

and since the current `tsconfig.json` for the sdk does not need `Common JS` modules nor the `node` types (in fact, importing the types => breaks all the `~System` imports, I might just be missing some conf but didn't want to spend more time on it)

Oh, and you might see some weird indentation:
```ts
const listCommandsStr = () =>
  Array.from(commands)
    .map(
      ($) => `
  * npx sdk ${$}`
    )
    .join('')
```

👆 that's prettier doing it's magic


packages/@dcl/sdk/src/cli/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  readdirSync(COMMANDS_PATH)
+    .filter((fileName) => isJSFile(fileName))
+    .map((fileName) => fileName.slice(0, -3))
+)
+
+const listCommandsStr = () =>
+  Array.from(commands)
+    .map(
+      ($) => `
+  * npx sdk ${$}`
```
menduz, line 31: Should it be @dcl/sdk instead? https://www.npmjs.com/package/sdk

```suggestion
  * npx @dcl/sdk ${$}`
```

packages/@dcl/sdk/tsconfig.cli.json
```diff
@@ -0,0 +1,15 @@
+{
```
menduz, line 1: Is it really needed to have two tsconfigs? everything in the src folder is compiled

packages/@dcl/sdk/package.json
```diff
@@ -7,8 +7,11 @@
   "types": "./index.d.ts",
   "scripts": {
     "build": "tsc -p tsconfig.json",
-    "start": "npm run build && tsc -p tsconfig.json --watch"
+    "build:cli": "tsc -p tsconfig.cli.json",
```
menduz, line 10: Make it executable

```suggestion
    "build:cli": "tsc -p tsconfig.cli.json && chmod +x cli/index.js",
```

packages/@dcl/sdk/package.json
```diff
@@ -7,8 +7,11 @@
...
+    "start": "npm run build && tsc -p tsconfig.json --watch",
+    "init": "@dcl/sdk init"
```
menduz, line 12: Does this work?

menduz (changes requested):

menduz (approved):
:shipit:

PR #382, opened 2023-01-12
Add CLI 'build' command

nicoecheza:
Added new command "build", which validates project structure, install deps & runs typescript


packages/@dcl/sdk/cli/commands/build/helpers.ts
```diff
@@ -0,0 +1,101 @@
...
+  TS_CONFIG_JSON: 'tsconfig.json',
+  SCENE_JSON: 'scene.json'
+}
+
+/**
+ * Required "package.json" structure
+ */
+export const REQUIRED_PACKAGE_JSON = {
+  devDependencies: {
+    '@dcl/sdk': ['latest']
```
gonpombo8, line 22: is `latest` mandatory ? Do we check this string or is just some context ? 
Because you can have a pinned version of the sdk

packages/@dcl/sdk/cli/commands/build/helpers.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Returns true if the project follows a valid scene structure
+ */
+export const validateProjectStructure = async (
+  dir: string,
+  fileList: string[]
+): Promise<boolean> => {
+  const files = await readdir(dir)
+  const requiredFiles = new Set(fileList)
+
+  for (const file of files) {
```
gonpombo8, line 41: wont be better to iterate the `requiredFiles` instead of all the files of the project ? 
If you have only const requiredFiles = new Set('package.json'), you iterate this only once instead of checking every file if its equal to package.json

packages/@dcl/sdk/cli/commands/build/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  await buildTypescript({
+    dir,
+    watch: !!options.args['--watch'],
+    production: !!options.args['--production']
+  })
+
+  succeed('Project built successfully!')
+
+  // track stuff...
```
gonpombo8, line 93: what about this ? 
is there a ticket created to do all the analytics stuff?

packages/@dcl/sdk/cli/utils/fs.ts
```diff
@@ -25,12 +25,28 @@ export async function ensureFolder(path: string | string[]): Promise<void> {
...
+  return fs.readFile(path, 'utf8')
+}
+
 /**
  * Check's if directory is empty
  * @param dir Directory to check for emptyness
  */
 export async function isDirectoryEmpty(dir: string): Promise<boolean> {
-  const files = await fs.readdir(dir)
+  const files = await _this.readdir(dir)
```
gonpombo8, line 49: cant we call `readdir` function istead of `_this.` .
```suggestion
  const files = await readdir(dir)
```

packages/@dcl/sdk/cli/utils/spinner.ts
```diff
@@ -1,44 +1,44 @@
...
-  }
-  succeed(message: string) {
-    this.log('succeed', message)
-  }
-}
-
-const spinner = new Spinner()
-
-export default spinner
+// import ora, { Ora } from 'ora'
```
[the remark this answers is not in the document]
nicoecheza: yeah, I left it here because I'm planning on using it in the future, but I can save it for later

packages/@dcl/sdk/cli/commands/build/helpers.ts
nicoecheza: we just have to defined what versions we are going to allow. It's just a placeholder

packages/@dcl/sdk/cli/commands/build/helpers.ts
nicoecheza: it would be actually the same, creating the Set from `await readdir(dir)` implies iterating all the files. I don't think there is a way to avoid that.

packages/@dcl/sdk/cli/commands/build/index.ts
nicoecheza: yeah! https://github.com/decentraland/sdk/issues/529

packages/@dcl/sdk/cli/utils/fs.ts
nicoecheza: not if we want jest to be able to mock it 😛 

https://stackoverflow.com/questions/45111198/how-to-mock-functions-in-the-same-module-using-jest

packages/@dcl/sdk/cli/utils/out-messages.ts
```diff
@@ -0,0 +1,3 @@
+export function generateStringList(strs: string[]): string {
```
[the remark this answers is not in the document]
nicoecheza: ✅

packages/@dcl/sdk/cli/commands/build/helpers.ts
nicoecheza: in fact, I think an improvement would be to just early-return in case all the required files are found before iterating over all the files in the directory. That way we improve the "best-case"  scenario. The worst case sceneario will remain the same (the last required file is the last file on the files array)

nicoecheza: pushed this:
```typescript
for (let i = 0; i < files.length && requiredFiles.size > 0; i++) {
  if (requiredFiles.has(files[i])) {
    requiredFiles.delete(files[i])
  }
}
```

nicoecheza:
closed in favor of https://github.com/decentraland/js-sdk-toolchain/pull/389

PR #384, opened 2023-01-17
feat: auto calculate component numbers and hide them in common cases

menduz:
https://github.com/decentraland/adr/pull/165/


packages/@dcl/ecs/src/engine/input.ts
```diff
@@ -99,15 +99,14 @@ const InternalInputStateSchema = {
...
-const InternalInputStateComponentId = 1500
 const TimestampUpdateSystemPriority = 1 << 20
 const ButtonStateUpdateSystemPriority = 0
 
 export function createInputSystem(engine: IEngine): IInputSystem {
   const PointerEventsResult = components.PointerEventsResult(engine)
   const InternalInputStateComponent = engine.defineComponent(
-    InternalInputStateSchema,
-    InternalInputStateComponentId
+    '@dcl/sdk/InternalInputStateSchema',
```
leanmendoza, line 108: should we use a unique convention like `dcl::sdk::internalInputState` like `core::Transform` ?

test/ecs/component-numbers.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      'int8'
+    ]
+
+    for (const key of testCases) {
+      expect(componentNumberFromName(key)).toBeGreaterThan(0)
+    }
+  })
+})
+
+describe('bytebuffer', () => {
```
leanmendoza, line 38: there is a dedicated `.spec.ts` for bytebuffer, maybe we should move there?

leanmendoza (approved):
:stonks: , great work 😄 , minor comments

packages/@dcl/ecs/src/engine/types.ts
```diff
@@ -99,72 +99,72 @@ export type IEngine = {
...
-   * @param componentId - component number used to identify the component descriptor
+   * @param componentId - component number or name used to identify the component descriptor
    * @returns the component definition, throw an error if it doesn't exist
    * ```ts
    * const StateComponentId = 10023
    * const StateComponent = engine.getComponent(StateComponentId)
    * ```
    */
-  getComponent<T>(componentId: number): ComponentDefinition<T>
+  getComponent<T>(componentId: number | string): ComponentDefinition<T>
```
gonpombo8, line 154: where is the implementation for `getComponent(componentId: string)` ? 
is ok that this can be a string or a number ?

leanmendoza: something in the typescript server is going wrong, right? because tests are passing 🤔

packages/@dcl/ecs/src/runtime/crc.ts
```diff
@@ -0,0 +1,57 @@
...
+  0x54de5729, 0x23d967bf, 0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94,
+  0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d
+])
+
+function _crc32(buf: Uint8Array, previous: number) {
+  let crc = ~~previous ^ -1
+  for (let n = 0; n < buf.length; n++) {
+    crc = CRC_TABLE[(crc ^ buf[n]) & 0xff] ^ (crc >>> 8)
+  }
+  return crc ^ -1
```
nicoecheza, line 52: 🪄

packages/@dcl/ecs/src/engine/types.ts
menduz: I will revert to `number` only, we may add the `| string` soon

packages/@dcl/ecs/src/engine/input.ts
menduz: I think it makes sense to get a convention here, nevertheless, this one lives in the same realm as the ones created and used from the scenes and libraries. It is to all extent an opaque identifier. Is it OK if we change it once we get agreement on the convention and merge it as it is?

PR #386, opened 2023-01-19
Remove Enums from react ecs

gonpombo8:
## What ??
Remove all the enums that were required for creating an UI.

## Why ??
It was a bad devexp to import Enum Types for every component prop .
Also, this way we remove all the YG prefix, because we don't need them anymore. We still are sending this enums to the renderer so we dont need to change anything there. It's just an abstraction layer.
i.e.

```tsx
// Old 
<Label font={Font.SansSerif} uiTransform={{ display: YGDisplay.YGD_FLEX }} />

// Now
<Label font="sans-serif" uiTransform={{ display: 'flex' }} />

### CODE REVIEW

File: packages/@dcl/react-ecs/src/components/Dropdown/index.tsx [hunk 1]
```diff
@@ -1,17 +1,21 @@
...
+  const { textAlign, font, ...otherProps } = props
   return {
     acceptEmpty: false,
     options: [],
     selectedIndex: props.acceptEmpty ? -1 : 0,
     disabled: false,
-    ...props
+    ...otherProps,
+    ...getTextAlign(textAlign),
+    ...getFont(font)
```
nicoecheza, line 18: seems like we are repeating this in a bunch of files, is it possible to have a fn that just receives props and executes this stuff if the value exists?

packages/@dcl/react-ecs/src/components/Label/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  monospace: Font.F_MONOSPACE
+}
+/**
+ * @internal
+ */
+export function getFont(
+  font: UiFont | undefined
+): Record<'font', Font> | undefined {
+  if (!font) return undefined
+  return { font: parseFont[font] }
```
nicoecheza, line 16: just curiosity, why do these `getX` fns return an object? 🤔

packages/@dcl/react-ecs/src/components/uiBackground/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  return !!(props as UiTexture).texture
+}
+
+/**
+ * @internal
+ */
+export function getTexture(
+  props: UiBackgroundProps
+): PBUiBackground['texture'] {
+  if (isTexture(props) && props.texture) {
```
nicoecheza, line 51: `isTexture` is already checking for `.texture` to exist

packages/@dcl/react-ecs/src/components/uiBackground/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    return {
+      tex: {
+        $case: 'texture',
+        texture: parseTexture(props.texture)
+      }
+    }
+  }
+
+  if (isAvatarTexture(props) && props.avatarTexture) {
```
nicoecheza, line 60: same as above

packages/@dcl/react-ecs/src/components/uiTransform/index.ts
```diff
@@ -69,28 +73,34 @@ const defaultUiTransform: PBUiTransform = {
...
+    ...parseSize(props.maxWidth, 'maxWidth'),
+    ...getDisplay(props.display),
+    ...getAlign('alignContent', props.alignContent, YGAlign.YGA_FLEX_START),
+    ...getAlign('alignSelf', props.alignSelf, YGAlign.YGA_FLEX_START),
+    ...getAlign('alignItems', props.alignItems, YGAlign.YGA_FLEX_START),
+    ...getJustify(props.justifyContent),
+    ...getFlexDirection(props.flexDirection),
+    ...getFlexWrap(props.flexWrap),
+    ...getOverflow(props.overflow),
+    ...getPoistionType(props.positionType)
```
nicoecheza, line 104: just thinking about this (probably not for this PR)...

What about an approach where we have a dictionary of css props mapping to their corresponding parsing fn (ex: `{ position: parsePosition, height: parseSize, etc }`) that allows us to create a an executor that basically runs against all the keys in the `props` field and run the proper parser? 
That way, adding a new parser is just creating the function and map it to the dictionary, like "installing" it, and we avoid having to remember to update all this stuff

nicoecheza (changes requested):

packages/@dcl/react-ecs/src/components/uiTransform/index.ts
gonpombo8: i didn't get this one 🤔

packages/@dcl/react-ecs/src/components/uiBackground/utils.ts
nicoecheza: we need to update the type-guard then, so it returns the proper values

```typescript
function isTexture(
  props: UiBackgroundProps
): props is UiBackgroundProps & Required<UiTexture> {
  return !!(props as UiTexture).texture
}
```

packages/@dcl/react-ecs/src/components/uiTransform/index.ts
nicoecheza: something like this? => https://tsplay.dev/mAdRPN

packages/@dcl/react-ecs/src/components/Label/types.ts
```diff
@@ -3,4 +3,27 @@ import { PBUiText } from '@dcl/ecs'
...
+}
+
+/**
+ * @public
+ */
+export type UiFont = 'sans-serif' | 'serif' | 'monospace'
+/**
+ * @public
+ */
+export type TextAlign =
```
menduz, line 20: I think we are going to have trouble with this typings names unless we change the names to very specific ones like "TextAlignType"

menduz (approved):
Awesome job! Only one comment about exporting generic names

leanmendoza (approved):

packages/@dcl/react-ecs/src/components/Label/types.ts
menduz: Or use namespaces

packages/@dcl/react-ecs/src/components/uiTransform/index.ts
gonpombo8: could be.
The only problem that i see is when the value is not defined, if we create the prop then protobuf will serialize that value. So we need to do something like filter undefined values, etc.
It's more readable that way, but its the same that the interface is doing right now. It's too much work for something more readable, maybe in another iteration.

PR #393, opened 2023-01-26
Add @dcl/inspector

nicoecheza:
New `@dcl/packages/inspector` directory, which basically is a CRA. The important files are the ones for the CI, etc


packages/@dcl/inspector/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
gonpombo8, line 3: maybe es2020 ? 
es5 is too old and have to compile/polyfill a lot of things

packages/@dcl/inspector/src/setupTests.ts
```diff
@@ -0,0 +1,5 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom'
```
menduz, line 5: can you ditch CRA in benefit of `estrella`? It will remove hundreds of MBs in deps and will make build times 100x faster

menduz: https://github.com/decentraland/holodeck/blob/main/build.js

nicoecheza: the thing with `estrella` is that I think that it's best suited for `.ts` projects (thinking about CLI 🤔 ) since otherwise we will have to mantain the build against all the file types we will have on the project (svg, png, css, html, etc). CRA does all that stuff for us + it's today's (unfortunately) standard way to create react apps.
If we don't care about that, I will make the changes 👍

menduz: estrella is only for the react project, it is the best package we've found to make the SDK work on react applications. the webviews from the editor, kernel and holodeck use it

cazala: Hey, sorry I just saw this. I would not recommend using `estrella` to make a production build of any webapp, CRA already handles (and optimizes a lot) this for us. We use estrella to bundle tyspecript files and treeshake them, it does that amazingly fast and easily, but outputting an optimized build for a webapp (css, media files, etc), that's a whole different thing. 

We don't care about size of dependencies here (they are not included in the artifact) as much as size/efficiency of the production build, since we are not building a library but a webapp.

cazala (approved):

nicoecheza:
talked with @cazala, and agreed to merge this for now

PR #403, opened 2023-02-01
feat: retrieve commands from multiple entities

leanmendoza:
# What? 
Instead of getting all commands from RootEntity, the renderer would send the commands by a range of entities.

# Why? 
This allows the renderer only sends the new commands and the ECS state would keep the commands from the last N frames.
This is the solution proposed [here](https://github.com/decentraland/sdk/issues/358#issuecomment-1403673267).


packages/@dcl/ecs/src/engine/input.ts
```diff
@@ -119,14 +119,23 @@ export function createInputSystem(engine: IEngine): IInputSystem {
     }
   }
 
+  function* commandIterator() {
+    // TODO: should filter by entity?
+    for (const [entity, value] of engine.getEntitiesWith(PointerEventsResult)) {
+      for (const command of value.commands) {
+        yield command
+      }
```
menduz, line 127:
```suggestion
      yield* value.commands
```

packages/@dcl/ecs/src/engine/input.ts
```diff
@@ -201,11 +207,8 @@ export function createInputSystem(engine: IEngine): IInputSystem {
     pointerEventType: PointerEventType,
     entity?: Entity
   ): PBPointerEventsResult_PointerCommand | null {
-    const component = PointerEventsResult.getOrNull(engine.RootEntity)
-    if (!component) return null
-
     // We search the last pointer Event command sorted by timestamp
-    const command = findLastAction(component.commands, pointerEventType, inputAction, entity)
+    const command = findLastAction(Array.from(commandIterator()), pointerEventType, inputAction, entity)
```
menduz, line 211: It would be nice for `findLastAction` to receive an enumerator to prevent re-allocating and re-iterating the `commandIterator()`

packages/@dcl/ecs/src/engine/input.ts
leanmendoza: thx

PR #405, opened 2023-02-02
Add '--move-out' parametet to init command

nicoecheza:
New `--move-out` parameter for init command to move files out of the templates folders.

EX: 
1. Clone `https://github.com/decentraland/sdk7-scene-template/archive/refs/heads/main.zip`
2. Extract
3. You will now have a folder with the scene files inside (folder name doesn't matter)

If the user provides the `--move-out` parameter, it will move those files out of that folder, if the user doesn't provide it, it will leave the directory structure as it is.

### DIFF
--- a/packages/@dcl/sdk/cli/commands/init/repos.ts
+++ b/packages/@dcl/sdk/cli/commands/init/repos.ts
@@ -1,5 +1,17 @@
-export const REPOS = {
-  'scene-template': 'https://github.com/decentraland/sdk7-scene-template/archive/refs/heads/main.zip'
+type Scene = 'scene-template'
+
+type Repos = {
+  [key in Scene]: {
+    url: string
+    contentFolders: string[]
+  }
 }
 
-export const get = (repo: keyof typeof REPOS) => REPOS[repo]
+const REPOS: Repos = {
+  'scene-template': {
+    url: 'https://github.com/decentraland/sdk7-scene-template/archive/refs/heads/main.zip',
+    contentFolders: ['sdk7-scene-template-main']
+  }
+}
+
+export const get = (scene: Scene): Repos[Scene] => REPOS[scene]
--- a/packages/@dcl/sdk/cli/commands/init/index.ts
+++ b/packages/@dcl/sdk/cli/commands/init/index.ts
@@ -33,7 +33,29 @@ export const main = handler(async function main(options: Options) {
   }
 
   const scene = 'scene-template'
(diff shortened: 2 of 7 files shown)


packages/@dcl/sdk/cli/commands/init/repos.ts
```diff
@@ -1,5 +1,17 @@
...
+  [key in Scene]: {
+    url: string
+    contentFolders: string[]
+  }
+}
+
+const REPOS: Repos = {
+  'scene-template': {
+    url: 'https://github.com/decentraland/sdk7-scene-template/archive/refs/heads/main.zip',
+    contentFolders: ['sdk7-scene-template-main']
```
cazala, line 13: instead of having `type Repo = { url: string, contentFolders: string[] }` why don't we just have each repo be a string like `decentraland/sdk7-scene-template` and then generate the url and content folder from that?

ie:

```ts
function getRepoUrl(repo: string) {
  return `https://github.com/${repo}/archive/refs/heads/main.zip`
}

function getRepoContentFolder(repo: string) {
  const [_organization, repository] = repo.split('/')
  return `${repository}-main`
}
```

That way can we keep it, also there shouldn't be multiple content folders since github wraps the content of the repo in a single directory

menduz (commented):
I think this should be the default behavior. I can't think of any tool that creates a folder inside the folder I'm running the init command

packages/@dcl/sdk/cli/commands/init/repos.ts
nicoecheza: didn't want to get attached to github URL's in case we add other templates from some cdn, etc

nicoecheza:
> I think this should be the default behavior. I can't think of any tool that creates a folder inside the folder I'm running the init command

that's a good point. Do we want to maintain the possibility of not moving the files out of the folder?

menduz:
> > I think this should be the default behavior. I can't think of any tool that creates a folder inside the folder I'm running the init command
> 
> that's a good point. Do we want to maintain the possibility of not moving the files out of the folder?

What do you think about this:

- `dcl-command init` initializes in the current folder
- `dcl-command init --dir xxx` initializes in the xxx folder

Removing the option completely, and keeping --dir to customize the target dir. Because I think that having a `sdk7-test-scene` folder is a bit obscure. Where does that name comes from? Will it be consistent over time? Will it change with a different template? With the --dir thing we can simply not have these questions at all

cazala (approved):

PR #415, opened 2023-02-09
feat: change enum schema to force the final serialized schema

leanmendoza:


test/ecs/serialization.spec.ts
```diff
@@ -229,11 +229,11 @@ describe('Serialization Types', () => {
       Default = 0,
       Red = 2,
       Green = 0x33,
-      Pink = 0xff290323
+      Pink = 0x1f290323
```
menduz, line 232: 0xffffffff didn't work?

leanmendoza: nope, typescript hex representation isn't a 32bit one, so numbers > 0x7fff_ffff are out of range

leanmendoza: But `-1` is a valid number :)

menduz (approved):

PR #421, opened 2023-02-13
Add entity tree to inspector

nicoecheza:
file tree

the icons are temp and just for easier debugging


packages/@dcl/inspector/src/components/tree/Tree.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const handleToggleEdit = (e: React.MouseEvent) => {
+    e.stopPropagation()
+    setEditMode(true)
+  }
+
+  const onChangeEditValue = (newValue: string) => {
+    tree.rename(newValue)
+    setEditMode(false)
+    update()
```
cazala, line 63: this is an issue that i see in all the handlers, we are mutating instead of informing of the change intended to the parent and letting the parent perform the update to the state. 

Having that mutation is what forces you to have this `update()` call to force a re-render, but if instead the component just comunicated the action, then the parent can return a new tree which should in turn render it again organically.

For example, instead of performing the `tree.rename()` above, we should only call a callback informing the id of the node and the new label, like:

```
onRename(node.id, newLabel)
```

Then the parent is who decides how to handle that change and generate a new tree, otherwise this component can only work with the `Tree` class, which is not where the state will live.

packages/@dcl/inspector/src/components/tree/Tree.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Input } from './input'
+import { Controls } from './controls'
+import { Icon } from './icon'
+
+interface Props {
+  value: Tree
+}
+
+interface TreeProps {
+  tree: Tree
```
cazala, line 16: The tree here is an instance of a class, where it should be a static type. 

For example:

```ts
type Tree = {
  id: string
  value: string
  label: string
  children: Tree[]
}
```

That's so if we want to make a change to it, we should create a new tree.

packages/@dcl/inspector/src/components/tree/Tree.tsx
```diff
@@ -0,0 +1,126 @@
...
+  update: () => void
```
cazala, line 17: Instead of an `update` callback that forces a re-render, we should comunicate the desired changes through callbacks, for example:

```ts
onRename: (id: string, newName: string) => void,
onSetParent: (id: string, newParent: string | null) => void
// and so on...
```

So then the parent can decide how to perform those changes and provide a new tree.

packages/@dcl/inspector/src/components/tree/Tree.tsx
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const getEditModeStyles = (active: boolean) => ({ display: active ? 'none' : 'block' })
+const getExpandStyles = (active: boolean) => ({ height: active ? 'auto' : '0', overflow: 'hidden', display: 'block' })
+
+const MemoTree = React.memo(TreeComponent)
+
+function TreeComponent({ tree, update }: TreeProps) {
+  const { value, childs, parent, expanded = false, type } = tree
```
cazala, line 26: The `expanded` value should be stored in the component's state, since it's not part of the data itself from the tree

packages/@dcl/inspector/src/utils/tree.ts
```diff
@@ -0,0 +1,85 @@
+export type TreeType = 'directory' | 'file'
```
cazala, line 1: There's no need to have different types of nodes in the tree to know if they can be expanded or collapsed, this can be known just by telling whether if a node is a leaf or not (ie. if it has children it can be expanded, otherwise not).

cazala (changes requested):

cazala (commented):
Since the state of the entity tree will be handled by the data service, I would recomend not having a class at all to represent the tree within the `TreeComponent`, since that goes against the top-bottom design of react (the components should only the data they receive and inform of changes via callbacks).

On one hand, we need to separate the data from the UI state, like the `expanded` boolean, that needs to leave on the component.

For the data, I would suggest using a flat structure (ie. an `Array`) since it's easier to work with (rename a node, reparent it, etc), and then compute the tree from it.

This is an example storing everthing on the `App` component state, but then we can replace that with the comunication with the Data Service:

```ts
// App.ts

// Types for flat node and tree structure
type Node = {
  id: string
  value: any
  label: string
  parent: string | null
}

// This is the actual type the TreeComponent expects for the props, the one above is just to store it and modify it easily
type Tree = {
  id: string
  value: any
  label: string
  children: Tree[]
}

// Store nodes in flat array
const [nodes, setNodes] = useState<Node[]>([
  {
    id: '0',
    value: 'someNode',
    label: 'This is a node',
    parent: null
  },
  {
    id: '1',
    value: 'someSibling',
    label: 'This is a sibling node',
    parent: null
  },
  {
    id: '2',
    value: 'someChild',
    label: 'This is a child of the first node',
    parent: '0'
  }
])

// Compute nodes into tree, memoize it
const toTree = (nodes: Node[], parent: string | null = null): Tree[] => {
  return nodes
    .filter((node) => node.parent === parent)
    .map<Tree>((node) => ({ ...node, children: toTree(nodes, node.id) }))
}

const tree = useMemo(() => toTree(nodes), [nodes])

// Handlers for rename, reparent
const handleRename = useCallback(
  (id: string, newLabel: string) =>
    setNodes(
      nodes.map((node) =>
        node.id === id ? { ...node, label: newLabel } : node
      )
    ),
  [setNodes]
)

const handleSetParent = useCallback(
  (id: string, newParent: string | null) =>
    setNodes(
      nodes.map((node) =>
        node.id === id ? { ...node, parent: newParent } : node
      )
    ),
  [setNodes]
)

// ... render tree component
<TreeComponent tree={tree} onRename={handleRename} onSetParent={handleSetParent} />

```

cazala (approved):

PR #426, opened 2023-02-16
Fix MessageBus

gonpombo8:
## What ?? 
Fix message bus & onCommsMessage observable.

Closes https://github.com/decentraland/sdk/issues/572

### DIFF
--- a/test/__mocks__/~system/EngineApi.js
+++ b/test/__mocks__/~system/EngineApi.js
@@ -1,0 +1,4 @@
+const EngineApi = jest.mock('~system/EngineApi')
+EngineApi.subscribe = async () => ({})
+
+module.exports = EngineApi
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,8 +15,8 @@
     "test/snapshots/package-lock.json": true,
     "packages/@dcl/*/*.tgz": true,
     "packages/@dcl/*/*.js": true,
+    "packages/@dcl/sdk/*.d.ts": true,
     "**/*.js": { "when": "$(basename).ts" },
-    "**/*.d.ts": true
   },
   "[typescript]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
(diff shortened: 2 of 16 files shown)


test/__mocks__/~system/EngineApi.js
```diff
@@ -0,0 +1,4 @@
+const EngineApi = jest.mock('~system/EngineApi')
+EngineApi.subscribe = async () => ({})
+
+module.exports = EngineApi
```
menduz, line 4: you won't need this if you restore the `setSubscribeFunction` function

gonpombo8: since I changed [this line](https://github.com/decentraland/js-sdk-toolchain/pull/426/files#diff-de648aea11bc222f713557aec7af806ebac01c23c0eafcb844d9c2ca56069f52R5) is needed.

gonpombo8: the reason behind this is that when you create a file `observables.ts` in your scene, and run some code there related to observables like
```tsx
onEnterSceneObservable.add(someCallback)
```

Then, this observable was executed with the mocked subscribe function instead of the import { subscribe } from '~system/EngineApi`.
So then, the callback would never be executed because there was no subscribe event.

menduz: changing that line you are unwillingly coupling a side effect that requires mocking a module with the actual implementation.

using the funciton to set the listener prevents you from needing that mock in the first place, by design it doesn't require any side effect to run.

the import of @dcl/sdk is mandatory in scenes to work... I'd like very much to have it decoupled from node's require in this case

The important part of this PR, the actual fix is only this:
Screenshot 2023-02-16 at 13 06 21

gonpombo8: nope, it didn't work with only that line. Only works if you define the observables in the index.ts file.
If you create a separate file, then each observable was executed with the mocked fn

PR #433, opened 2023-02-27
add analytics to sdk-commands

gonpombo8:


packages/@dcl/sdk-commands/src/commands/start/index.ts
```diff
@@ -83,7 +84,8 @@ export async function main(options: Options) {
   const skipBuild = args['--skip-build']
   const watch = !args['--no-watch']
   const enableWeb3 = args['--web3']
-  const baseCoords = { x: 0, y: 0 }
+
+  // TODO: FIX this hardcoded values ?
```
pravusjif, line 88: should this comment remain or changing the old `baseCoords` hardcoded values to be dynamic as you did is good enough?

pravusjif (approved):
LGTM!

packages/@dcl/sdk-commands/src/commands/start/index.ts
gonpombo8: yes ! but there is also a const that is still hardocded `  const hasPortableExperience = false`

PR #435, opened 2023-02-27
Add 'deploy' command

nicoecheza:
Same deploy command we already had in CLI but with a refactored Linker API

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
+++ b/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
@@ -3,16 +3,17 @@ import { EntityType, ChainId, Scene } from '@dcl/schemas'
 import { CatalystClient, ContentClient, DeploymentBuilder } from 'dcl-catalyst-client'
 import { Authenticator } from '@dcl/crypto'
 import { hexToBytes } from 'eth-connect'
-import { ethSign, recoverAddressFromEthSignature } from '@dcl/crypto/dist/crypto'
+import { ethSign } from '@dcl/crypto/dist/crypto'
 
 import { CliComponents } from '../../components'
 import { main as build } from '../build'
 import { IFile, getFiles, getValidSceneJson, validateFilesSizes } from '../../logic/scene-validations'
 import { getArgs } from '../../logic/args'
 import { npmRun } from '../../logic/project-validations'
-import { link, LinkerResponse } from './linker-dapp/api'
+import { runLinkerApp, LinkerResponse } from './linker-dapp/api'
 import { CliError } from '../../logic/error'
 import { printProgressInfo, printSuccess } from '../../logic/beautiful-logs'
+import { createWallet } from '../../logic/account'
 
 interface Options {
   args: typeof args
@@ -35,7 +36,7 @@ export const args = getArgs({
   '--yes': Boolean,
   '--no-browser': Boolean,
   '-b': '--no-browser',
-  '--port': String,
+  '--port': Number,
   '-p': '--port'
 })
 
@@ -64,9 +65,7 @@ export async function main(options: Options) {
   const dir = resolve(process.cwd(), options.args['--dir'] || '.')
   const openBrowser = !options.args['--no-browser']
   const skipBuild = options.args['--skip-build']
-  const port = options.args['--port']
-  const parsedPort = typeof port === 'string' ? parseInt(port, 10) : void 0
-  const linkerPort = parsedPort && Number.isInteger(parsedPort) ? parsedPort : void 0
+  const linkerPort = options.args['--port']
   const { error } = options.components.logger
 
   const comps = { components: options.components }
@@ -157,22 +156,6 @@ interface LinkOptions {
   skipValidations: boolean
 }
 
-const runLinkerApp = async (
-  components: CliComponents,
-  rootCID: string,
-  scene: Scene,
-  files: IFile[],
-  options: LinkOptions
-): Promise<LinkerResponse> => {
-  const { linkerPort, ...opts } = options
-  try {
-    const res = await link(components, scene, files, linkerPort!, rootCID, opts)
-    return res
-  } catch (e) {
-    throw new CliError(`Error running linking app: ${e}`)
-  }
-}
-
 async function getAddressAndSignature(
   components: CliComponents,
   messageToSign: string,
@@ -186,23 +169,6 @@ async function getAddressAndSignature(
     return { signature, address: wallet.address }
   }
 
-  return runLinkerApp(components, messageToSign, scene, files, linkOptions)
-}
-
-function createWallet(privateKey: string) {
-  let length = 64
-
-  if (privateKey.startsWith('0x')) {
-    length = 66
-  }
-
-  if (privateKey.length !== length) {
-    throw new CliError('Addresses should be 64 characters length.')
-  }
-
-  const pk = hexToBytes(privateKey)
-  const msg = Math.random().toString()
-  const signature = ethSign(pk, msg)
-  const address: string = recoverAddressFromEthSignature(signature, msg)
-  return { address, privateKey, publicKey: '0x' }
+  const { linkerPort, ...opts } = linkOptions
+  return runLinkerApp(components, scene, files, linkerPort!, messageToSign, opts)
 }
--- a/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
+++ b/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
@@ -7,7 +7,7 @@ import { ethSign, recoverAddressFromEthSignature } from '@dcl/crypto/dist/crypto
 
 import { CliComponents } from '../../components'
 import { main as build } from '../build'
-import { IFile, getFiles, getSceneJson } from '../../logic/scene-validations'
+import { IFile, getFiles, getValidSceneJson, validateFilesSizes } from '../../logic/scene-validations'
 import { getArgs } from '../../logic/args'
 import { npmRun } from '../../logic/project-validations'
 import { link, LinkerResponse } from './linker-dapp/api'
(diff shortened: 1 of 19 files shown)


packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/api.ts
```diff
@@ -0,0 +1,234 @@
...
+            const { address, signature, chainId } = value
+            printSuccess(cliComponents.logger, `\nContent successfully signed.`, '')
+            logger.info(`Address: ${address}`)
+            logger.info(`Signature: ${signature}`)
+            logger.info(`Network: ${getChainName(chainId!)}`)
+            resolve(value)
+          }
+        }
+
+        const router = setRoutes(cliComponents, finish, files, scene, rootCID, skipValidations)
```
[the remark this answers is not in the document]
menduz: And that way we would have something like this

```ts
const router = setRoutes(..)
const signaturePromise = signaturesRouterHandler(router)

components.server.setContext(components)
components.server.use(router.allowedMethods())
components.server.use(router.middleware())

await startComponents()

// wait for the signature
const signature = await signaturePromise

// deploy
await deployUsingSignature(signature)

await program.stop()
```

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/api.ts
```diff
@@ -0,0 +1,234 @@
...
+  skipValidations: boolean
+) {
+  const router = new Router()
+  const linkerDapp = dirname(require.resolve('@dcl/linker-dapp/package.json'))
+  let deployInfo: DeployInfo = {}
+
+  router.get('/', async () => {
+    return {
+      headers: { 'Content-Type': 'text/html' },
+      body: fs.createReadStream(resolve(linkerDapp, 'index.html'))
```
menduz, line 119: please use `components.fs`

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/catalyst-pointers.ts
```diff
@@ -0,0 +1,70 @@
+import { Entity } from '@dcl/schemas'
+import fetch from 'node-fetch'
```
menduz, line 2: Please use fetch component instead

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/config.ts
```diff
@@ -0,0 +1,109 @@
+import path from 'path'
+import fs from 'fs/promises'
+
+function getUserHome(): string {
+  return process.env[process.platform === 'win32' ? 'USERPROFILE' : 'HOME'] || ''
```
menduz, line 5:
```suggestion
  return require('os').homedir();
```

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/config.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+async function readDCLInfo(): Promise<DCLInfo | null> {
+  const filePath = getDCLInfoPath()
+  try {
+    const file = await fs.readFile(filePath, 'utf-8')
+    return JSON.parse(file) as DCLInfo
+  } catch (e) {
+    return null
+  }
+}
```
menduz, line 52: please use fs component and extract the dclinfo functions to its own file

packages/@dcl/sdk-commands/src/logic/project-files.ts
```diff
@@ -74,3 +75,81 @@ export async function getProjectContentMappings(
...
+  size: number
+}
+
+export const MAX_FILE_SIZE_BYTES = 50 * 1e6 // 50mb
+
+/**
+ * Returns a promise of an array containing all the file paths for the given directory.
+ * @param dir The given directory where to list the file paths.
+ */
+export const getAllFilePaths = async (
```
menduz, line 91: duplicated of `getPublishableFiles`

packages/@dcl/sdk-commands/src/logic/project-files.ts
```diff
@@ -74,3 +75,81 @@ export async function getProjectContentMappings(
...
+  }
+}
+
+/**
+ * Returns a promise of an array of objects containing the path and the content for all the files in the project.
+ * All the paths added to the `.dclignore` file will be excluded from the results.
+ * Windows directory separators are replaced for POSIX separators.
+ * @param ignoreFile The contents of the .dclignore file
+ */
+export const getFiles = async (
```
menduz, line 129: please use `getPublishableFiles().filter(size)` for this validation

packages/@dcl/sdk-commands/src/logic/scene-validations.ts
```diff
@@ -14,6 +15,17 @@ export function getSceneFilePath(projectRoot: string): string {
...
+/**
+ * Returns `scene.json` file contents as JSON.
+ * @param components The necessary components to read the file
+ * @param projectRoot The path to the directory containing the scene file.
+ */
+export async function getSceneJSON(components: Pick<CliComponents, 'fs'>, projectRoot: string) {
+  const sceneJsonRaw = await components.fs.readFile(getSceneFilePath(projectRoot), 'utf8')
+  const sceneJson = JSON.parse(sceneJsonRaw) as Scene
+  return sceneJson
+}
```
menduz, line 27: exporting this function is unnecessary, use or rename `validateSceneJson` instead

packages/@dcl/sdk-commands/src/logic/project-files.ts
```diff
@@ -74,3 +75,81 @@ export async function getProjectContentMappings(
...
   return ret
 }
+
+export interface IFile {
+  path: string
+  content: Buffer
+  size: number
+}
+
+export const MAX_FILE_SIZE_BYTES = 50 * 1e6 // 50mb
```
menduz, line 85: move this const and logic to "scene-validations.ts"

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/api.ts
```diff
@@ -0,0 +1,234 @@
...
+      await open(`${url}?${params}`)
+    } catch (e) {
+      logger.error(`Unable to open browser automatically`)
+    }
+  }, 5000)
+
+  logger.info(`Signing app ready at ${url}`)
+}
+
+function setRoutes(
```
menduz, line 104: please move this to a `deploy/routes.ts` file for readability of this file

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      printProgressInfo(options.components.logger, response.message)
+    }
+    printSuccess(options.components.logger, 'Content uploaded', sceneUrl)
+  } catch (e: any) {
+    error('Could not upload content:')
+    console.log(e)
+  }
+}
+
+const getCatalyst = async (target?: string, targetContent?: string) => {
```
menduz, line 141: Let's converge towards function declarations instead of consts for functions. Also it maybe should have its own file somewhere else

```suggestion
export async function getCatalyst(target?: string, targetContent?: string) => {
```

menduz (changes requested):

packages/@dcl/sdk-commands/src/logic/project-files.ts
nicoecheza: the thing is, just a `filter` won't do it since we also need to map the result into `{ path: string, content: buffer, size: number }`. We could `reduce` them into `Promise.all` but in my opinion, that's uglier than using a for loop

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  options: LinkOptions
+): Promise<LinkerResponse> => {
+  const { linkerPort, ...opts } = options
+  try {
+    const res = await link(components, scene, files, linkerPort!, rootCID, opts)
+    return res
+  } catch (e) {
+    throw new CliError(`Error running linking app: ${e}`)
+  }
+}
```
menduz, line 174: can this logic be moved into link? It doesn't make much sense to have this wrapping function.

Also I'd find a more expressive name for `link` function. It is hard to tell what it does

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  if (privateKey.length !== length) {
+    throw new CliError('Addresses should be 64 characters length.')
+  }
+
+  const pk = hexToBytes(privateKey)
+  const msg = Math.random().toString()
+  const signature = ethSign(pk, msg)
+  const address: string = recoverAddressFromEthSignature(signature, msg)
+  return { address, privateKey, publicKey: '0x' }
+}
```
menduz, line 208: Maybe move this function to logic/account?

packages/@dcl/sdk-commands/src/logic/get-free-port.ts
```diff
@@ -1,13 +1,13 @@
 import portfinder from 'portfinder'
 
-export async function previewPort() {
-  let resolvedPort = 0
+export async function getPort(port = 0, failoverPort = 2044) {
```
menduz, line 3: Prefer no optionals or change all the occurrences. There is still logic like `args.port || await getPort()` which is inconsistent.

```suggestion
export async function getPort() {
```

packages/@dcl/sdk-commands/src/commands/deploy/linker-dapp/api.ts
```diff
@@ -0,0 +1,129 @@
...
+    parcels,
+    rootCID,
+    landRegistry: LANDRegistry,
+    estateRegistry: EstateRegistry,
+    debug: !!process.env.DEBUG,
+    title: display?.title,
+    description: display?.description,
+    skipValidations
+  }
+}
```
menduz, line 129: this datatype seems very fragile. we may want to refactor it after merging

menduz (approved):

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
nicoecheza: just kept it for consistency with the old CLI, I think we don't even need to move it to the `link` fn, we can just remove it completely and expect errors to bubble up and it will be the same

PR #438, opened 2023-02-28
add packageversion for analytics old dashboards

gonpombo8:
remove the src folder when packing the sdk-commnads pacakage

### DIFF
--- a/packages/@dcl/sdk-commands/src/components/analytics.ts
+++ b/packages/@dcl/sdk-commands/src/components/analytics.ts
@@ -77,7 +77,11 @@ export async function createAnalyticsComponent({
           cliVersion: await dclInfoConfig.getVersion(),
           isCI: dclInfoConfig.isCI(),
           isEditor: dclInfoConfig.isEditor(),
-          devId: userId
+          devId: userId,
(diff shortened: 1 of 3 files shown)


nicoecheza (approved):

packages/@dcl/sdk-commands/src/components/analytics.ts
```diff
@@ -77,7 +77,11 @@ export async function createAnalyticsComponent({
           cliVersion: await dclInfoConfig.getVersion(),
           isCI: dclInfoConfig.isCI(),
           isEditor: dclInfoConfig.isEditor(),
-          devId: userId
+          devId: userId,
+          ecs: {
```
menduz, line 81:
```suggestion
          sdk: {
```

menduz (changes requested):
Maybe SDK is better name than ECS

packages/@dcl/sdk-commands/src/components/analytics.ts
gonpombo8: compatibility with old dashboards

PR #439, opened 2023-02-28
Add tracking analytics to 'deploy' command

nicoecheza:
As required in https://github.com/decentraland/sdk/issues/529

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
+++ b/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
@@ -7,13 +7,14 @@ import { ethSign } from '@dcl/crypto/dist/crypto'
 
 import { CliComponents } from '../../components'
 import { main as build } from '../build'
-import { IFile, getFiles, getValidSceneJson, validateFilesSizes } from '../../logic/scene-validations'
+import { IFile, getBaseCoords, getFiles, getValidSceneJson, validateFilesSizes } from '../../logic/scene-validations'
 import { getArgs } from '../../logic/args'
 import { npmRun } from '../../logic/project-validations'
 import { runLinkerApp, LinkerResponse } from './linker-dapp/api'
 import { CliError } from '../../logic/error'
 import { printProgressInfo, printSuccess } from '../../logic/beautiful-logs'
 import { createWallet } from '../../logic/account'
+import { b64HashingFunction } from '../../logic/project-files'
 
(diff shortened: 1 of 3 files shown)


packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -62,31 +63,38 @@ export function help() {
...
-  await build({ args: { '--dir': dir }, ...comps })
+  await options.components.analytics.track('Scene deploy started', analyticsOpts)
 
-  if (options.args['--target'] && options.args['--target-content']) {
-    throw new CliError(`You can't set both the 'target' and 'target-content' arguments.`)
+  if (!skipBuild) {
+    await npmRun(projectRoot, 'build')
   }
 
+  await build({ args: { '--dir': projectRoot }, ...comps })
```
gonpombo8, line 90: this build command isn't the same as `npmRun(projectRoot, 'build')` ?

nicoecheza: it might be or it might be not, it depends on the content creator preferences. We are also doing these steps in the `start` command. If the user has some custom build step, we want to run it for him + ensure we run our own build also

gonpombo8: mmm but why ? If i use the command skip build i expect that the `build` command is not going to be called 🤔

gonpombo8:
```
      --skip-build              Skip build before deploy
```

nicoecheza: we can do:
```ts
if (!skipBuild) {
  await npmRun(projectRoot, 'build')
  await build({ args: { '--dir': projectRoot }, ...comps })
}
```

but I think we should leave both just in case the content creator customizes `npm run build` .

what do you think?

gonpombo8: if you didn't change your build script then it will be executed twice.
My thoughts about this is that the deploy command should run always npm run build, unless you add the `skip-build` param. This way its easy to understand what is going to do and what is expected.
Then if you update the script for npm run build is up to you. Its like ejecting the react-scripts, we can't do anything there.

gonpombo8 (approved):

PR #443, opened 2023-03-03
feat: asset catalog

nicoecheza:
closes https://github.com/decentraland/sdk/issues/615


cazala (commented):
Hello @nicoecheza, a few things that I noticed see from testing:

1. The rendering of the selected entity seems to be broken: even though I can select entities on the tree and they show on the tree as selected (with the `@`) they are not always rendered with the highlight in the 3D preview. Also clicking on the cubes in the 3D preview does not always highlight the cube as selected.

https://user-images.githubusercontent.com/2781777/223206767-97792038-bcfb-4faf-8884-2da135038551.mp4

2. The gizmos seem to have disappeared completely 
3. The styles of the catalog when it's showing the different packs looks broken:
Screenshot 2023-03-06 at 15 53 42

4. We should not render empty categories:

Screenshot 2023-03-06 at 16 11 42

nicoecheza:
> Hello @nicoecheza, a few things that I noticed see from testing:
> 
> 1. The rendering of the selected entity seems to be broken: even though I can select entities on the tree and they show on the tree as selected (with the `@`) they are not always rendered with the highlight in the 3D preview. Also clicking on the cubes in the 3D preview does not always highlight the cube as selected.
> 
> 2. The gizmos seem to have disappeared completely
> 3. The styles of the catalog when it's showing the different packs looks broken:
> 
> Screenshot 2023-03-06 at 15 53 42
> 
> 4. We should not render empty categories:
> 
> Screenshot 2023-03-06 at 16 11 42


1 & 2 got broken when removing these lines from the gizmos manager (which are actually duplicated with inverted values 🤷 ):
Screen Shot 2023-03-06 at 16 25 48

just removed the commit and it works again (idk)

3 & 4 fixed 👍

packages/@dcl/inspector/src/components/AssetsCatalog/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  updated_at: string
+}
+
+export interface ThemeProps {
+  onClick: (value: ITheme) => void
+  value: ITheme
+}
+
+export interface CategoriesProps {
+  goBack: () => void
```
cazala, line 46: For the sake of consistency with the standards, would you rename this to `onBack` or `onGoBack`? We usually name the callbacks always as `onSomething`.

cazala (approved):

PR #445, opened 2023-03-08
feat: entity inspector

nicoecheza:


cazala:
I few issues I found:

1) The changes in babylon are not shown in react: selecting an entity on the renderer does not select it in the tree, modifying the transforms by using the gizmos does not change the values in the inputs:

https://user-images.githubusercontent.com/2781777/224346821-867575bb-7f28-4785-bd87-2f7ed006a09f.mp4

This is due to the condition of `if (entitiesLength === numberOfEntities) return null` that avoids the update of the engine. It's true that we can't re-render react every 16ms, but also we can't rely on the amount of entities changing to do it, since changes in the components should also trigger a re-render. We should use the `onChangeFunction` of the `engine` created in the `createInspectorEngine` function. We could create a hook that reacts to changes on particular entities or particular components to trigger re-renders on react.

2) The rotation is shown as X,Y,Z but in reality it is a Quaterion, so the values are X,Y,Z,W. The quaternion values are not relevant for the user, so we must convert them to Euler before showint them to the user (we can do so with `quaternion.toEurlerAngles() // { x, y, z }` and then when the user makes changes to the values, we need to convert the x,y,z eurler angler into a quaternion before updating the Transform: `Quaternion.FromEulerAngler(x, y, z) // { x, y, z, w }`

nicoecheza:
> I few issues I found:
> 
> 1. The changes in babylon are not shown in react: selecting an entity on the renderer does not select it in the tree, modifying the transforms by using the gizmos does not change the values in the inputs:
> 
> This is due to the condition of `if (entitiesLength === numberOfEntities) return null` that avoids the update of the engine. It's true that we can't re-render react every 16ms, but also we can't rely on the amount of entities changing to do it, since changes in the components should also trigger a re-render. We should use the `onChangeFunction` of the `engine` created in the `createInspectorEngine` function. We could create a hook that reacts to changes on particular entities or particular components to trigger re-renders on react.
> 
> 2. The rotation is shown as X,Y,Z but in reality it is a Quaterion, so the values are X,Y,Z,W. The quaternion values are not relevant for the user, so we must convert them to Euler before showint them to the user (we can do so with `quaternion.toEurlerAngles() // { x, y, z }` and then when the user makes changes to the values, we need to convert the x,y,z eurler angler into a quaternion before updating the Transform: `Quaternion.FromEulerAngler(x, y, z) // { x, y, z, w }`

1. Agreed with juanca, this is an issue with the 'hack' we implemented on the tree updates. We should first fix that in order to merge this PR
2. Fixed 👍

cazala (approved):

PR #446, opened 2023-03-09
refactor: change to serialization to little endian

leanmendoza:


eordano (approved):
Should this lead to a runtime version bump? from 7 to 8?

PR #447, opened 2023-03-09
feat: composite instancing

leanmendoza:
# What?
The very first approach for composite instantiation in runtime. This also introduces the auto-instance of the composite with id `main`.

The two concepts introduced here are:
- Composite
    - no matter how is serialized, is described by the next type:
```ts
type Composite = {
  id: string
  components: {
    name: string
    schema?: JsonSchemaExtended
    data: Map<Entity, unknown>
  }[]
}
```

- CompositeProvider
```ts
type CompositeProvider = {
  getCompositeOrNull: (id: string) => Composite | null
}
```

The `CompositeProvider` suggested is "`FetchContentCompositeProvider`":
- get the contents of the scene and look for all `.composite` and `.composite.json` files which are binary ones and json serialized ones respectively.
- fetch all composite and deserialize their data to get them available for the instantiation

In the `onStart` function:
1. The composite provider `FetchContentCompositeProvider` is created
2. Instance the composite with `main` id if exists.

The instantiation is deep, if the composite to instance has another composite inside then instance them. Recursive instantiation is now allowed.

# Why?
https://github.com/decentraland/sdk/discussions/603

# Example
Here is an example with json composite (asset folder)
https://github.com/decentraland/sdk7-scene-template/pull/17/files

You can clone the repo and checkout the branch `test-composite-branch` to test this branch.

# Next steps
- Support re-reference entity in custom components
- Support binary format for composite to increase the load time, especially in large composites.
- Versioning JSON-composite schema?
- Support overrides 
- Optional?: Have the possibility of picking an `onStart` function without composites feature, to avoid .composite fetching and load time: this would be useful for those creators that don't use composites.


packages/@dcl/ecs/src/engine/index.ts
```diff
@@ -188,7 +192,9 @@ function preEngine(): PreEngine {
     return componentsDefinition.values()
   }
 
-  function removeComponentDefinition(componentId: number) {
+  function removeComponentDefinition(componentIdOrName: number | string) {
+    const componentId =
```
menduz, line 196:
```suggestion
    if (sealed) throw new Error('Engine is already sealed. No components can be removed at this stage')
    const componentId =
```

packages/@dcl/ecs/src/composite/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  id: string
+  components: {
+    name: string
+    schema?: JsonSchemaExtended
+    data: Map<Entity, unknown>
+  }[]
+}
+
+export type CompositeProvider = {
+  getCompositeOrNull: (id: string) => Composite | null
```
menduz, line 14: If loading composites from the deployed scene is going to be a thing, I recommend to converge towards a Promise from moment zero
```suggestion
  getCompositeOrNull: (id: string) => Promise<Composite> | null
```

packages/@dcl/sdk/src/index.ts
```diff
@@ -18,6 +19,12 @@ export async function onUpdate(deltaTime: number) {
...
 export async function onStart() {
+  const compositeProvider = await createContentFetchCompositeProvider()
+  const mainComposite = compositeProvider.getCompositeOrNull('main')
+  if (mainComposite) {
+    instanceComposite(engine, mainComposite, () => engine.addEntity(), compositeProvider)
+  }
+
   await engine.seal()
 
   const response = await crdtGetState({ data: new Uint8Array() })
```
menduz, line 30: The composite instantiation should only happen if the renderer does not have any "scene owned entity" to report.

https://github.com/decentraland/protocol/blob/d0a21d73e4d50e6d94f161269c021688ef719083/proto/decentraland/kernel/apis/engine_api.proto#L187

```suggestion
  const compositeProvider = await createContentFetchCompositeProvider()

  await engine.seal()

  const response = await crdtGetState({ data: new Uint8Array() })
  
  if (!response.hasEntities) {
    const mainComposite = compositeProvider.getCompositeOrNull('main')
    if (mainComposite) {
      instanceComposite(engine, mainComposite, () => engine.addEntity(), compositeProvider)
    }
  }
```

leanmendoza: I'm temporarily not adding this because first, we need to discriminate the previously stashed data from the renderer.
Now, `hasEntities` is always true because of the initialization of `PlayerEntity` and `CameraEntity`.
Also, the renderer is not storing the non-core components (custom components)
cc @pbosio

nicoecheza (changes requested):

test/snapshots/ui.ts.crdt
```diff
@@ -1,20 +1,21 @@
-SCENE_COMPILED_JS_SIZE_PROD=292.16k bytes
+SCENE_COMPILED_JS_SIZE_PROD=297.91k bytes
```
menduz, line 1: Good enough delta

menduz (approved):

packages/@dcl/ecs/src/composite/components.ts
```diff
@@ -0,0 +1,37 @@
...
+    src: Entity
+    dest: Entity
+  }[]
+}
+
+/**
+ * @public
+ * @deprecated composite is not being supported so far, please do not use this feature
+ */
+export function CompositeRoot(engine: IEngine): LastWriteWinElementSetComponentDefinition<CompositeRootType> {
```
gonpombo8, line 21:
```suggestion
export function getCompositeRoot(engine: IEngine): LastWriteWinElementSetComponentDefinition<CompositeRootType> {
```

packages/@dcl/ecs/src/composite/instance.ts
```diff
@@ -0,0 +1,155 @@
...
+  // If there is no `rootEntity` passed, we assign one from `getNextAvailableEntity`
+  const compositeRootEntity = rootEntity ?? getCompositeEntity(0)
+  if (rootEntity) {
+    mappedEntities.set(0 as Entity, rootEntity)
+  }
+
+  // ## 2 ##
+  // If there are more composite inside this one, we instance first.
+  // => This is not only a copy, we need to instance.
+  // => Other reason for composite instanciation first is the overides. If we instance first and then load the component, the override
```
gonpombo8, line 70: unfinished comment

packages/@dcl/ecs/src/composite/instance.ts
```diff
@@ -0,0 +1,155 @@
...
+    // ## 3a ##
+    // We find the component definition
+    let componentDefinition
+    const existingComponentDefinition = engine.getComponentOrNull(component.name)
+
+    if (!existingComponentDefinition) {
+      if (component.schema) {
+        componentDefinition = engine.defineComponentFromSchema(component.name, Schemas.fromJson(component.schema))
+      } else if (component.name.startsWith('core::')) {
+        if (Object.keys(componentDefinitionByName).includes(component.name)) {
```
gonpombo8, line 112:
```suggestion
        if (componentDefinitionByName[component.name]) {
```

packages/@dcl/ecs/src/composite/instance.ts
```diff
@@ -0,0 +1,155 @@
...
+      // ## 3c ##
+      // All entities referenced in the composite probably has a different resolved EntityNumber
+      // We'll know with the mappedEntityes
+      if (componentDefinition.componentId === TransformComponentNumber) {
+        const transform = componentValue as TransformType
+        if (transform.parent) {
+          transform.parent = getCompositeEntity(transform.parent)
+        }
+      } else {
+        // TODO: with static reflection, look for `Schema.Entity` in custom components
```
gonpombo8, line 139: nice catch !

packages/@dcl/ecs/src/composite/serialization.ts
```diff
@@ -0,0 +1,46 @@
+import { Entity } from '../engine/entity'
+import { Composite } from './types'
+
+/**
+ * @public
+ * @deprecated composite is not being supported so far, please do not use this feature
+ */
+export function compositeFromJson(jsonComposite: any): Composite {
```
gonpombo8, line 8:
```ts
type CompositeJson  = {
  id?: OriginalCompoisteType['id'] | unknown
  components?: OriginalCompoisteType['components'] | unknown
}
```
```suggestion
export function compositeFromJson(jsonComposite: CompositeJson: Composite {
```
in case we have a typo or anything in the future :panic:

packages/@dcl/sdk/src/composite-provider.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        return null
+      }
+    } catch (err) {
+      console.error(`Error loading composite ${compositeUrl}: ${(err as any).toString()}`)
+      return null
+    }
+  })
+
+  const composites = (await Promise.all(compositePromises)).filter((item) => item !== null) as Composite[]
```
gonpombo8, line 39:
```suggestion
  const composites = (await Promise.all(compositePromises)).filter((item) => !!item) as Composite[]
```

packages/@dcl/sdk/src/index.ts
```diff
@@ -18,6 +19,12 @@ export async function onUpdate(deltaTime: number) {
  * Function that is called before the first update and after the evaluation of the code.
  */
 export async function onStart() {
+  const compositeProvider = await createContentFetchCompositeProvider()
```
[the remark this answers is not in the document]
leanmendoza: done in https://github.com/decentraland/js-sdk-toolchain/pull/447/commits/a1fb2ac9f5d1279ed38883cbeea212b74923d226
see how some bundle-size (like cube.ts) decrease drastically, and the `composite.ts` snapshot is the only which imports the `~system/Scene` because it needs it for fetch content mappings
TLDR, now to add this feature, instead of `export * from '@dcl/sdk'` in your scene, you put `export * from '@dcl/sdk/with-composite'`

packages/@dcl/ecs/src/composite/serialization.ts
leanmendoza: this will be removed with another composite serialization

PR #450, opened 2023-03-09
fix: push identical pointer events (leak memory)

leanmendoza:


packages/@dcl/ecs/src/systems/events.ts
```diff
@@ -99,8 +99,23 @@ export function createPointerEventSystem(engine: IEngine, inputSystem: IInputSys
...
     if (opts.hoverText || opts.showFeedback) {
-      const pointerEvent = PointerEvents.getMutableOrNull(entity) || PointerEvents.create(entity)
+      const existingPointerEvent = PointerEvents.getMutableOrNull(entity)
+      //
+      if (existingPointerEvent) {
+        const alreadyExist = existingPointerEvent.pointerEvents.find(
+          (item) =>
+            item.eventType === type &&
+            item.eventInfo &&
+            item.eventInfo.button === opts.button &&
```
menduz, line 109: I think we could only find by button, then mutate it
```suggestion
            item.eventInfo.button === opts.button &&
```

leanmendoza:
fix here https://github.com/decentraland/js-sdk-toolchain/pull/451

PR #451, opened 2023-03-10
fix leak of memory in react-ecs

gonpombo8:


leanmendoza (approved):
Tested and there is no more leak :)

pravusjif (approved):
Tested without problems as well, great job!

PR #452, opened 2023-03-10
fix: gltf container and other issues

cazala:
This PR fixes a few issues:
- The mappings are now included in the scene definition, so they are available to be resolved by the scene context
- The GLTFs would not be rendered due to a void environment texture: that's been fixed now
- The convertion from mouse X,Y into the scene's X,Y,Z now is being done correctly (the items show up where you drop them)
- There was a reparenting issue where after placing an entity within another, you would not be able to move it into another entity again. This was due to an undefined not being handled correctly in the `getChildren` function of the tree.

https://user-images.githubusercontent.com/2781777/224338079-5421449d-eb80-4bc7-8231-099481ca4bd2.mp4


nicoecheza (approved):

nicoecheza:
🚀

PR #453, opened 2023-03-10
fix: transform default values

nicoecheza:
Transform component now has default values


packages/@dcl/ecs/src/components/manual/Transform.ts
```diff
@@ -135,10 +135,10 @@ export function defineTransformComponent(
   return {
     ...transformDef,
     create(entity: Entity, val?: TransformTypeWithOptionals) {
-      return transformDef.create(entity, val as TransformType)
+      return transformDef.create(entity, { ...TransformSchema.create(), ...val } as TransformType)
```
gonpombo8, line 138:
```suggestion
      return transformDef.create(entity, { ...TransformSchema.create(), ...val })
```
do we still need the `as` ? 
TransformSchema.create() already returns `TransformType`

gonpombo8: also why dont use the `TransformSchema.extend` that already does the spread operator

nicoecheza: good catch, fixed 👌

leanmendoza (approved):

gonpombo8 (approved):

PR #454, opened 2023-03-10
chore: add bundle js size for snapshots (tests)

leanmendoza:
# What?
Add the JS compiled file-size of snapshots

# Why?
To compare in future SDK versions.


gonpombo8 (approved):
🤍

PR #461, opened 2023-03-14
chore: ui button disabled property

pravusjif:
Added disabled property for Ui Button helper, when the button is **disabled** the onMouse components are not attached to the uiEntity.

Tested working with the following test code in a local scene:
```
ReactEcsRenderer.setUiRenderer(
  () => (
    <UiEntity
    uiTransform={{
      width: '50%',
      height: '35%',
      margin: '10% 0 5% 25%',
      padding: 4,
    }}
    uiBackground={{ color: Color4.create(0.5, 0.8, 0.1, 0.6) }}
    >
        <UiEntity
          uiTransform={{
            width: '100%',
            height: '100%',
            flexDirection: 'column',
            alignItems: 'center',
            justifyContent: 'space-between'
          }}
          uiBackground={{ color: Color4.fromHexString("#70ac76ff") }}
        >
            <Dropdown
                uiTransform={{
                    width: '30%',
                    height: '20%'
                }}
                acceptEmpty={false}
                emptyLabel={'Select your team (BOEDO)'}
                options={['BUTTON ENABLED', 'BUTTON DISABLED']}
                selectedIndex={selectedIndex}
                color={Color4.Red()}
                textAlign="bottom-center"
                font="sans-serif"
                fontSize={20}
            />
            <Button
                uiTransform={{ width: 100, height: 40, margin: 8 }}
                value='Spawn cube'
                variant='primary'
                color={Color4.Black()}
                // uiBackground={{
                //     color: Color4.Blue()
                //     // color: Color4.create(1, 1, 1, 0.2)
                // }}
                disabled={buttonDisabled}
                fontSize={14}
                onMouseDown={() => {
                    console.log("BUTTON CLICKED!")
                }}
            />
        </UiEntity>
    </UiEntity>
  )
)

let selectedIndex = 0
let buttonDisabled = false
let valueChangeTimer = 5
engine.addSystem((deltaTime: number) => {
    valueChangeTimer -= deltaTime
    if(valueChangeTimer <= 0) {
        buttonDisabled = !buttonDisabled
        selectedIndex = buttonDisabled ? 1 : 0
        
        valueChangeTimer = 5
    }
})
```

Every 15 seconds the button `disabled` property is toggled (and the select is changed as well). During its `disabled = true` state, the button doesn't trigger the `onMouseDown()` configured.

Issue: https://github.com/decentraland/sdk/issues/547


packages/@dcl/react-ecs/src/components/Button/index.tsx
```diff
@@ -53,10 +53,11 @@ export function Button(props: UiButtonProps) {
     height: 36,
     ...uiTransform
   })
+
   return (
     <entity
-      onMouseDown={onMouseDown}
-      onMouseUp={onMouseUp}
+      onMouseDown={props.disabled ? undefined : onMouseDown}
```
gonpombo8, line 59: should we added some logic to the primary/secondary buttons so when the Button is disabled we change the background color ?
This way the user will know when the button is disabled or not

pravusjif: Totally, I wasn't sure why there were the primary or secondary variants and now it makes sense XD, I'll update it, so if the new 'disabled' property is being used, based on that I'll force either the 'primary' or 'secondary' variant

pravusjif: Done! found a way of affecting the alpha of the text and background colors regardless of their configuration, check it out

packages/@dcl/react-ecs/src/components/Button/index.tsx
```diff
@@ -53,10 +53,20 @@ export function Button(props: UiButtonProps) {
     height: 36,
     ...uiTransform
   })
+
+  if(!textProps.color) {
+    textProps.color = { r: 1, g: 1, b: 1, a: 1 }
+  }
+  const disabled : boolean = props.disabled ? props.disabled : false
+  if(disabled) {
```
gonpombo8:
```suggestion
  if(!!props.disabled) {
```

packages/@dcl/react-ecs/src/components/Button/index.tsx
```diff
@@ -53,10 +53,20 @@ export function Button(props: UiButtonProps) {
     height: 36,
     ...uiTransform
   })
+
+  if(!textProps.color) {
+    textProps.color = { r: 1, g: 1, b: 1, a: 1 }
+  }
+  const disabled : boolean = props.disabled ? props.disabled : false
```
gonpombo8:
```suggestion
  const disabled : boolean = !!props.disabled
```

gonpombo8 (approved):

PR #471, opened 2023-03-17
feat: add composite init to inspector

leanmendoza:


packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -13,8 +13,11 @@ export function initRpcMethods(fs: Fs, engine: IEngine): DataLayerInterface {
     // This method receives an incoming message iterator
     // and returns an async iterable. consumption and production of messages
     // are decoupled operations
-    stream: function (iter) {
-      return stream(iter, { engine })
+    async *stream(iter, ctx) {
+      const gen = stream(iter, { engine })
+      for await (const it of gen) {
```
gonpombo8, line 18: whats the logic behind this ? 
I think stream fn already does this.

packages/@dcl/inspector/src/lib/data-layer/client/local-data-layer.ts
```diff
@@ -0,0 +1,69 @@
+import { createEngine } from '../host/engine'
+import { DataLayerRpcClient, DataLayerRpcServer, FileSystemInterface } from '../types'
+import { initRpcMethods } from '../host/rpc-methods'
+import { Composite, compositeFromJson, CompositeProvider, instanceComposite } from '@dcl/ecs'
+
+export async function createFsCompositeProvider(fs: FileSystemInterface): Promise<CompositeProvider> {
+  const compositePaths = (await fs.getDirectoryFiles('/')).filter(
```
gonpombo8, line 7: why we have this fn here ? I think both client/server rpc can use the same logic by only wrapping the fileSystem 🤔

leanmendoza: you're right, moving it :)

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
leanmendoza: just something wrong in the types, adding a TODO

gonpombo8 (approved):

PR #473, opened 2023-03-17
fix: prevent re-rendering react every 16ms by using hooks

cazala:
This PR introduces a few hooks, and refactors the code to use them, in order to prevent the current loop that re-render the whole app every 16ms.

The hooks introduced are:

- `useSdk`: It returns an `SdkContext` that contains a reference to the engine, the components, the babylon scene, and an event emitter that notifies changes in the engine. With this hook we can avoid passing around the now deprecated `inspectorEngine` that used to be passed as props from the app level to the other components down the tree.

- `useChange`: It's used to register a callback that will be called every time there's a change in the engine, as long as the component is mounted.

- `useRenderer`: this is only used to collect the reference to the canvas in the `Renderer` component in an easy way, and avoid passing the reference via the now deprecated `onLoad` callback to the top level.

I also introduced a `withSdk` higher order component that can be used to wrap a component and make it render it only after the sdk context has been initialized, and pass the sdk context as a prop (so the wrapped component doesn't have to deal with an uninitialized (ie `null`) context.


leanmendoza (approved):
Awesome work!

PR #479, opened 2023-03-21
feat: integrate file system to data-layer

leanmendoza:
# What? 
Add `sdk-commands` integration of file system interface based on Node.js FileSystem API

### DIFF
--- a/packages/@dcl/inspector/src/lib/data-layer/client/local-data-layer.ts
+++ b/packages/@dcl/inspector/src/lib/data-layer/client/local-data-layer.ts
@@ -1,3 +1,4 @@
+import { OnChangeFunction } from '@dcl/ecs'
 import { createEngine } from '../host/engine'
 import { initRpcMethods } from '../host/rpc-methods'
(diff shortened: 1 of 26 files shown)


packages/@dcl/inspector/src/lib/data-layer/client/local-data-layer.ts
```diff
@@ -1,32 +1,12 @@
...
+import { DataLayerRpcClient, FileSystemInterface } from '../types'
 
 /**
- * This RpcClient creates internally the server, implementing its own file system interface and engine.
+ * This RpcClient creates internally the DataLayer HOST, implementing its own file system interface and engine.
  * @param fs
  * @returns
  */
 export async function createLocalDataLayerRpcClient(fs: FileSystemInterface): Promise<DataLayerRpcClient> {
-  const engine = createEngine()
```
leanmendoza, line 15: This function only creates the `RpcClient`, the logic was moved to `host/index.ts` (and now it describes the DataLayer, and reused in the `sdk-commands`)

gonpombo8 (approved):

PR #481, opened 2023-03-21
fix: set renderer within an effect to avoid updating the state during a render

cazala:
Fixes https://github.com/decentraland/sdk/issues/664


leanmendoza (approved):

leanmendoza:
Integrated in https://github.com/decentraland/js-sdk-toolchain/pull/479

PR #485, opened 2023-03-23
chore: put composite in a namespace and rename proto-message to avoid collissions

leanmendoza:
# What?
Move the types and function in top of library to a namespace `Composite`

Before
```ts
compositeFromJson(...)
instanceComposite(...)
const provider: CompositeProvider = createProvider()
```

Now
```ts
Composite.fromJson(...)
Composite.instance(...)
const provider: Composite.Provider = createProvider()
```


packages/@dcl/ecs/src/composite/instance.ts
```diff
@@ -6,16 +6,23 @@ import { ComponentDefinition, IEngine, LastWriteWinElementSetComponentDefinition
 import { Schemas } from '../schemas'
 import { ReadWriteByteBuffer } from '../serialization/ByteBuffer'
 import { getCompositeRootComponent } from './components'
-import { ComponentData, Composite, CompositeComponent, CompositeProvider } from './types'
+import { ComponentData, CompositeComponent, CompositeDefinition } from './proto/gen/composite.gen'
```
gonpombo8, line 9: why not use the types defined in the Composite namespace ? They are called the same

gonpombo8 (approved):

packages/@dcl/ecs/src/composite/instance.ts
leanmendoza: because there would be a cyclic dependency.
both `instance.ts` and `index.ts` (which holds the namespace) use the `proto/gen/composite.gen`

PR #488, opened 2023-03-25
Env config component instead of dclinfo

menduz:
Loads the following .env files

1. `@dcl/sdk-commands/.dclrc`
1. `~/.dclrc`
1. `$(pwd)/.dclrc`
1. `process.env`

in that order. Every .env composes and overrides the keys from the previous one. This will enable everybody to have granular control of their configuration files per-project.

It also simplifies a lot the code, using a very battle tested solution


gonpombo8 (approved):

gonpombo8 (approved):
LGTM, thanks menduz.
The only thing that I see is that we change the file name so we are going to have a spike in the metrics because there are going to be new users/events.
JFYI @nearnshaw

PR #491, opened 2023-03-28
fix: entity selected reparented

cazala:
Fixes https://github.com/decentraland/sdk/issues/671

The issue was not related to the removal of the EntitySelected component, the issue was that at that same time, the EntityInspector would sync the inputs with the engine with the same value that the engine had, but it removed the `parent` property every time because it was not included when creating the new tranform from the values in the inputs.

### DIFF
--- a/packages/@dcl/inspector/src/components/Tree/Tree.tsx
+++ b/packages/@dcl/inspector/src/components/Tree/Tree.tsx
@@ -67,7 +67,7 @@ function Tree<T>(props: Props<T>) {
   const quitInsertMode = () => setInsertMode(false)
 
   const handleToggleExpand = (_: React.MouseEvent) => {
-    onToggle(value, !open)
+    onToggle(value, !selected || !open)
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -67,7 +67,7 @@ function Tree<T>(props: Props<T>) {
   const quitInsertMode = () => setInsertMode(false)
 
   const handleToggleExpand = (_: React.MouseEvent) => {
-    onToggle(value, !open)
+    onToggle(value, !selected || !open)
```
cazala, line 70: This is not related to the issue that this PR fixes, but I took the chance to solve it because it was very annoying: every time you selected a different entity in the tree, it would close it (toggle it). Now it only toggles when you click the currently selected entity.

leanmendoza (approved):

PR #494, opened 2023-03-30
add styles to inspector tree and context menu

nicoecheza:
gd is not my passion but did what I could...

Screen Shot 2023-03-30 at 17 57 13

TDB => when do we want to remove focus from current selected entity?
* right now, the current behaviour is to always leave the last clicked entity as selected until you click in a different entity. Maybe this is desired behaviour, maybe not. We need to define it


packages/@dcl/inspector/src/components/Controls/Controls.css
```diff
@@ -0,0 +1,39 @@
...
+.component.controls.active {
+  background-color: #424141;
+}
+
+.component.controls > div:first-child {
+  display: flex;
+  align-items: center;
+}
+
+.component.controls > div:nth-child(2) {
```
menduz, line 16: Wouldn't it be better to use .class to prevent this style from being broken if the markup changes?

nicoecheza: I come from the approach of "changing the markup should affect styles" (an old ux developer colleague directive). I honestly don't have a preference on this one (besides classes cluttering a little bit more our JSX). 
Is using classes the way to go on DCL? 
@cazala what do you think?

packages/@dcl/inspector/src/components/Tree/Tree.css
```diff
@@ -1,39 +1,70 @@
-.tree {
-  background-color: #ffffff;
-  overflow: scroll;
-  height: 100vh;
-  width: 20rem;
-  border-right: 0.1rem solid #dedede;
+.Hierarchy {
```
cazala, line 1: One component's style should not be dependant on other's. The `Tree` component should work even when not used within `Hierarchy`. Let's change this to be `.Tree` instead (and lets add the `Tree` class to the components outter div).

packages/@dcl/inspector/src/components/Controls/Controls.css
cazala: mm i'm okay on using the tag name as selector when it is semantic (using label, input, button). For instance I would do this:

```tsx
<div className="MyComponent">
  <label>{label}</label>
</div>
```

```css
.MyComponent label { ... }
```

Instead of:

```tsx
<div className="MyComponent">
  <div className="label">{label}</div>
</div>
```

```css
.MyComponent .label { ... }
```

But using `div`s is not very semantic and i think it makes the code more difficult to follow than instead seeing `.menu-item` or something more meaningful, i would suggest as mendez said using classes here.

packages/@dcl/inspector/src/components/Controls/Controls.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (cancelingKeys.has(e.key)) onCancel(e)
+  }
+
+  useEffect(() => {
+    if (active) {
+      document.body.addEventListener('click', onCancel)
+      document.body.addEventListener('keyup', onKeyUp)
+    } else {
+      document.body.removeEventListener('click', onCancel)
+      document.body.removeEventListener('keyup', onKeyUp)
```
cazala, line 46: if the component is unmounted while it was active, these lines will not execute. We should also remove the listeners when the component unmounts to avoid calling these handlers which would fail when the component is not mounted:

```ts
useEffect(() => {
  return () => {
    document.body.removeEventListener('click', onCancel)
    document.body.removeEventListener('keyup', onKeyUp)
  }
}, [onCancel, onKeyUp])
```

And also lets wrap all the hanlders (onCancel,  onKeyUp, handleClick, handleAction) with `useCallback` to prevent extra renders and effects.

packages/@dcl/inspector/src/components/Controls/Controls.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const handleAction = (cb: (e: React.MouseEvent) => void) => (e: React.MouseEvent) => {
+    setActive(false)
+    cb(e)
+  }
+
+  const someActionIsEnabled = enableAdd || enableEdit || enableRemove
+  const shouldRenderActions = someActionIsEnabled && active
+
+  return (
+    <div className={`component controls ${getActiveClass(active)}`}>
```
cazala, line 64: Following the standards the outter class of the component should match the component name `Controls`.

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -115,41 +118,50 @@ function Tree<T>(props: Props<T>) {
...
+      {insertMode && <Input value="" onCancel={quitInsertMode} onSubmit={handleAddChild} />}
+    </div>
+  )
+}
+
+function Arrow({ isOpen }: { isOpen: boolean }) {
+  return <>{isOpen ? <ArrowDown /> : <ArrowRight />}</>
+}
+
+function TreeChildren<T>(props: Props<T>) {
```
cazala, line 152: Why moving this part into a new component for this instead of doing it within `Tree`?

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -115,41 +118,50 @@ function Tree<T>(props: Props<T>) {
...
+        {editMode && <Input value={label || ''} onCancel={quitEditMode} onSubmit={onChangeEditValue} />}
+        <Controls {...controlsProps} />
+      </div>
+      <TreeChildren {...props} />
+      {insertMode && <Input value="" onCancel={quitInsertMode} onSubmit={handleAddChild} />}
+    </div>
+  )
+}
+
+function Arrow({ isOpen }: { isOpen: boolean }) {
```
cazala, line 148: We tend to keep one component per file. Making this a separate component seems a bit of an overkill to me tho, is it going to be used somewhere else? if not, then we could just inline this in the Tree component.

cazala (changes requested):
Looking good 👍 I left a few comments mostly around standards and a few questions

packages/@dcl/inspector/src/components/Tree/Tree.tsx
nicoecheza: the `Tree` component already has a lot of logic in it. It reads better by removing all the `{...logic}` stuff from it and just leave markup with props in the main component

cazala:
It looks like it's possible to open several menus without the previous one closing (by right-clicking on each row without ever closing the menu). Maybe we could keep a context in the outtermost Tree to know on which entity the menu is open and only show the menu that particulat entity.

Screenshot 2023-04-03 at 15 39 01

PR #496, opened 2023-03-31
chore: bundle media files, split reusable components, hot reload 🔥

cazala:
In this PR I tackled a few issues:

- Configured the bundler to load and inline media files. They end up as part of our current bundles (css, js) so they don't require doing requests to load them independently, which was a problem when loading the inspector without using the iframe.
- Added `decentraland-ui` and `@vscode/webview-ui-toolkit`
- I tidied up a little bit some files (there were inconsistencies on default exports vs named exports, we do named exports on `index.ts` files and default exports on `[Component].ts` files, this is so VSCode knows which one to use when auto importing a component).
- I spilt the Transform into some reusable components: Container, Block, TextField. So we can use them when building the next component inspectors.
- I added hot reloading for css files and live reloading for js files in the dev server 🔥


nicoecheza (approved):
works like 🔥

PR #503, opened 2023-04-04
chore: update protocol version

pravusjif:


packages/@dcl/ecs/package.json
```diff
@@ -5,7 +5,8 @@
   "author": "DCL",
   "bugs": "https://github.com/decentraland/ecs/issues",
   "dependencies": {
-    "@dcl/js-runtime": "file:../js-runtime"
+    "@dcl/js-runtime": "file:../js-runtime",
+    "@dcl/protocol": "^1.0.0-4597420155.commit-e877adc"
```
menduz, line 9: this change doesn't actually do what you are trying. please update only the (root)/package.json

menduz (changes requested):
the protocol package was migrated

PR #508, opened 2023-04-06
feat: add/remove components on tree from context menu

nicoecheza:
closes https://github.com/decentraland/sdk/issues/641


packages/@dcl/inspector/src/hooks/sdk/useTree.ts
```diff
@@ -135,20 +152,35 @@ export const useTree = () => {
...
     tree,
+    addComponent,
     addChild,
     setParent,
     rename,
     remove,
     toggle,
     getId,
     getChildren,
+    getEntityComponents,
```
leanmendoza, line 183: Why these methods in `useTree`?

leanmendoza (approved):
LGTM! I just left a comment 😄

packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
```diff
@@ -1,19 +1,32 @@
...
 import { ROOT } from '../../lib/sdk/tree'
+import { useSdk } from '../../hooks/sdk/useSdk'
+import { isLastWriteWinComponent } from '../../hooks/sdk/useComponentValue'
 
 export const EntityInspector: React.FC = () => {
   const selectedEntities = useEntitiesWith((components) => components.EntitySelected)
+  const sdk = useSdk()
   const entity = useMemo(() => (selectedEntities.length > 0 ? selectedEntities[0] : null), [selectedEntities])
 
+  const handleRemove = (entity: Entity, componentId: number) => {
```
cazala, line 16: why is this piece of logic here instead of within the `TransformInspector`? To me it seems an unnecesary coupling of the parent and the children, constructing this here and passing it as a prop, instead of simply doing a `Transform.deleteFrom(entity)` inside the component.

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -31,147 +24,143 @@ type Props<T> = {
   canToggle?: (value: T) => boolean
   onSetParent: (value: T, parent: T) => void
   onRename: (value: T, label: string) => void
+  onAddComponent: (value: T, componentId: number) => void
```
cazala, line 27: mm I'm worried that this component, that is supposed to be an abstract tree, it's growing too coupled with the composite inspector. For instance we could use this tree component to render the files in the Asset Explorer, but then we would need that tree to provide these methods that only make sense when `T` is an entity. I think I would rather add a `getContextMenu: (value: T) => MenuItem[]` function where each tree can provide custom actions for the context menu, and keep this component pure/abstract.

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -36,142 +28,130 @@ type Props<T> = {
...
-    (target: T, source: T): boolean => {
-      if (getId(target) === getId(source)) return false
-      return getChildren(target).every(($) => canDrop($, source))
-    },
-    [getId, getChildren]
-  )
+function Tree<T>(_props: Props<T>) {
+  const Component = withContextMenu<Props<T>>((props) => {
+    const {
+      extraContextMenu,
```
cazala, line 39: what do you think about renamig this to `getExtraContextMenu` in order to keep it consistent with the other `(value: T) => ...` getters in the props?

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -1,24 +1,16 @@
...
-}
-
-interface ContextMenuProps<T> {
-  contextMenu?: ContextMenu<T>
-  onContextMenuChange: (value?: ContextMenu<T>) => void
-}
-
 type Props<T> = {
   value: T
+  extraContextMenu: (value: T) => JSX.Element | null
```
cazala, line 13: would you make this optional so it's not mandatory for all the trees to have an extra context menu getter?

cazala (approved):
looking good! I left a few suggestions but non blockers, approved ✅

packages/@dcl/inspector/src/components/Tree/Tree.tsx
nicoecheza: sure!

packages/@dcl/inspector/src/components/Tree/Tree.tsx
nicoecheza: fixed 👍

cazala (approved):

PR #509, opened 2023-04-07
feat: add reading data-layer ws from globalthis

leanmendoza:
Editor part: https://github.com/decentraland/editor-sdk7/pull/20
Closes https://github.com/decentraland/sdk/issues/696

### DIFF
--- a/packages/@dcl/inspector/public/index.html
+++ b/packages/@dcl/inspector/public/index.html
@@ -13,7 +13,12 @@
   <div id="root"></div>
   <link href="bundle.css" rel="stylesheet" />
   <script>
+    // The next line should be replace with the next regex 
+    //  indexHtmlContent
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/public/index.html
```diff
@@ -12,6 +12,12 @@
   <noscript>You need to enable JavaScript to run this app.</noscript>
   <div id="root"></div>
   <link href="bundle.css" rel="stylesheet" />
+  <script>
+    const config = '$CONFIG'
+    if (config !== '$CONFIG') {
```
cazala, line 17: Would you like to add a comment that his is meant to be replaced by the Editor? Otherwise just by looking at it seems like the line below the if would never be reached :)

cazala (approved):

packages/@dcl/inspector/public/index.html
leanmendoza: added 👍 thanks

PR #510, opened 2023-04-07
update UI for project assets explorer

gonpombo8:
image

### DIFF
--- a/packages/@dcl/inspector/src/components/App/App.css
+++ b/packages/@dcl/inspector/src/components/App/App.css
@@ -24,8 +24,9 @@ code {
 }
 
 .sidebar {
-  max-width: 300px;
-  min-width: 300px;
+  height: 100%;
+  max-width: var(--sidebar);
+  min-width: var(--sidebar);
   background: var(--main-bg-color);
 }
 
@@ -42,15 +43,12 @@ code {
   position: absolute !important;
 }
 
+.main-editor {
+  height: calc(100vh - 54px);
+}
+
+
 /* Babylon frame */
-#main-editor {
-  height: 100vh;
-}
-#main-editor > div:first-child {
-  height: 100%;
-  display: flex;
-  flex-direction: column;
-}
 #renderer {
   width: 100%;
   height: 100%;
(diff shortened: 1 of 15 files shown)


packages/@dcl/inspector/src/components/ProjectAssetExplorer/ProjectView.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+    setOpen(new Set(open))
+  }, [open, setOpen])
+
+  if (!folders.length) return null
+
+  const expandedClass = open.has(ROOT) && 'expanded'
+
+  return (
+    <div className={`testing ${expandedClass}`}>
```
cazala, line 80: `testing` classname

packages/@dcl/inspector/src/components/ProjectAssetExplorer/ProjectAssetExplorer.css
```diff
@@ -1,53 +1,11 @@
...
-.asset-explorer-item i {
-  margin-right: 10px;
-}
-
-.asset-explorer-right {
-  flex: 1;
-  background-color: #088270;
-  padding: 20px;
-  overflow: scroll;
+.editor .Tree span {
```
cazala, line 9: These styles are dependant on parent classes, that could lead to errors if the on this component if the parent component changes. We usually let all the rules in a component css to be scoped to the component's outter class, which should match the name of the component, ie: `.ProjectAssetExplorer`

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -1,11 +1,12 @@
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useState } from 'react'
 import { useDrag, useDrop } from 'react-dnd'
 
 import { Input } from '../Input'
 import { Controls, Position } from '../Controls'
 import { RxDoubleArrowRight as ArrowRight, RxDoubleArrowDown as ArrowDown } from 'react-icons/rx'
 
 import './Tree.css'
+import { className } from '@babylonjs/core'
```
cazala, line 9: this is probably incorrectly imported right?

cazala:
This looks really good :D but as @menduz pointed above, it seems like the ability to use the assets from the asset packs is gone. They used to be able to be dropped into the Asset Explorer, but now that does not seem to be possible anymore. Maybe we should let the assets from the asset packs section be droppable stright into the renderer? And they would be downloaded to the file system if necessary at that point?

packages/@dcl/inspector/src/components/App/App.css
```diff
@@ -24,27 +24,68 @@ code {
...
   outline: none;
 }
+
+/* Editor assets */
+.footer {
+  position: absolute;
+  bottom: 0px;
+  display: flex;
+  flex-direction: column;
+  width: calc(100% - var(--sidebar) - 16px) !important;
```
gonpombo8, line 65: move 16px to padding var

packages/@dcl/inspector/src/components/App/App.tsx
```diff
@@ -1,35 +1,65 @@
...
-        <div>
+        <Box className="main-editor">
           <Renderer />
-        </div>
-        <div style={{ flex: '1', display: 'flex', overflow: 'auto' }}>
-          <div style={{ width: '50%', overflow: 'auto' }}>
-            <ProjectAssetExplorer />
+        </Box>
+        <Box className="footer">
+          {tab && (
```
gonpombo8, line 46: move this to a new component file

PR #511, opened 2023-04-07
feat: added Scene component

cazala:
https://user-images.githubusercontent.com/2781777/230673114-9b78987a-fa3f-464d-8a06-0f57212b01b7.mp4

---

Added `Scene` component. This contains a `layout` property with a `base` and `parcels`. There's a `SceneInspector` component that for now I left always visible, although I think we should find a better place for this global settings rather than the Entity Inspector.

I also modified the renderer to show the scene boundaries, and the gizmo manager to prevent the user from moving stuff outside of the margins.


menduz:
amazing!

I;d be careful with preventing arbitrary entities from being outside the boundaries of the scenes.. often the pivot point of models makes the 0,0,0  arbitrary and due to rotations we may not be able to place things correctly. I'd say we defer that one, the vertices cannot be of the models outside the scene, the entity _can_

gonpombo8 (approved):

PR #514, opened 2023-04-10
feat: gizmos

cazala:
### <samp>🤖 Generated by Copilot at 4c5962b</samp>

This pull request adds a Toolbar component to the inspector module, which allows users to switch between different gizmo modes for manipulating the selected entity. It also updates the inspector module to use the new Selection component and function, which replace the EntitySelected component and the putEntitySelectedComponent function, for marking and changing the selected entity. Additionally, it fixes some bugs, improves some types and styles, and refactors some code for readability and consistency. 

This PR sets the translation gizmo as default and fixes it to the world rotation insteaf of the local one. It also adds buttons to change the selected gizmo.

Closes https://github.com/decentraland/sdk/issues/689
Fixes https://github.com/decentraland/sdk/issues/683


leanmendoza (approved):
LGTM! 
Is there any enum for the gizmos? It seems like babylon uses the hard-coded values also

cazala:
mm @menduz would you be more specific? The green orbit is the y-axis. The gizmo does not rotate along the model, but the axis are the local ones of the model (that' sthe way the gizmo works). So rotating the green orbit should rotate the model around it's local y-axis (in the case of the model, it would make it spin around its pivot's vertical axis):


https://user-images.githubusercontent.com/2781777/231243970-c0d997c7-7504-4824-896d-75a23c32ad20.mp4

menduz:
@cazala I like that even more than the global reference

PR #521, opened 2023-04-13
fix: inputs

cazala:
Fixes https://github.com/decentraland/sdk/issues/703

The issue was formatting the input value within the render of the react component. The conversion from component value into input value should be done within the `fromComponentValueToInput` function provided to the `useComponentInput` hook:

```ts
fromComponentValueToInput(componentValue: ComponentValueType) => InputType
``` 

Otherwise the hook fails to compare the values and fires unnecessary syncs.

I also fixed an issue with the `Scene > Parcels` input which was bound to the wrong value

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/TransformInspector/utils.ts
+++ b/packages/@dcl/inspector/src/components/EntityInspector/TransformInspector/utils.ts
@@ -2,7 +2,7 @@ import { TransformType } from '@dcl/ecs'
 import { Quaternion } from '@dcl/ecs-math'
 import { TransformInput } from './types'
 
-export function fromTranform(value: TransformType): TransformInput {
+export function fromTransform(value: TransformType): TransformInput {
   const angles = Quaternion.toEulerAngles(value.rotation)
   return {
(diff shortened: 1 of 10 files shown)


packages/@dcl/inspector/src/components/EntityInspector/TransformInspector/utils.ts
```diff
@@ -1,38 +1,38 @@
 import { TransformType } from '@dcl/ecs'
-import { Quaternion } from '@dcl/ecs-math'
+import { Quaternion } from '@babylonjs/core'
```
cazala, line 2: I haad to change the implementation to use babylon's Quaternion instead of the `@dcl/ecs-math` because I could not make jest work with that package :(

cazala, line 2: I haad to change the implementation to use babylon's Quaternion instead of the `@dcl/ecs-math` because I could not make jest work with that package :(

gonpombo8 (approved):

PR #525, opened 2023-04-17
feat: add Gltf UI component

nicoecheza:


packages/@dcl/inspector/src/components/Block/Block.tsx
```diff
@@ -1,12 +1,40 @@
...
+  useEffect(() => {
+    onDropHover(isActive)
+    return () => onDropHover(false)
+  }, [isActive])
+
   return (
-    <div className="Block">
-      {props.label && <label>{props.label}</label>}
-      <div className="content">{props.children}</div>
+    <div ref={drop} className="Block">
```
cazala, line 35: I would suggest, instead of adding the drop functionality into the `Block` component, which ends up with logic split between this component and the parent, and props being passed around, to just foward the reference in the `Block` component and let the parent use the ref for any use case that would require a `ref`. 

For example: 

```tsx
const Block = React.fowardRef((props, ref) => {
  // ...
  return (
    <div className="Block" ref={ref}>
      {/* ... */}
    </div>
  )
})
```

The the parent of `Block` can do do the `useDrop` or any other hook and pass the ref to like `<Block ref={drop}>`

packages/@dcl/inspector/src/components/ProjectAssetExplorer/ProjectAssetExplorer.tsx
```diff
@@ -1,35 +1,16 @@
...
-    setSelectedFolder(tree)
-  }
-
-  useSdk(async ({ dataLayer }) => {
-    await updateAssetCatalog(dataLayer)
-  })
+  const { files } = useFileSystem()
 
-  const folders = selectedFolder.children.filter((item) => item.type === 'folder') as AssetNodeFolder[]
+  const tree = buildAssetTree(files.assets.map((item) => item.path))
```
cazala, line 12: could this be wrapped in a `useMemo(() => buildAssetTree(...), [files])` so it does not get re-build on every render? Maybe it can be done within a `useAssetTree` hook so it's always done instead of using `buildAssetTree` directly.

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -28,6 +28,8 @@ type Props<T> = {
   onAddChild: (value: T, label: string) => void
   onRemove: (value: T) => void
   onToggle: (value: T, isOpen: boolean) => void
+  getDragValue?: (value: T) => unknown
```
cazala, line 31: mm I think this here is dangerous. It changes the type of `value` to be `unknown`, but the drop handler types it as `T` and will handle it as such. That could cause some runtime errors if the actual type of `value` is not compatible with the new `unknown` type, here for example: https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/inspector/src/components/Tree/Tree.tsx#L86

cazala (changes requested):

packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -78,11 +82,11 @@ function Tree<T>(_props: Props<T>) {
       [getId, getChildren]
     )
 
-    const [, drag] = useDrag(() => ({ type: 'tree', item: { value } }), [value])
+    const [, drag] = useDrag(() => ({ type: dndType, item: { value, transform: getDragValue(value) } }), [value])
```
cazala, line 85: mm okay this now does not break the drop handler, but IMO this seems too specific for this use case. I understand that with the current implementation just having `value` is not enough and that you want to be able to get the full `TreeNode`.

I would suggest the following to address the problem of passing `tree` from the context of the `ProjectView` into the one of the `GltfInspector`:

1) In `Tree.tsx`:

Replace line 31
```ts
getDragValue?: (value: T) => any
```` 
with 
```ts
getDragContext?: () => any
```

And then instead of this on line 85:

```ts
transform: getDragValue(value)
```
do this:
```ts
context: getDragContext && getDragContext()
```

Now we have a generic way off passing data between the source (ProjectView) and the target (GltfInspector) that is not constrained to only transforming the `value`.

2) In `ProjectView.tsx`:

Replace 
```ts
getDragValue={(val) => tree.get(val)!}
```
with
```ts
getDragContext={() => ({ tree })}
```

We pass the whole tree within the context

3) Finally on `GltfInspector.tsx`:

Replace this:

```ts
 drop: ({ transform }: { transform: AssetNodeItem }, monitor) => {
  if (monitor.didDrop()) return
  handleDrop(transform)
}
```
With this
```ts
drop: ({ value, context }: { value: string; context: { tree: Map<string, AssetNodeItem> } }, monitor) => {
  if (monitor.didDrop()) return
  handleDrop(context.tree.get(value)!)
},
```

It's a little bit more code, but I believe it's more generic and could be useful for any use case that requires passing stuff from source to target that is different to `T`

cazala (changes requested):

packages/@dcl/inspector/src/components/Tree/Tree.tsx
nicoecheza: sound good. Instead of providing a way to change the value before dragging, we provide context to change it after dropping 👍

cazala (approved):
Looking and working great! 🚀 ✅

PR #528, opened 2023-04-18
fix deploy command

gonpombo8:
- ServerComponent was being intialized with the wrong config component
- config component was expecting to have LANDRegistry string always declared.
So when you run sdk-commands deploy if failes because there was no config.LANDRegistry value, and since the default values are already hardcoded on the linker-dapp, we can ommit the requiredString to a readString

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
+++ b/packages/@dcl/sdk-commands/src/commands/deploy/index.ts
@@ -1,21 +1,18 @@
 import { resolve } from 'path'
-import { EntityType, ChainId, Scene } from '@dcl/schemas'
-import { CatalystClient, ContentClient, DeploymentBuilder } from 'dcl-catalyst-client'
+import { EntityType, ChainId } from '@dcl/schemas'
 import { Authenticator } from '@dcl/crypto'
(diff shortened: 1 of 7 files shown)


menduz (commented):
Alternatively to have a single source of truth, I may add the contract addresses to `@dcl/sdk/.dclrc` and make them required configs again

gonpombo8:
> Alternatively to have a single source of truth, I may add the contract addresses to `@dcl/sdk/.dclrc` and make them required configs again


They deppend on the network, so they cant be hardcoded on the .dclrc. 
Maybe we can add the some logic to the file and return each key depending on the network but we already have this logic on the linker-dapp.

pravusjif:
I could deploy the scene with this PR's package, although now the scene never finishes loading in production for some reason... (-129, -77; to see it loading forever you have to enter the world nearby and walk up to the scene)

pravusjif:
> now the scene never finishes loading in production for some reason... 

The problem is not from this PR, it's the `@dcl-sdk/utils` package that breaks scenes in production.

pravusjif (approved):

PR #532, opened 2023-04-19
fix: grid

cazala:
Fixes https://github.com/decentraland/sdk/issues/721

### DIFF
--- a/packages/@dcl/inspector/__mocks__/@babylonjs/inspector.ts
+++ b/packages/@dcl/inspector/__mocks__/@babylonjs/inspector.ts
@@ -1,0 +1,1 @@
+export default {}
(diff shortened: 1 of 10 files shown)


menduz:
It is possible to disable the zDepth of the ground grid and use both grids on the same height to prevent the oddity of the screenshot

cazala:
@menduz I tried that, but I still get this happening where the grid is not rendered on top of the ground in certain angles:

Screenshot 2023-04-19 at 17 21 15

while using the 0.01 gap it does not happen:

Screenshot 2023-04-19 at 17 20 39

Something I could do is place both the ground grid and the layout grid at 0.01 so they match (you would not see the double lines if you zoom in). Though if you zoom in enough on a cube placed on Y=0 you would see it a bit under the grid.

menduz:
I think it's perfect now thanks

Screenshot 2023-04-19 at 18 04 32

gonpombo8 (approved):

PR #537, opened 2023-04-24
feat: video events

pravusjif:
* Updated protocol to use new video event component
* Created video events helper system to read video events component change and trigger helper callback

Usage example:
```
import {
  engine,
  Entity,
  VideoPlayer,
  videoEventsSystem,
  VideoState
} from '@dcl/sdk/ecs'

// ... Create videoPlayerEntity with VideoPlayer component, Transform, MeshRenderer.setPlane(), etc. ...

videoEventsSystem.registerVideoEventsEntity(videoPlayerEntity, (videoEvent) => {
  console.log('video event - state: ' + videoEvent.state
    + '\ncurrent offset:' + videoEvent.currentOffset
    + '\nvideo length:' + videoEvent.videoLength)

  switch (videoEvent.state) {
    case VideoState.VS_READY:
      console.log('video event - video is READY')
      break
    case VideoState.VS_NONE:
      console.log('video event - video is in NO STATE')
      break
    case VideoState.VS_ERROR:
      console.log('video event - video ERROR')
      break
    case VideoState.VS_SEEKING:
      console.log('video event - video is SEEKING')
      break
    case VideoState.VS_LOADING:
      console.log('video event - video is LOADING')
      break
    case VideoState.VS_BUFFERING:
      console.log('video event - video is BUFFERING')
      break
    case VideoState.VS_PLAYING:
      console.log('video event - video started PLAYING')
      break
    case VideoState.VS_PAUSED:
      console.log('video event - video is PAUSED')
      break
  }
})
```

other related PRs:
* https://github.com/decentraland/protocol/pull/137
* https://github.com/decentraland/unity-renderer/pull/5042


packages/@dcl/playground-assets/etc/playground-assets.api.md
```diff
@@ -503,6 +503,7 @@ export const componentDefinitionByName: {
     "core::UiInputResult": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBUiInputResult>>;
     "core::UiText": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBUiText>>;
     "core::UiTransform": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBUiTransform>>;
+    "core::VideoEvent": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBVideoEvent>>;
```
menduz, line 506: Pravs, I think we may create this component as grow only set. Did you think about that?

menduz: The use case is to also control the play/pause from the evets themselves

pravusjif: Uhmm, you are right I'll change it to a GrowOnlySet since the event is only to read data after I finish testing some stuff using the LWW, thx!

pravusjif:
Current implementation works (with unity-renderer counterpart) as the other helper systems in existence:

```
videoEventsSystem.registerVideoEventsEntity(videoPlayerEntity, (videoEvent) => {

})
```

Next it will be changed to a GrowOnlyValueSet and updated through APPEND

packages/@dcl/ecs/src/systems/videoEvents.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface VideoEventsSystem {
+  removeVideoEventsEntity(entity: Entity): void
+
+  registerVideoEventsEntity(entity: Entity, callback: VideoEventsSystemCallback): void
+}
+
+/**
+ * @internal
+ */
+export function createVideoEventsSystem(engine: IEngine): VideoEventsSystem {
```
menduz, line 23: I think we should prevent default systems as much as possible. This entire system can be replaced by a function to obtain the final state of each video like this:


```suggestion
// this function returns the latest state of the video filtering commands from video events.
// the result also provides some functions to mutate the state of the entity like .play(), .pause() and .seek()
declare function getLatestVideoState(entity: Entity): InternalVideoStateNotComponent
```

We don't need to run this every frame. We may not need events at all if people adopt ECS style of programming like:

```
function mySystem() {
  for (const [entity] of engine.getEntitiesWith(Video)) {
    const video = getLatestVideoState(entity)
    Visible.getMustable(spinnerEntity, { visible: video.currentState != PLAYING })
  }
}
```

pravusjif: Uhmmm but the point of these helper systems (also [pointer-events helper system](https://github.com/decentraland/js-sdk-toolchain/blob/8b4e6e0dafaebbfef593428a8e6478512a31c815/packages/%40dcl/ecs/src/systems/events.ts), [raycast helper system](https://github.com/decentraland/js-sdk-toolchain/blob/8b4e6e0dafaebbfef593428a8e6478512a31c815/packages/%40dcl/ecs/src/systems/raycast.ts)) is  that creators can easily just "subscribe" to these specific "events" and not have to implement their own system checking timestamps and stuff.

That's why if we had [the "OnChange" event](https://github.com/decentraland/sdk/issues/740) for components, even these helper systems could be greatly simplifed and let creators have a super friendly way of subscribing to the change of a component (like a pointer event result component, raycast result component or video event component)

menduz: but those are technical debt also.. @gonpombo8 can go in detail about that.. we used the events because that was the legacy way of doing it to enable the events for react UI, not because we believe it is a good way to expose the state of the components to the scene cretaors

menduz: the raycast helpers can also be a single function. it inherited the flaw in design intended for React

gonpombo8: Yes !. We can handle those three events in one single PR just to be consistent.

We can merge this now so we can iterate it after and unblock pravus.

gonpombo8 (approved):

packages/@dcl/ecs/src/systems/videoEvents.ts
pravusjif: cool, now I understand that other approach, I created this issue to tackle the new helper functions as an alternative usage of these "result components": https://github.com/decentraland/sdk/issues/760, feel free to modify it if I miswrote something there

menduz (approved):

PR #543, opened 2023-04-26
feat: snap

cazala:
Closes https://github.com/decentraland/sdk/issues/706

- [x] Add `snapManager` to control snap distance for position/scale/rotation and enable/disable
- [x] Add keybinding: hold `shift` to toggle snapping
- [x] Add UI to edit snap distances and enable/disable
- [x] Add tests

https://user-images.githubusercontent.com/2781777/235001155-52e2b31f-3fc4-480a-ba3b-b17bf4f05654.mp4

*Hint: hold `shift` to toggle the snapping*


packages/@dcl/inspector/jest.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
   collectCoverageFrom: ["src/**/*.ts"],
   verbose: true,
   testMatch: ["**/*.spec.(ts)"],
-  testEnvironment: "node",
+  testEnvironment: "jsdom",
```
gonpombo8, line 28: thankkkkkkkkkks ❤️

packages/@dcl/inspector/src/components/Toolbar/Gizmos/icons/rotate.svg
gonpombo8, line 1: should this svg be react components 🤔 ?
Just to be consistent with the app, no idea just thinking out loud

packages/@dcl/inspector/src/hooks/useOutsideClick.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useRef } from 'react'
+
+export const useOutsideClick = (callback: () => void) => {
```
gonpombo8, line 3: yes!

gonpombo8 (approved):

packages/@dcl/inspector/src/components/Toolbar/Gizmos/icons/rotate.svg
cazala: Sure, we can do that but it would require to configure a loader for svgs so que can `import Icon from "./path/to/icon.svg"` and then use it as `<Icon />`. I would rather do that as a separate PR.

PR #544, opened 2023-04-26
update: editor sidebar to allow vertical resizement

nicoecheza:
closes https://github.com/decentraland/sdk/issues/722

### <samp>🤖 Generated by Copilot at f3b0bac</samp>

This pull request enhances the inspector tool by adding a resizeable hierarchy panel, a memoized tab switching function, and vertical scrolling for the inspector panel. It also fixes a CSS syntax error, disables text selection, and refactors some code for readability and performance. The changes affect the `App`, `Resizable`, `Container`, `Hierarchy`, and `Tree` components and their corresponding CSS and types files.

### DIFF
--- a/packages/@dcl/inspector/src/components/Resizable/types.ts
+++ b/packages/@dcl/inspector/src/components/Resizable/types.ts
@@ -1,5 +1,34 @@
 export interface PropTypes {
-  minWidth?: 'initial' | number
-  initialWidth?: number
+  type: 'horizontal' | 'vertical'
+  min?: 'initial' | number
+  initial?: number
+  max?: number
   onChange?(value: [number, number]): void
+}
+
+export interface TypeProps {
+  offsetValue: 'offsetWidth' | 'offsetHeight'
+  eventClientValue: 'clientX' | 'clientY'
+  css: {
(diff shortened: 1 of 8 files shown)


packages/@dcl/inspector/src/components/Resizable/types.ts
```diff
@@ -1,5 +1,34 @@
...
+export interface TypeProps {
+  offsetValue: 'offsetWidth' | 'offsetHeight'
+  eventClientValue: 'clientX' | 'clientY'
+  css: {
+    childs: 'width' | 'height'
+    handle: 'left' | 'top'
+  }
+}
+
+export const HORIZONTAL_PROPS: TypeProps = {
```
gonpombo8, line 18: these are not types 👀

gonpombo8 (approved):

packages/@dcl/inspector/src/components/Resizable/types.ts
nicoecheza: yes, but they're not important enough to live in their own file and I don't want them in the actual component either, so until this component grows bigger, I think this file is the best for now

PR #545, opened 2023-04-27
import glb & gltf self-contained assets

gonpombo8:
Closes https://github.com/decentraland/sdk/issues/712
Import self contained models for the moment.
You can import directly via the + Import Button, or with the context menu in the asset catalog.

https://user-images.githubusercontent.com/16782417/234830062-9d44b6ee-9767-45d2-9b53-1c875fcb7ce0.mov


.github/workflows/update-unity-renderer.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           node-version: 16.x
       - uses: actions/checkout@v2
       - name: Update explorer
-        run: npm i @dcl/explorer@latest
+        run: npm i --save-exact @dcl/explorer@latest
```
leanmendoza, line 16: 👏

packages/@dcl/inspector/src/components/EntityInspector/GltfInspector/GltfInspector.tsx
```diff
@@ -23,7 +22,7 @@ const DROP_TYPES = ['project-asset-gltf']
 
 export default withSdk<Props>(
   withContextMenu<WithSdkProps & Props>(({ sdk, entity, contextMenuId }) => {
-    const { files } = useFileSystem()
+    const [files] = useFileSystem()
```
leanmendoza, line 25: Don't get it this change, is there any difference?

leanmendoza (approved):
LGTM

packages/@dcl/inspector/package.json
```diff
@@ -30,6 +27,7 @@
     "react-dnd": "^16.0.1",
     "react-dnd-html5-backend": "^16.0.1",
     "react-dom": "^18.2.0",
+    "react-dropzone": "^14.2.3",
```
nicoecheza, line 30: is it necessary to add a new dep for file dropping? we already have `react-dnd` which supposedly support that 

https://react-dnd.github.io/react-dnd/examples/other/native-files

packages/@dcl/inspector/src/components/Button/Button.css
```diff
@@ -0,0 +1,26 @@
+.Button {
+  border: none;
+  color: #fff;
+  background-color: #4b4852;
```
nicoecheza, line 4: is it possible to use variables for these?

nicoecheza (changes requested):

packages/@dcl/inspector/src/components/EntityInspector/GltfInspector/GltfInspector.tsx
gonpombo8: just for consistensy. We were returning arrays for hooks, and also this is the way useEffect uses.

PR #549, opened 2023-04-27
feat: composite loader (bundle)

leanmendoza:
# What?
At build time:
1. Set a virtual entry point with each original entry point. For this, a loader was implemented that adds the `export * from '@dcl/sdk'` when `customEntryPoint` is not passed.
2. Get the all composites, and if it has at least one, it bundles with the composite provider, instancing hardcoded composite `main.composite` when the `crdtGetState` is comming **WITHOUT** entities (hasEntities=false)

Summary:
- Now, `export * from '@dcl/sdk'` is no longer required
- The `.composite` are automatically bundled with the code and it exposes a provider. The provider can be used to instance composite in runtime, but `main.composite` will be instanced at `onStart` function if there is no `main.crdt` already instanced by the Runtime

# Copilot

### <samp>🤖 Generated by Copilot at 13aee32</samp>

This pull request adds new features to the js-sdk-toolchain to enable custom entry point generation and composite loading for the SDK. It refactors the ci.yml workflow and the sdk exports, and updates the tests and snapshots to match the new functionality. It also removes some unused or redundant code, such as the `with-composite` export and the `createContentFetchCompositeProvider` function.


packages/@dcl/sdk-commands/package.json
```diff
@@ -7,6 +7,7 @@
     "sdk-commands": "./dist/index.js"
   },
   "dependencies": {
+    "@dcl/ecs": "file:../ecs",
```
menduz, line 10: this may cause many trouble because dcl/ecs is ESM

packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -74,8 +74,8 @@ export async function bundleProject(components: BundleComponents, options: Compi
     const entryPointCode = options.customEntryPoint
       ? `export * from '${filePath}'`
```
menduz, line 75:
```suggestion
      ? `export * from '@dcl/sdk'; export * from '${filePath}'`
```

leanmendoza: For `customEntryPoint` you take the responsibility of using `@dcl/sdk` or a modified one.

packages/@dcl/js-runtime/sdk.d.ts
```diff
@@ -0,0 +1,5 @@
+
+// @internal
+declare module '~sdk/all-composites' {
+    export const compositeFromLoader: Record<string, Uint8Array | string>
```
menduz, line 4: I think that importing the json as an object will be _way_ more efficient than the serialized versions. Not in final bundle size, but yes in processing and allocations

```suggestion
    export const compositeFromLoader: Record<string, CompositeType>
```

packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -54,15 +57,47 @@ export async function bundleProject(components: BundleComponents, options: Compi
     throw new CliError(`File ${tsconfig} must exist to compile the Typescript project`)
   }
 
-  const input = globSync(options.single ?? 'src/index.ts', { cwd: options.workingDirectory, absolute: true })
+  const dclFolderPath = path.resolve(process.cwd(), '.decentraland', 'ts-entry-points')
```
menduz, line 60:
```suggestion
  const dclFolderPath = path.resolve(options.workingDirectory, '.decentraland', 'ts-entry-points')
```

packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -170,3 +206,51 @@ function runTypeChecker(components: BundleComponents, options: CompileOptions) {
...
+    if (compositeName.endsWith('.bin')) {
+      compositeLines.push(`'${compositeName}':new Uint8Array(${JSON.stringify(Array.from(bin))})`)
+    } else {
+      const textDecoder = new TextDecoder()
+      const json = JSON.stringify(JSON.parse(textDecoder.decode(bin)))
+      compositeLines.push(`'${compositeName}':'${json}'`)
+    }
+  }
+
+  const contents = `export const compositeFromLoader = {${compositeLines.join(',')}}`
```
menduz, line 224: (cont) pseudo code from my previous suggestion


```suggestion
  // this will contain the parsed composite objects
  const compositeObjects: Record<string, any> = {}

  for (const compositeName in composites) {
    const bin = composites[compositeName]
    const textDecoder = new TextDecoder()
    compositeObjects[compositeName] = JSON.parse(textDecoder.decode(bin))
  }

  const contents = `export const compositeFromLoader: Record<string, CompositeType> = ${JSON.stringify(compositeObjects)}`
```

packages/@dcl/sdk/src/with-composite.ts
```diff
@@ -1,19 +1,30 @@
...
 /* @__PURE__ */ export async function onStart() {
-  const compositeProvider = await createContentFetchCompositeProvider()
-  const mainComposite = compositeProvider.getCompositeOrNull('main.composite')
-  if (mainComposite) {
-    Composite.instance(engine, mainComposite, compositeProvider)
+  const response = await crdtGetState({ data: new Uint8Array() })
+
+  // when this condition is true something like `main.crdt` was pre-loaded from the runtime, we don't need to instance the main.composite
+  if (!response.hasEntities) {
+    const mainComposite = compositeProvider.getCompositeOrNull('main.composite')
```
menduz, line 17: should we make this the default and _only_ behavior? It will be easier to reason if "if there is a main.composite it will always be loaded". that way we won't need the --customEntryPoint

leanmendoza: make sense to me, but `--customEntryPoint` is for those that want to expose their own `onUpdate` and `onStart`

menduz (approved):

PR #563, opened 2023-05-03
feat: spawn `main` function as entrypoint in the entrypoint file`

leanmendoza:
```ts
import { engine, GltfContainer, MeshRenderer, Transform } from '@dcl/sdk/ecs'

// Now this is OK
export function main() {
  const cube = engine.addEntity()
  Transform.create(cube)
  MeshRenderer.setBox(cube)
}

// This is OK
engine.addSystem(() => { })

// This is OK
engine.defineComponent('some::component', Schemas.....)

// This always could introduce conflicts, but this is WRONG now, the right mechanism is inside main()
const cube = engine.addEntity()
Transform.create(cube)
MeshRenderer.setBox(cube)


```


packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -241,15 +241,29 @@ async function getAllComposite(
...
+export * from '@dcl/sdk'
+import { executeTask as _INTERNAL_startupInitialTask } from '@dcl/sdk/ecs'
+declare const main: (() => void) | undefined;
+// END AUTO GENERATED CODE\n\n`
+const postIndexTsCode = `\n\n;
+// BEGIN AUTO GENERATED CODE
+if (main) {
+  _INTERNAL_startupInitialTask(async () => {main()})
+}
+// END AUTO GENERATED CODE\n\n`
```
menduz, line 255: I think we can leverage better ways to do this. Like adopting a `export default function main() {`

and then only importing the entire file from a synthetic .ts file, without concatenating everything.

The `function main()` without any export is something not idiomatic of typescript and we must export it to prevent linters to fail and dead code ellimination

menduz (changes requested):

packages/@dcl/sdk-commands/src/logic/bundle.ts
leanmendoza: Changing the entry-point on the flight could be tricky, `esbuild` doesn't allow having a virtual entry-point, that's why in the previous approach of this I generated the code in `.decentraland/entry-point-something.ts`.
I agree the `export function name` is more idiomatic. (without export is not something that I like)

menduz: AFAIK we can use virtual entry points very easily with a plugin or writing a tmp file, did you discard those options?

leanmendoza: done

gonpombo8 (approved):

PR #569, opened 2023-05-05
Remove self-contained GLTF/GLB asset

nicoecheza:
closes https://github.com/decentraland/sdk/issues/757

May-05-2023 13-21-56

### <samp>🤖 Generated by Copilot at ddee27e</samp>

This pull request adds and updates several components and files in the `@dcl/inspector` package, mainly to improve the UI and functionality of the inspector tool. It introduces the `react-modal` and `classnames` modules as dependencies, and uses them to create a `Modal` component and to apply conditional styles to the `Button` and `Block` components. It also adds the ability to remove an asset from the project, and shows a warning modal if the asset is used by some components. It simplifies some code by using hooks and variables, and fixes some minor issues with the input validation and the context menu.

### DIFF
--- a/packages/@dcl/inspector/src/components/Block/types.ts
+++ b/packages/@dcl/inspector/src/components/Block/types.ts
@@ -1,3 +1,4 @@
 export type Props = {
   label?: string
+  broken?: boolean
 }
--- a/packages/@dcl/inspector/src/components/Block/types.ts
+++ b/packages/@dcl/inspector/src/components/Block/types.ts
@@ -1,4 +1,4 @@
 export type Props = {
   label?: string
-  broken?: boolean
+  error?: boolean
 }
--- a/packages/@dcl/inspector/src/hooks/sdk/useComponentInput.ts
+++ b/packages/@dcl/inspector/src/hooks/sdk/useComponentInput.ts
@@ -29,7 +29,7 @@ export const useComponentInput = <ComponentValueType extends object, InputType e
   )
   const [isFocused, setIsFocused] = useState(false)
   const [skipSync, setSkipSync] = useState(false)
-  const [monitor, setMonitor] = useState({ isValid: false })
+  const [isValid, setIsValid] = useState(false)
 
   const updateInputs = useCallback((value: InputType | null, skipSync = false) => {
     setSkipSync(skipSync)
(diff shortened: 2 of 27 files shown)


packages/@dcl/inspector/src/components/Block/types.ts
```diff
@@ -1,3 +1,4 @@
 export type Props = {
   label?: string
+  broken?: boolean
```
cazala, line 3: usually for these kind of properties we use the name `error` instead of broken, what do you think?

packages/@dcl/inspector/src/hooks/sdk/useComponentInput.ts
```diff
@@ -20,14 +20,16 @@ export const useComponentInput = <ComponentValueType extends object, InputType e
   component: Component<ComponentValueType>,
   fromComponentValueToInput: (componentValue: ComponentValueType) => InputType,
   fromInputToComponentValue: (input: InputType) => ComponentValueType,
-  isValidInput: (input: InputType) => boolean = () => true
+  validateInput: (input: InputType) => boolean = () => true,
+  deps: unknown[] = []
```
cazala, line 24: what do you need these `deps` for? They are passed to the `validateInput` callback dependencies, but they are not used by it.

packages/@dcl/inspector/src/hooks/sdk/useComponentInput.ts
```diff
@@ -20,14 +20,16 @@ export const useComponentInput = <ComponentValueType extends object, InputType e
...
 ) => {
   const [componentValue, setComponentValue, isEqual] = useComponentValue<ComponentValueType>(entity, component)
   const [input, setInput] = useState<InputType | null>(
     componentValue === null ? null : fromComponentValueToInput(componentValue)
   )
   const [isFocused, setIsFocused] = useState(false)
   const [skipSync, setSkipSync] = useState(false)
+  const [monitor, setMonitor] = useState({ isValid: false })
```
cazala, line 32: do we need to wrap this property in this object? The to use it we need to double unwrap it:

```ts
const { getInputProps, monitor: { isValid } } = useComponentInput()
```

When it could simply be another value at the same level:

```ts
const { getInputProps, isValid } = useComponentInput()
```

right?

packages/@dcl/inspector/src/hooks/sdk/useComponentInput.ts
nicoecheza: they are used by the `useEffect` that triggers the monitor setup. In this specific case, is for re-triggering the validation in case the `files` from the `useFileSystem` hook changes

cazala (approved):

cazala (approved):

PR #577, opened 2023-05-10
fix: renderer resizing

nicoecheza:
closes https://github.com/decentraland/sdk/issues/762

May-10-2023 15-43-08

### <samp>🤖 Generated by Copilot at 1941e40</samp>

> _We are the inspectors of doom_
> _We wield the power of the `.Resizable` class_
> _We slice the panel with the `.resize-handle`_
> _We refactor and simplify, we are the masters of CSS_

### DIFF
--- a/packages/@dcl/inspector/src/components/Resizable/Resizable.css
+++ b/packages/@dcl/inspector/src/components/Resizable/Resizable.css
@@ -1,10 +1,10 @@
 .Resizable {
   display: flex;
+  width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


cazala:
is it intended that resizing the window resizes the sidebar too?

nicoecheza:
> is it intended that resizing the window resizes the sidebar too?

mmmm good catch. I will add the `min` width/height prop as a style to the first children on the `Resizable` component so we always maintain those constrains 👍

PR #584, opened 2023-05-15
feat: validate composite before include it in the bundle

leanmendoza:
# What?
- Support CommonJS build for `@dcl/ecs`
- Include `@dcl/ecs` in `sdk-commands`
- At build-time, only composite that can be instanced are added to composite provider
- Double-check with adding a try-catch while instancing the main.composite

### DIFF
--- a/packages/@dcl/ecs/tsconfig.cjs.json
+++ b/packages/@dcl/ecs/tsconfig.cjs.json
@@ -1,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "CommonJS",
+    "outDir": "dist-cjs"
+  }
+}
--- a/packages/@dcl/sdk-commands/src/logic/bundle.ts
+++ b/packages/@dcl/sdk-commands/src/logic/bundle.ts
@@ -12,6 +12,7 @@ import { CliComponents } from '../components'
 import { colors } from '../components/log'
 import { printProgressInfo, printProgressStep } from './beautiful-logs'
 import { CliError } from './error'
+import { getAllComposites } from './composite'
 
 export type BundleComponents = Pick<CliComponents, 'logger' | 'fs'>
 
@@ -203,6 +204,12 @@ function compositeLoader(components: BundleComponents, options: CompileOptions):
             )
             contents = `export const compositeFromLoader = {${data.compositeLines.join(',')}}`
             watchFiles = data.watchFiles
+
+            if (data.withErrors) {
+              components.logger.log(
+                'Warning: some composites are not included because of errors while compiling them. There can be unexpected behavior in the scene, check the errors and try to fix them.'
+              )
+            }
           }
           shouldReload = false
         }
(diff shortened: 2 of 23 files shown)


packages/@dcl/ecs/tsconfig.cjs.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "CommonJS",
+    "outDir": "dist-cjs"
```
gonpombo8, line 5: maybe `dist/cjs` ?

leanmendoza: I'll try it, I guessed there could be some problems, but I didn't tried

menduz: it is better to not mix two module systems inside `dist`, i'd keep dist-cjs

packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -203,6 +204,12 @@ function compositeLoader(components: BundleComponents, options: CompileOptions):
             )
             contents = `export const compositeFromLoader = {${data.compositeLines.join(',')}}`
             watchFiles = data.watchFiles
+
+            if (data.withErrors) {
+              components.logger.log(
```
menduz, line 209: we have a visually impacting "beautiful logger" for errors if you'd like to use it

packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -0,0 +1,54 @@
...
+    const engine = Engine()
+    try {
+      const composite = compositeProvider.getCompositeOrNull(compositeSource)!
+      Composite.instance(engine, composite, compositeProvider)
+      compositeLines.push(`'${composite.src}':${JSON.stringify(Composite.toJson(composite.composite))}`)
+    } catch (err: any) {
+      components.logger.log(`Composite '${compositeSource}' can't be instanced ${err.stack}`)
+      withErrors = true
+    }
+  }
```
menduz, line 51:
```suggestion
  const composites: Record<string, string> = {}

  for (const compositeSource in composites) {
    const engine = Engine()
    try {
      const composite = compositeProvider.getCompositeOrNull(compositeSource)!
      Composite.instance(engine, composite, compositeProvider)
      composites[composite.src] = JSON.stringify(Composite.toJson(composite.composite))
    } catch (err: any) {
      components.logger.log(`Composite '${compositeSource}' can't be instanced ${err.stack}`)
      withErrors = true
    }
  }
```

then `JSON.stringify(composites)`

packages/@dcl/sdk-commands/src/logic/bundle.ts
leanmendoza: I'll do the improvement later

packages/@dcl/sdk-commands/src/logic/composite.ts
leanmendoza: I defer this to another PR

PR #588, opened 2023-05-15
feat: manual save

nicoecheza:
closes https://github.com/decentraland/sdk/issues/767

### <samp>🤖 Generated by Copilot at c9c9151</samp>

This pull request adds a save feature to the inspector tool that allows the user to save the engine state to the file system. It also fixes some CSS and formatting issues in the toolbar component and removes some unused imports and variables from the data layer. It updates the `data-layer.proto` file to include the new `Save` RPC method.

### DIFF
--- a/packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
+++ b/packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
@@ -11,7 +11,7 @@ import { ToolbarButton } from './ToolbarButton'
 import './Toolbar.css'
 
 const Toolbar = withSdk(({ sdk }) => {
-  const [isDirty] = useSave()
+  const [save, isDirty] = useSave()
   const handleInspector = useCallback(() => {
     const { debugLayer } = sdk.scene
     if (debugLayer.isVisible()) {
@@ -31,13 +31,9 @@ const Toolbar = withSdk(({ sdk }) => {
     []
   )
 
-  const handleSave = useCallback(async () => {
-    await sdk.dataLayer.save({})
-  }, [])
-
   return (
     <div className="Toolbar">
-      <ToolbarButton className="save" onClick={handleSave}>
+      <ToolbarButton className="save" onClick={save}>
         {isDirty ? <BiSave /> : <BiBadgeCheck/>}
       </ToolbarButton>
       <ToolbarButton className="undo" onClick={handleUndoRedo(sdk?.dataLayer.undo)}>
--- a/packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
+++ b/packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
@@ -1,13 +1,17 @@
 import { useCallback } from 'react'
-import { BiUndo, BiRedo } from 'react-icons/bi'
+import { BiUndo, BiRedo, BiSave, BiBadgeCheck } from 'react-icons/bi'
 import { RiListSettingsLine } from 'react-icons/ri'
+
+import { fileSystemEvent } from '../../hooks/catalog/useFileSystem'
+import { useSave } from '../../hooks/editor/useSave'
 import { withSdk } from '../../hoc/withSdk'
 import { Gizmos } from './Gizmos'
 import { ToolbarButton } from './ToolbarButton'
+
 import './Toolbar.css'
-import { fileSystemEvent } from '../../hooks/catalog/useFileSystem'
 
 const Toolbar = withSdk(({ sdk }) => {
+  const [isDirty] = useSave()
   const handleInspector = useCallback(() => {
     const { debugLayer } = sdk.scene
     if (debugLayer.isVisible()) {
@@ -27,8 +31,15 @@ const Toolbar = withSdk(({ sdk }) => {
     []
   )
 
+  const handleSave = useCallback(async () => {
+    await sdk.dataLayer.save({})
+  }, [])
+
   return (
     <div className="Toolbar">
+      <ToolbarButton className="save" onClick={handleSave}>
+        {isDirty ? <BiSave /> : <BiBadgeCheck/>}
+      </ToolbarButton>
       <ToolbarButton className="undo" onClick={handleUndoRedo(sdk?.dataLayer.undo)}>
         <BiUndo />
       </ToolbarButton>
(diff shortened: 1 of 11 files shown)


packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
```diff
@@ -27,8 +31,15 @@ const Toolbar = withSdk(({ sdk }) => {
     []
   )
 
+  const handleSave = useCallback(async () => {
+    await sdk.dataLayer.save({})
```
cazala, line 35: what do you think about adding the `save` to the `useSave` hook, so if needed somewhere else it can be retrieved from the `useSave` hook too?

packages/@dcl/inspector/src/hooks/editor/useSave.ts
```diff
@@ -0,0 +1,17 @@
+import mitt from 'mitt'
+import { useState } from 'react'
+
+import { useSdk } from '../sdk/useSdk'
+
+type SaveEvent = { change: boolean }
+export const saveEvent = mitt<SaveEvent>()
+
+/* istanbul ignore next */
```
cazala, line 9: this seems like a very trivial hook to test, why ignore it?

nicoecheza: copy/paste from another hook. Will add test to it 👌

packages/@dcl/inspector/src/components/Toolbar/Toolbar.tsx
nicoecheza: good idea

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -1,14 +1,44 @@
...
+  return function dumpEngine(): CompositeDefinition {
+    // TODO: hardcoded for the moment
+    const composite = dumpEngineToComposite(engine, 'json')
+    // TODO: the ID should be the selected composite id name
+    // composite.id = 'main'
+
+    const mainCrdt = dumpEngineToCrdtCommands(engine)
+    fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error(`Failed saving main.crdt: `, err))
+
+    compositeProvider
```
gonpombo8, line 28: await the promise for the rpc method
```suggestion
try {
  await compositeProvider.save({}))
} catch (e) {
// log
}
```

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -43,25 +74,22 @@ export async function initRpcMethods(
...
-      const mainCrdt = dumpEngineToCrdtCommands(engine)
-      fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error(`Failed saving main.crdt: `, err))
-
-      compositeProvider
-        .save({ src: currentCompositeResourcePath, composite }, 'json')
-        .catch((err) => console.error(`Save composite ${currentCompositeResourcePath} fails: `, err))
-    }
+  onChanges.push(() => {
+    dirty = true
+    saveEvent.emit('change', dirty)
```
gonpombo8, line 79: this is not going to work for the editor since the saveEvent emitter lives on the inspector (client-side), and this code is for the data-layer (server side)

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -123,6 +151,10 @@ export async function initRpcMethods(
         undoRedo.addUndoFile([{ prevValue, newValue: null, path: filePath }])
       }
       return {}
+    },
+    async save() {
+      save()
```
gonpombo8, line 156: await the promise

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
nicoecheza: I'm just moving it to another fn. This is already like this today because we are using it in a system, and systems cannot be thenables...

packages/@dcl/inspector/src/hooks/editor/useSave.ts
```diff
@@ -0,0 +1,21 @@
+import mitt from 'mitt'
+import { useCallback, useState } from 'react'
+
+import { useSdk } from '../sdk/useSdk'
+
+type SaveEvent = { change: boolean }
+export const saveEvent = mitt<SaveEvent>()
+
+export const useSave = (): [() => Promise<any>, boolean] => {
```
gonpombo8, line 9:
```suggestion
export const useSave = (): [() => Promise<void>, boolean] => {
```

packages/@dcl/inspector/src/lib/sdk/operations/index.ts
```diff
@@ -16,7 +23,10 @@ export function createOperations(engine: IEngine) {
     setParent: setParent(engine),
     updateSelectedEntity: updateSelectedEntity(engine),
     removeSelectedEntities: removeSelectedEntities(engine),
-    dispatch: () => engine.update(1)
+    dispatch: ({ dirty = true }: Dispatch = {}) => {
```
gonpombo8, line 26: why we have this ? if it is always true then use it explicit with a comment.

packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -3,12 +3,43 @@ import { Entity, EntityMappingMode, IEngine, Composite, OnChangeFunction, Compos
...
 
+function setupEngineDump(
+  fs: FileSystemInterface,
+  engine: IEngine,
+  compositeProvider: CompositeManager,
+  compositePath: string
+) {
+  return function dumpEngineAndGetComposite(dump: boolean = true): CompositeDefinition {
+    // TODO: hardcoded for the moment
+    const composite = dumpEngineToComposite(engine, 'json')
```
gonpombo8, line 20: we are doing this on every tick even if we didnt change anything. Is that ok ? Previuosly we had the isDirty condition

cazala (approved):

PR #601, opened 2023-05-24
fix: file removing with undo not working properly

nicoecheza:
closes https://github.com/decentraland/sdk/issues/794

copilot:summary


cazala:
I tested this by following the steps in the issue, and I'm still having problems when undoing the removal of a file. 

Here's a video where I create a new entity, I add a transform and an gltf container to it, I assign an asset to the gltf container **that is already used by other entity in the scene**, I delete the asset and both entities that use it dissapear from the renderer ✅ then I undo it, and then only one of the two entities is shown again ❌ 

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/4670da1b-5b44-43fb-8d92-cf6b2be378ba

packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
```diff
@@ -1,13 +1,21 @@
...
 import { SceneContext } from '../SceneContext'
 
-const sceneContextMap = new Map<string /*sceneId*/, WeakRef<SceneContext>>()
+type SceneId = string
+
+const sceneContextMap = new Map<
+  SceneId,
+  {
+    ctx: WeakRef<SceneContext>
+    entities: Set<Entity>
```
cazala, line 16: Don't we have a map of entities already in the `SceneContext`? Why having another one here?

cazala (approved):

PR #604, opened 2023-05-25
move footer to its own component

gonpombo8:
[Justin Sorry]. I have a toc with classes, makes it hard to follow. 

- [ ] refactor undoRedoBuffer to be a function instead of a class
- [ ] move footer to its own component


packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
```diff
@@ -0,0 +1,66 @@
...
+const ONE_GB_IN_BYTES = ONE_MB_IN_BYTES * 1024
+
+export function UndoRedoArray(maxEntries: number, maxSize: number = ONE_GB_IN_BYTES) {
+  const valueList: UndoRedo[] = []
+  let memorySize: number = 0
+
+  function valueSize(val?: UndoRedo) {
+    let fileSize: number = 0
+    if (val?.$case === 'file' && val) {
+      for (const opreation of val.operations) {
```
nicoecheza, line 13: `operation`

packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
```diff
@@ -0,0 +1,66 @@
...
+      // Add current value size to the memory
+      const currentSize = valueSize(value)
+      memorySize += currentSize
+
+      // add the value to the list.
+      valueList.push(value)
+
+      // if we reach the max_size limit, start removing files.
+      while (memorySize >= maxSize) {
+        for (const index in valueList) {
```
nicoecheza, line 38: `for in` doesn't ensure return order for arrays, I think we should use just a `for` loop (or `forEach`, etc)

nicoecheza: and you will be able to avoid casting `index` to a number :P

packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
```diff
@@ -0,0 +1,66 @@
...
+          const indexNumber: number = Number(index)
+          const size = valueSize(valueList[index])
+          if (size) {
+            memorySize -= size
+            valueList.splice(indexNumber, 1)
+            break
```
nicoecheza, line 44: if you break here, you will just remove the first `$case === file` entry, and that could possible not be enough

nicoecheza (changes requested):

nicoecheza:
seems like some context menus are broken 😢 

Screen Shot 2023-05-25 at 10 33 25

Screen Shot 2023-05-25 at 10 33 31

packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
gonpombo8: nope, you have the while loop

packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
```diff
@@ -0,0 +1,66 @@
...
+      memorySize -= valueSize(value)
+      return value
+    },
+    clear() {
+      memorySize = 0
+      valueList.length = 0
+    },
+    values() {
+      return Array.from(valueList)
+    },
```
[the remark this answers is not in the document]
gonpombo8: if we do this, then you can mutate the valueList

packages/@dcl/inspector/src/lib/data-layer/host/utils/undo-redo-array.ts
gonpombo8: mmm are u sure about the order? I think that was for some browsers in the past.
I'll change anyways but if this is the case then we have to refactor in a lot of places.

nicoecheza (approved):

PR #605, opened 2023-05-25
Free camera in the editor

AN-DCL:
Closes decentraland/sdk#796.

Switches camera to a Free mode.

W / S and UP / DOWN move camera forward / backward along its direction.
A / D and LEFT / RIGHT move camera sideways.
When either left or right mouse button is held, mouse move changes camera's direction.

When camera is not moving and is not rotating, mouse wheel "zooms" in and out (to be precise, it just moves camera forward / backward, keeping same FoV).

When camera is moving or rotating (i.e. mouse button is down), mouse wheel changes camera's speed. Whenever that happens, a message appears in the top left corner of renderer's window displaying camera's speed value.


cazala:
A few observations/questions from testing this PR:

1) Using my Macbook's trackpad, the mousewheel speed is very fast. When I zoom in/out using the wheel it's way faster than using the arrows, and when trying to set the camera speed is almost impossible to get a value different than 1 or 100

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/bbffb90f-3e8e-48f2-abcf-8a524ea2ec96

2) Should we limit the camera's ability to go under the ground? Or should we add a way to "reset" the camera position? It happened to me that since the speed of the mouse wheel is so fast for me, I ended up under the ground a few times, and if I move/drag the camera there, I don't see anything above, and then It was imposible to me to go above the ground again. I had to change the scene's parcels to force the camera to point to the origin and only then I was able to zoom back above the ground.

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/b4019e92-03d2-4678-bf52-287a6956fe49

3) This might be just me, and from what I see the current behavior is the default one of the free camera, but I feel like the drag of the mouse is inverted to what I would consider the intuitive way. Currently if there's an object on the right side of the screen and I want to center it, insted of dragging the screen from right to left ◀️ , which to me would be the natural way, I need to drag from left to right ▶️:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/fbdb93ac-78b1-4f5a-a3c4-f8debd555397

AN-DCL:
1. It looks like normalizing mouse wheel behavior across different devices is difficult: https://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers. SO users recommend the approach that is currently implemented in the camera code. That said, they mention that it fails for Mac trackpads. I'm tempted to say that for now we only support mouses :).

2. Good suggestion, committed corresponding change.

3. Oh well, only the setting could heal that issue. We could add it later.

cazala (approved):
Now with the change on the speed it's way easier to zoom for me using the track pad.
What do you think about adding a test for the new `CameraManager`? It's pretty self contained, should be straight forward to do some unit testing on it.

PR #608, opened 2023-05-26
Fix rotation gizmo working incorrectly with world axes

AN-DCL:
There's a bug in Babylon's `PlaneRotationGizmo`'s constructor which breaks rotation about world axes.
We provide patched version of `PlaneRotationGizmo`.
Patched `PlaneRotationGizmo` is used by patched version of Babylon's `RotationGizmo`.
In turn, patched `RotationGizmo` is used by patched version of Babylon's `GizmoManager`.
Finally, patched `GizmoManager` is used by Editor code.


cazala (commented):
Working great now 🚀

cazala (approved):

PR #610, opened 2023-05-26
feat: dump to scene.json

nicoecheza:
### <samp>🤖 Generated by Copilot at b12d1ac</samp>

Refactored and updated the data layer modules to improve the synchronization and persistence of the scene metadata. Added a function to parse scene metadata from a component value and generate a valid `scene.json` file. Modified the `feeded-local-fs.ts` and `rpc-methods.ts` modules to use this function and read and write the `scene.json` file accordingly.

### DIFF
--- a/packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
+++ b/packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
@@ -8,6 +8,7 @@ import { stream } from './stream'
 import { FileOperation, initUndoRedo } from './undo-redo'
 import { minimalComposite } from '../client/feeded-local-fs'
 import upsertAsset from './upsert-asset'
+import { initSceneProvider } from './scene'
 
 function setupEngineDump(
   fs: FileSystemInterface,
@@ -24,7 +25,7 @@ function setupEngineDump(
     if (!dump) return composite
 
     const mainCrdt = dumpEngineToCrdtCommands(engine)
-    fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error(`Failed saving main.crdt: `, err))
+    fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error('Failed saving main.crdt: ', err))
 
     compositeProvider
       .save({ src: compositePath, composite }, 'json')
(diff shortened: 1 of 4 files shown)


packages/@dcl/inspector/src/lib/data-layer/host/rpc-methods.ts
```diff
@@ -24,7 +27,21 @@ function setupEngineDump(
     if (!dump) return composite
 
     const mainCrdt = dumpEngineToCrdtCommands(engine)
-    fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error(`Failed saving main.crdt: `, err))
+    fs.writeFile('main.crdt', Buffer.from(mainCrdt)).catch((err) => console.error('Failed saving main.crdt: ', err))
+
+    fs.readFile('scene.json')
```
gonpombo8, line 32: do we need to do this on every engine.update ?
Can we attach it only if the SceneComponent is dirty or if the onChange cb ?

nicoecheza: done :D

packages/@dcl/inspector/src/lib/data-layer/host/scene.ts
```diff
@@ -0,0 +1,39 @@
...
+      _: Entity,
+      operation: CrdtMessageType,
+      component: ComponentDefinition<unknown> | undefined,
+      componentValue: unknown
+    ) {
+      if (operation === CrdtMessageType.PUT_COMPONENT && component?.componentName === EditorComponentIds.Scene) {
+        fs.readFile('scene.json')
+          .then((content) => {
+            const value = componentValue as EditorComponentsTypes['Scene']
+            console.log('asd', getUpdatedSceneBuffer(content, value))
```
leanmendoza, line 30: console.log

packages/@dcl/inspector/src/lib/data-layer/host/utils/component.ts
```diff
@@ -20,3 +22,12 @@ export function findPrevValue(composite: CompositeDefinition, componentName: str
   }
   return value.data.json
 }
+
+export function parseSceneFromComponent(value: DeepReadonlyObject<EditorComponentsTypes['Scene']>) {
+  return {
+    scene: {
```
leanmendoza, line 28: Then add the `description` and `title`

nicoecheza: yeah, since now we don't provide a way for editing those props, I think we should leave the ones that the `scene.json` already has. We will definitely add those when we update the Scene component

leanmendoza (approved):

PR #612, opened 2023-05-29
Fix child being added twice

AN-DCL:
The problem is caused by `onBlur` event firing before hook's clean up function (which removes `onBlur` listener). When a user presses ENTER, input submits (creating a new child), then react re-renders and removes an input (which fires `onBlur` which submits again), and only then calls hook's clean up.

Furthermore, the blur behavior of search input and "add child" input should be different. For search input we don't want to do anything when blur happens, and for "add child" we want to either submit or cancel. We've chosen the latter in this PR.

### DIFF
--- a/packages/@dcl/inspector/src/components/Input/Input.tsx
+++ b/packages/@dcl/inspector/src/components/Input/Input.tsx
@@ -1,13 +1,13 @@
 import React, { useEffect, useRef, useState } from 'react'
 
-import { PropTypes } from './types'
+import { BlurBehavior, PropTypes } from './types'
 
 import './Input.css'
 
 const submittingKeys = new Set(['Enter'])
 const cancelingKeys = new Set(['Escape', 'Tab'])
 
-const Input = ({ value, onCancel, onSubmit, onChange, placeholder }: PropTypes) => {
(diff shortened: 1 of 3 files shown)


packages/@dcl/inspector/src/components/Input/Input.tsx
```diff
@@ -17,22 +17,18 @@ const Input = ({ value, onCancel, onSubmit, onChange, placeholder }: PropTypes)
...
-    }
-
     const onKeyUp = (e: KeyboardEvent) => {
       if (cancelingKeys.has(e.key)) onCancel && onCancel()
       if (submittingKeys.has(e.key)) onSubmit && onSubmit(getValue())
     }
 
-    const onBlur = (_: Event) => onSubmit && onSubmit(getValue())
+    const onBlur = (_: Event) => {
+      if (blurBehavior == BlurBehavior.CANCEL) onCancel && onCancel()
```
gonpombo8, line 26: what if we add an onBlur property ? 
like
```tsx
<Input onChange={someChangeFn} onCancel={someCancelFn} onBlur={someCancelFn} />
```

AN-DCL: Done.

gonpombo8 (approved):

PR #628, opened 2023-06-06
add Label Component and ReadOnly Component type

gonpombo8:
```tsx
import { engine } from '@dcl/sdk/ecs'
import * as components from '@dcl/ecs/dist/components'

export function main() {
  const Label = components.Label(engine)!
  console.log({ label: engine.getEntityOrNullByLabel('Snow') })
  for (const [entity] of engine.getEntitiesWith(Label)) {
    console.log({ entity })
  }
}
```


packages/@dcl/ecs/src/components/index.ts
```diff
@@ -33,3 +38,17 @@ export const MeshRenderer: LwwComponentGetter<MeshRendererComponentDefinitionExt
...
+/* @__PURE__ */
+export const Label: (
+  engine: Pick<IEngine, 'defineComponentFromSchema' | 'getComponentOrNull'>
+) => ReadOnlyLastWriteWinElementSetComponentDefinition<{ label: string }> = (engine) => {
+  const LabelComponent = engine.getComponentOrNull(
+    'inspector::EntityNode'
+  ) as ReadOnlyLastWriteWinElementSetComponentDefinition<{
+    label: string
+  }>
+  if (!LabelComponent) throw new Error('Label Component not found. Be sure you create this scene with the editor.')
```
menduz, line 52: This component should be part of the SDK! I'd move its definition into the @dcl/sdk or @dcl/ecs repos.

The use case is clear: You need to find an entity, differentiate a single door from the rest to animate it.

gonpombo8: yes and no. The Label Editor has the parent property also 😢 . So if we create this we are going to expose that prop to the user and we dont want that.

menduz: It makes sense that you label an entity in the editor and you can use that same label from the SDK, otherwise this PR wouldn't make sense. "why would be the SDK exposing this function that doesn't work and always throws?" people will ask themselves.

Let's ditch the parenting in the Label component and move it back to the transform. We can consider this message "an architecture decision", for the sake of clarity. Having it separated only adds unnecessary complexity and confuses us creating this exceptional cases.

PR #641, opened 2023-06-12
UI tweaks

AN-DCL:
1. Right mouse click should not select an entity in composite renderer.
2. Display entity's name in entity inspector.
3. Miscellaneous style tweaks in asset inspector.
4. Fix broken mechanism for rejecting imported assets with names that already exist.

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
+++ b/packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
@@ -1,24 +1,39 @@
+import { useEffect, useState } from 'react'
+
 import { Entity } from '@dcl/ecs'
 
 import { useSelectedEntity } from '../../hooks/sdk/useSelectedEntity'
 import { withSdk } from '../../hoc/withSdk'
-import { SdkContextValue } from '../../lib/sdk/context'
+import { SdkContextEvents, SdkContextValue } from '../../lib/sdk/context'
+import { useChange } from '../../hooks/sdk/useChange'
 
 import { SceneInspector } from './SceneInspector'
 import { TransformInspector } from './TransformInspector'
 import { GltfInspector } from './GltfInspector'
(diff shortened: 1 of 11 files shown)


packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
```diff
@@ -1,18 +1,31 @@
...
+    return (
+      <div className="EntityInspector" key={entity}>
+        <div className="entity-label">{label}</div>
+        {inspectors.map((Inspector, index) => entity !== null && <Inspector key={index} entity={entity} />)}
+      </div>
+    )
+  } else {
+    return <div className="EntityInspector"></div>
+  }
+})
```
nicoecheza, line 29: maybe we can avoid returning two different jsx by

```ts
const getLabel = (sdk: SdkContextValue, entity: Entity) => {
  const nameComponent = sdk.components.Name.getOrNull(entity)
  return entity === 0 as Entity ? 'Scene' : (nameComponent ? nameComponent.value : entity.toString())
}

export const EntityInspector = withSdk(({ sdk }) => {
  const entity = useSelectedEntity()
  const inspectors = [SceneInspector, TransformInspector, GltfInspector]

  const label = entity ? getLabel(sdk, entity) : null

  return (
    <div className="EntityInspector" key={entity}>
      <div className="entity-label">{label}</div>
      {inspectors.map((Inspector, index) => entity !== null && <Inspector key={index} entity={entity} />)}
    </div>
  )
})
```

what do you think?

nicoecheza (changes requested):

packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
AN-DCL: Sounds good.

nicoecheza (approved):

PR #643, opened 2023-06-13
fix: main composite to lower case

nicoecheza:
closes https://github.com/decentraland/sdk/issues/851

### DIFF
--- a/packages/@dcl/inspector/src/lib/logic/in-memory-storage.ts
+++ b/packages/@dcl/inspector/src/lib/logic/in-memory-storage.ts
@@ -28,6 +28,10 @@ export function createInMemoryStorage(initialFs: Record<string, Buffer> = {}) {
   }
 }
 
+function normalizePath(path: string) {
+  return path.replace(/^\/+/g, '')
+}
+
 export function createFsInMemory(initialFs: Record<string, Buffer> = {}): FileSystemInterface {
   const fs = createInMemoryStorage(initialFs)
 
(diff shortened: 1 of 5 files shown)


packages/@dcl/inspector/src/lib/data-layer/host/fs-utils.ts
```diff
@@ -41,5 +41,5 @@ export function getFileName(fileName: string, ext: string) {
 
 export function getCurrentCompositePath(sceneProvider: SceneProvider) {
   const scenePath = sceneProvider.getScene().display.title
-  return withAssetDir(`${scenePath}/main.composite`)
+  return withAssetDir(`${scenePath}/main.composite`).toLowerCase()
```
menduz, line 44: don't we have a normalizePath function over there? the effect is the same but the semantic meaning would add a lot to the devEx

menduz: or maybe `withAssetDir` could take care of the normalization

nicoecheza: changed everything to use methods from `path` module instead. We want to allow paths with uppercase letters, but `.composite` files are mandatory to be lowercase

packages/@dcl/inspector/src/lib/logic/in-memory-storage.ts
```diff
@@ -28,27 +28,40 @@ export function createInMemoryStorage(initialFs: Record<string, Buffer> = {}) {
   }
 }
 
+function normalizePath(path: string) {
+  return path.replace(/^\/+/g, '')
```
gonpombo8, line 32: add a comment or an example about this so i dont have to understand the regexp 🙏🏼

gonpombo8 (approved):

PR #644, opened 2023-06-13
add redux for handling data-layer ws (re)connection

gonpombo8:
Removed dataLayer from SDKContextValue. Now it lives in the store.

### Reconnecting

image

### Disconnected. No retries after 5 times

image


packages/@dcl/inspector/src/lib/logic/consume-stream.ts
```diff
@@ -12,9 +12,9 @@ export async function consumeAllMessagesInto(
       }
     }
   } catch (err) {
-    debugger
+    // debugger
```
gonpombo8, line 15: TODO: handle this reconnection so we can continue sending/receiving messages

cazala (approved):

PR #650, opened 2023-06-15
refactor: paths are always absolute

nicoecheza:
fixes assets path on `main.composite` to be always absolute, thus making preview work as expected 

Jun-15-2023 19-45-19


menduz:
Just in case, as per ADR, decentraland filesystem paths start without / and are all lower cased paths.

Being `a/b/c/d.png` resolved to `(ROOT OF THE PROJECT)/a/b/c/d.png`

Relative paths are not allowed. And `/opt/user/decentraland/my-scene/a/b/c/d.png` is illegal.

PR #654, opened 2023-06-19
feat: stable re-parenting

AN-DCL:
Closes decentraland/sdk#766

When re-parenting an entity, it is desirable to preserve its world matrix, so that its position, scale and rotation remain visually identical before and after re-parenting.

`ecs-math` was updated to version `2.0.2` to take advantage of `Matrix.decompose` fix.


packages/@dcl/inspector/src/lib/logic/math.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Matrix } from '@dcl/ecs-math/dist/Matrix'
+
+export function getWorldMatrix(entity: Entity, transformComponent: TransformComponentExtended): Matrix.ReadonlyMatrix {
+  const transform = transformComponent.getOrNull(entity)
+  if (!transform) return Matrix.Identity()
+
+  const localMatrix = Matrix.compose(transform.scale, transform.rotation, transform.position)
+
+  if (!transform.parent) return localMatrix
+  else return Matrix.multiply(localMatrix, getWorldMatrix(transform.parent, transformComponent))
```
nicoecheza, line 11: no need for an `else` since have an early return

nicoecheza (approved):

PR #657, opened 2023-06-20
feat: duplicate entity operation

AN-DCL:
Closes decentraland/sdk#709.

1. Add "Duplicate entity" option to composite inspector's context menu.
2. This operation creates a copy of entity's subtree, with properly adjusted Transform parenting.
3. Duplicate becomes selected.

### DIFF
--- a/packages/@dcl/inspector/src/lib/sdk/operations/duplicate-entity.ts
+++ b/packages/@dcl/inspector/src/lib/sdk/operations/duplicate-entity.ts
@@ -1,0 +1,37 @@
+import { Entity, IEngine, getComponentEntityTree, Transform } from '@dcl/ecs'
+import { isLastWriteWinComponent } from '../../../hooks/sdk/useComponentValue'
+import updateSelectedEntity from './update-selected-entity'
+
+export function duplicateEntity(engine: IEngine) {
+  return function duplicateEntity(entity: Entity) {
+    const originalToDuplicate: Map<Entity, Entity> = new Map()
+    const TransformComponent = engine.getComponent(Transform.componentId) as typeof Transform
+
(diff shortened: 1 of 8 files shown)


packages/@dcl/inspector/src/lib/sdk/operations/duplicate-entity.ts
```diff
@@ -0,0 +1,37 @@
...
+    const TransformComponent = engine.getComponent(Transform.componentId) as typeof Transform
+
+    for (const entityIterator of getComponentEntityTree(engine, entity, TransformComponent)) {
+      const duplicate = engine.addEntity()
+      originalToDuplicate.set(entityIterator, duplicate)
+      // copy values of all components of original entity
+      for (const component of engine.componentsIter()) {
+        if (component.has(entityIterator) && isLastWriteWinComponent(component)) {
+          const componentValue = component.get(entityIterator)
+          component.create(duplicate, JSON.parse(JSON.stringify(componentValue)))
```
gonpombo8, line 17: `JSON.parse(JSON.stringify(componentValue))`
is this necessary ? 🤔

gonpombo8 (approved):

packages/@dcl/inspector/src/lib/sdk/operations/duplicate-entity.ts
menduz: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

AN-DCL: I tried `structuredClone` but Jest run into an error. I think it's only available in Node 16.

PR #662, opened 2023-06-21
add sagas for all data layer methods

gonpombo8:
- [x] Remove all event emitters for fileChange and dirty state (canSave)
- [x] Add sagas for each rpc method (save, undo, redo, importAsset, etc)
- [x] Move assets catalog and dirty engine state to redux (app store)
- [x] Fix types for data-layer rpc host


packages/@dcl/inspector/src/redux/data-layer/index.ts
```diff
@@ -1,10 +1,20 @@
...
 export enum ErrorType {
   Disconnected = 'disconnected',
-  Reconnecting = 'reconnecting'
+  Reconnecting = 'reconnecting',
+  Save = 'save',
+  GetPreferences = 'get-preferences',
+  SetPreferences = 'set-preferences'
+}
+
+let dataLayerInterface: DataLayerRpcClient | undefined
```
gonpombo8, line 14: the idea of this is to not serialize the dataLayer at all. 
All the methods should be called via sagas

packages/@dcl/inspector/src/redux/data-layer/index.ts
```diff
@@ -31,21 +41,26 @@ export const dataLayer = createSlice({
...
+      dataLayerInterface = payload.dataLayer
       state.reconnectAttempts = 0
       state.error = undefined
     },
     error: (state, { payload }: PayloadAction<{ error: ErrorType }>) => {
       console.log('[WS] Error', payload.error)
       state.error = payload.error
-    }
+    },
+    save: () => {},
```
gonpombo8, line 57: we will have an action/saga for each method of the dataLayer

packages/@dcl/sdk-commands/src/commands/start/data-layer/rpc.ts
```diff
@@ -1,31 +1,27 @@
...
   const context: DataLayerContext = {
-    dataLayerHost
+    fs,
+    engine: dataLayerHost.engine
   }
-  const rpcServer = createRpcServer<DataLayerContext>({ logger })
+  const rpcServer = createRpcServer<DataLayerContext>({ logger: components.logger })
 
-  async function rpcHandler(serverPort: RpcServerPort<DataLayerContext>) {
-    // TODO: dataLayer as any
```
gonpombo8, line 25: any type fixed

packages/@dcl/inspector/src/components/Toolbar/Preferences/Preferences.tsx
```diff
@@ -7,39 +7,40 @@ import { ToolbarButton } from '../ToolbarButton'
...
-  }, [autosaveEnabled])
-
-  const FreeCameraInvertRotationIcon = freeCameraInvertRotation ? BiCheckboxChecked : BiCheckbox
-  const AutosaveEnabledIcon = autosaveEnabled ? BiCheckboxChecked : BiCheckbox
+    dispatch(setInspectorPreferences({ autosaveEnabled: !preferences.autosaveEnabled }))
+  }, [preferences.autosaveEnabled])
 
+  const FreeCameraInvertRotationIcon = preferences.freeCameraInvertRotation ? BiCheckboxChecked : BiCheckbox
+  const AutosaveEnabledIcon = preferences.autosaveEnabled ? BiCheckboxChecked : BiCheckbox
+  console.log('BOEDO: Changed', preferences)
```
cazala, line 40: oops

packages/@dcl/inspector/src/hooks/sdk/useSdkContext.ts
```diff
@@ -23,9 +25,9 @@ export const useSdkContext = () => {
   }, [dispatch])
 
   useEffect(() => {
-    if (!catalog || !canvas || sdk || isLoading || !dataLayer) return
+    if (!catalog || !canvas || sdk || isLoading || !dataLayer || !preferences) return
```
cazala, line 28: do we need the `!dataLayer` if it's not used anymore by `createSdkContext`?

packages/@dcl/inspector/src/hooks/sdk/useSdkContext.ts
```diff
@@ -35,7 +37,7 @@ export const useSdkContext = () => {
         setError(e)
       })
       .finally(() => setIsLoading(false))
-  }, [catalog, canvas, sdk, isLoading, dispatch, dataLayer])
+  }, [catalog, canvas, sdk, isLoading, dispatch, dataLayer, preferences])
```
cazala, line 40: Same as above with the `dataLayer` here

packages/@dcl/inspector/src/redux/app/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  name: 'app-state',
+  initialState,
+  reducers: {
+    updateCanSave: (state, { payload }: PayloadAction<{ dirty: boolean }>) => {
+      // TODO: this should check for autoSaveEnabled: !sdk?.preferences.data.autosaveEnabled
+      const isDirty = !state.preferences?.autosaveEnabled && payload.dirty
+      state.canSave = isDirty
+    },
+    updatePreferences: (state, { payload }: PayloadAction<{ preferences: InspectorPreferences }>) => {
+      console.log(' UPDATED ', payload.preferences)
```
cazala, line 26: log

packages/@dcl/inspector/src/redux/app/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+})
+
+export const { updateCanSave, updatePreferences } = appState.actions
+export const getCanSave = (state: RootState): boolean => state.app.canSave
+export const getInspectorPreferences = (state: RootState): InspectorPreferences => {
+  if (!state.app.preferences) {
+    // TODO: send to sentry
+    console.log('invalid inspector preferences')
```
cazala, line 37: log

cazala (approved):

packages/@dcl/inspector/src/hooks/sdk/useSdkContext.ts
gonpombo8: nope, i think that we are covered with the prefences. good catch

packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
```diff
@@ -22,18 +21,19 @@ import { Warnings } from '../Warnings'
...
 const fixedNumber = (val: number) => Math.round(val * 1e2) / 1e2
 
 const Renderer: React.FC = () => {
   const canvasRef = React.useRef<HTMLCanvasElement>(null)
   useRenderer(() => canvasRef)
   const sdk = useSdk()
-  const dataLayer = useAppSelector(getDataLayer)
-
+  const dispatch = useAppDispatch()
   const [isLoading, setIsLoading] = useState(false)
```
cazala, line 33: This loading state might need to be controlled from the saga, otherwise right now it set to `true` and then to `false` right after the action is dispatched (because the dispatch does not await)

packages/@dcl/inspector/src/redux/data-layer/index.ts
```diff
@@ -30,24 +46,48 @@ export const dataLayer = createSlice({
...
+    getInspectorPreferences: () => {},
+    setInspectorPreferences: (_state, _payload: PayloadAction<Partial<InspectorPreferences>>) => {},
+    getAssetCatalog: () => {},
+    undo: () => {},
+    redo: () => {},
+    importAsset: (_state, _payload: PayloadAction<ImportAssetRequest>) => {},
+    removeAsset: (_state, _payload: PayloadAction<Asset>) => {}
```
cazala, line 68: are these above unsued?

gonpombo8: no, this generates the action with that payload type.
So then in the saga you can listen to this importAsset action

packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
gonpombo8: will do it in another pr :)

PR #664, opened 2023-06-21
fix: replace string-based paths for fs.join

nicoecheza:
closes https://github.com/decentraland/sdk/issues/872


gonpombo8 (changes requested):
I think we should handle always posix paths. (assets/scene/main.composite)
And all this logic should be handled on the data-layer side, to avoid future issues/bugs of forgetting to use path.join. 
We are also using posix paths for the gltf files that are stored on the composite and used by the GltfContainer itself.

PR #666, opened 2023-06-21
remove logs / normalize path

gonpombo8:


test/sdk-commands/utils/data-layer-fs.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    test('[Win] should return the directory name of the given path', () => {
+      const input = 'scene\\assets\\main.composite'
+      const expectedOutput = 'scene/assets'
+      const dirname = path.dirname
+      path.dirname = path.win32.dirname
+      expect(fsInterface.dirname(input)).toEqual(expectedOutput)
+      path.dirname = dirname
+    })
+
+    test('should return an empty string if the path is a file name', () => {
```
cazala, line 42: nip tick here, but it should be a `"."` rather than an empty string no?

Makefile
```diff
@@ -56,7 +56,6 @@ lint-fix: sync-deps
 TESTARGS ?= test/
 test:
 	node_modules/.bin/jest --detectOpenHandles --colors $(TESTARGS)
-	make test-inspector
```
cazala, line 59: Why is this removed?

cazala (approved):

test/sdk-commands/utils/data-layer-fs.spec.ts
gonpombo8: yes !

Makefile
gonpombo8: cause when you want to run the inspector test you use the `make test-inspector` command.
and when you want tu run the sdk/ecs test you use the `make test`. Just to save time and ignore the inspecrtor tests when you want to run a single test using the TESTARGS

gonpombo8: this is not being used by the github actions.
there we use the make test-cli, make test-inspector and make test-coverage.

Maybe we can rename it 🤔  like `make test-sdk` instead of just test

PR #670, opened 2023-06-26
chore: e2e tests

cazala:
This PR adds an initial set of E2E tests. They run by using `make test-inspector-e2e` and it uses puppeteer to perform a set of actions (currently only on the `Hierarchy` component). It also adds a bunch of helpers and a `Hierarchy` page object to help writing test cases more easily.


packages/@dcl/inspector/test/e2e/utils/drag-and-drop.ts
```diff
@@ -0,0 +1,94 @@
...
+          bubbles: true,
+          cancelable: true,
+          screenX: destinationX,
+          screenY: destinationY,
+          clientX: destinationX,
+          clientY: destinationY
+        })
+      )
+
+      console.log('COCOCO')
```
nicoecheza: xd

packages/@dcl/inspector/test/e2e/Hierarchy.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+describe('Hierarchy', () => {
+  beforeAll(async () => {
+    await installMouseHelper(page)
+    await page.goto('http://localhost:8000')
+    await App.waitUntilReady()
+  }, 100_000)
+
+  test('create a new entity', async () => {
+    await Hierarchy.addChild(0, 'new entity')
+    await expect(Hierarchy.exists(514)).resolves.toBe(true)
```
nicoecheza, line 14: I'm assuming we are using `514` as first id for a child since we are using the `mainComposite` from the feeded fs file. If that's the case, maybe we can get the number of entities in the engine in the `beforeAll` hook and use that number to compare when adding new childs to the Hierarchy. This way we avoid breaking these tests by changing the mocked composite

nicoecheza (changes requested):

.github/workflows/ci.yml
```diff
@@ -66,6 +66,24 @@ jobs:
...
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: install npm
+        run: npm i -g npm
+      - name: install
+        run: make install
+      - name: build
+        run: make build
+      - name: test-inspector E2E
```
gonpombo8, line 83: what if instead of installing, building all the app from skratch, and creating a http-server we use the package/url that we stored at s3 ?

gonpombo8: https://chore-e2e-tests.js-sdk-toolchain.pages.dev/inspector

packages/@dcl/inspector/jest.config.js
```diff
@@ -1,4 +1,10 @@
 const useTsJestTransform = process.env['TS_JEST_TRANSFORMER'] !== undefined
+const isE2E = process.env['IS_E2E'] !== undefined
+
+if (isE2E) {
+  console.log('Is E2E')
```
gonpombo8, line 5: console

packages/@dcl/inspector/test/e2e/Hierarchy.spec.ts
cazala: good catch, I made added a `getId` helper to grab the id by the entity's label to avoid hardcoding them

.github/workflows/ci.yml
cazala: yup, I pushed some changes to be able to configure the URL via a process variable so we can use that instead, good catch!

nicoecheza (approved):

PR #683, opened 2023-07-13
feat: replace goerli with sepolia

lauti7:
this pr shouldn't be merged until we proceed with the migration


packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -124,21 +124,36 @@ export async function main(options: Options) {
...
+    printProgressInfo(options.components.logger, `Uploading data to: ${about.content.publicUrl}...`)
+    contentURL = about.content.publicUrl
+  } else {
+    printProgressInfo(options.components.logger, `Uploading data to: ${options.args['--target-content']}...`)
+  }
 
   const deployData = { entityId, files: entityFiles, authChain }
   const position = sceneJson.scene.base
-  const network = chainId === ChainId.ETHEREUM_GOERLI ? 'goerli' : 'mainnet'
+  const network = chainId === ChainId.ETHEREUM_SEPOLIA ? 'sepolia' : 'mainnet'
```
fzavalia, line 138: In this cases that is easy to support both testnets I would just do so instead of completely removing Goerli

```suggestion
  const network = chainId === ChainId.ETHEREUM_GOERLI ? 'goerli' : chainId === ChainId.ETHEREUM_SEPOLIA ? 'sepolia' : 'mainnet'
```

fzavalia (changes requested):

fzavalia (commented):
LGTM

Let's wait for the proper reviewer's approval

nicoecheza (changes requested):

gonpombo8 (approved):

packages/@dcl/sdk-commands/package.json
```diff
@@ -14,16 +14,17 @@
     "@dcl/mini-comms": "1.0.1-20230216163137.commit-a4c75be",
     "@dcl/protocol": "1.0.0-5555484353.commit-63c469d",
     "@dcl/rpc": "^1.1.1",
-    "@dcl/schemas": "^7.6.0",
+    "@dcl/schemas": "^8.2.1",
```
gonpombo8, line 17: I think you need to add the sepolia contracts and maybe some more config to the [linker-dapp](https://github.com/decentraland/linker-dapp) and bump it here too.

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
```diff
@@ -123,23 +122,25 @@ export async function main(options: Options) {
...
+  const { client, url } = await getCatalyst(options.args['--target'], options.args['--target-content'])
 
-  printProgressInfo(options.components.logger, `Uploading data to: ${catalyst.getContentUrl()}...`)
+  printProgressInfo(options.components.logger, `Uploading data to: ${url}...`)
 
   const deployData = { entityId, files: entityFiles, authChain }
   const position = sceneJson.scene.base
-  const network = chainId === ChainId.ETHEREUM_GOERLI ? 'goerli' : 'mainnet'
+  const network =
+    chainId === ChainId.ETHEREUM_GOERLI ? 'goerli' : chainId === ChainId.ETHEREUM_SEPOLIA ? 'sepolia' : 'mainnet'
```
gonpombo8, line 132: are we going to still support goerli ? If it has an end we should remove all of it so we dont have death code here. (Just like we did with ropsten)
if we want to support the two testnets for a while, we can add an issue and remove it later.

test/sdk-commands/commands/deploy/index.spec.ts
```diff
@@ -16,7 +17,7 @@ afterEach(() => {
 describe('deploy command', () => {
   it('happy path. Should open the linker-dapp', async () => {
     const components = await initComponents()
-    jest.spyOn(sceneValidations, 'getValidSceneJson').mockResolvedValue(sceneJson)
+    jest.spyOn(sceneValidations, 'getValidSceneJson').mockResolvedValue(sceneJson as Scene)
```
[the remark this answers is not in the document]
lauti7: it doesn't compile without casting

packages/@dcl/sdk-commands/src/commands/deploy/index.ts
lauti7: it's like that due to gamejam, removed now

nicoecheza (approved):

fzavalia (approved):

PR #710, opened 2023-08-17
feat: add UI rpc

cazala:
This PR adds a UI rpc and exports a UI client to control some aspects of the UI of the `@dcl/inspector` remotely.

- Toggle component inspectors
- Toggle panels
- Select tabs
- Toggle gizmos


packages/@dcl/inspector/src/components/EntityInspector/EntityInspector.tsx
```diff
@@ -37,12 +41,19 @@ export const EntityInspector = withSdk(({ sdk }) => {
   }
   useChange(handleUpdate, [entity])
 
-  const inspectors = [SceneInspector, TransformInspector, GltfInspector]
+  const inspectors = [
+    { name: EditorComponentNames.Scene, component: SceneInspector },
+    { name: 'core::Transform', component: TransformInspector },
```
gonpombo8, line 46: maybe we can use the componentName instead of the literal string here in case of a typo or adding a new component with diff name types ? (I remembered this happens to us with the Selector component).
`sdk.components[componentName].componentName`

packages/@dcl/inspector/src/lib/rpc/ui/types.ts
```diff
@@ -0,0 +1,26 @@
+import { AssetsTab, PanelName } from '../../../redux/ui/types'
```
gonpombo8, line 1: amazing, love this approach 🤍

gonpombo8 (approved):

PR #716, opened 2023-08-30
feat: Add the inspector action and trigger components

cyaiox:
This PR adds the inspector's actions and trigger components.

image


packages/@dcl/inspector/src/lib/sdk/components.ts
```diff
@@ -86,5 +94,26 @@ export function createEditorComponents(engine: IEngine): EditorComponents {
...
+      })
+    )
+  })
+
+  const Triggers = engine.defineComponent(EditorComponentNames.Triggers, {
+    value: Schemas.Array(
+      Schemas.Map({
+        type: Schemas.EnumString<AvailableTriggers>(AvailableTriggers, AvailableTriggers.ON_CLICK),
+        entity: Schemas.Optional(Schemas.Entity),
+        action: Schemas.Optional(
```
cazala, line 111: the `action` field of the trigger should be the `name` of the action rather than the `type` (so it should be a `string` instead of an enum).

This is because an entity could have multiple `play_animation` actions, like a door that has `open` and `close` animations.

packages/@dcl/inspector/src/components/EntityInspector/ActionInspector/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface Props {
+  entity: Entity
+}
+
+export enum Actions {
+  PLAY_ANIMATION = 'play_animation',
+  ACTIVATE = 'activate',
+  DEACTIVATE = 'deactivate',
+  GO_TO_END = 'go_to_end',
+  GO_TO_START = 'go_to_start'
```
cazala, line 12: what are the other besides `play_animation`?

packages/@dcl/inspector/src/lib/sdk/components.ts
cyaiox: you're right! updating it

packages/@dcl/inspector/src/components/EntityInspector/ActionInspector/types.ts
cyaiox: Added for testing purposes, removing them

cazala (approved):

PR #719, opened 2023-09-04
Shape Up - Cycle 2: Editor Improvements

nicoecheza:
* [Shape Up - Cycle 2: Editor Improvements](https://www.notion.so/decentraland/Editor-Improvements-More-configuration-via-Editor-UI-f20d47fc0c9f44c38b611b9b42b1511d)

* [New components + fields specification](https://www.notion.so/decentraland/Component-fields-in-Editor-UI-5fb2d1e31fc54890ab19f08dc381578e)

### <samp>🤖 Generated by Copilot at 455b9b7</samp>

This pull request adds new features and improvements to the @dcl/inspector package, which is a tool for inspecting and editing entities in a Decentraland scene. It updates the dependencies, adds support for more entity components, enhances the GLTF inspector, and introduces a new color field component. It also adds some type definitions, utility functions, and unit tests for the new components.

Leftovers:
* Proper numbers fields UI (we already have a design by mel)
* Proper "uvs" UX (currently disabled since there is no UI for this)
* "fontAutoSize" & "lineCount" for TextShape in babylon (I don't know if this is actually working on unity explorer, but...)
* "avatarTexture" & "videoTexture" UI + full implementation (need spec for this)


packages/@dcl/inspector/src/components/EntityInspector/MaterialInspector/Texture/Texture.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <Container label={label} className={cx(label, { hover: isHover })} initialOpen={false}>
+      <Block>
+        <SelectField label="Type" options={TEXTURE_TYPES} {...type} />
+      </Block>
+      <Block ref={drop} error={!isValid}>
+        {type.value === Texture.TT_TEXTURE && <TextField ref={inputRef} label="Path" type="text" {...src} />}
+        {/* {type.value === Texture.TT_AVATAR_TEXTURE && <TextField label="User ID" {...getInputProps(`${texture}.userId`)} />}
+        {type.value === Texture.TT_VIDEO_TEXTURE && <TextField label="Video player entity" {...getInputProps(`${texture}.videoPlayerEntity`)} />} */}
```
cazala, line 59: Are these intended? Should we delete them?

packages/@dcl/inspector/package.json
```diff
@@ -1,12 +1,15 @@
 {
   "name": "@dcl/inspector",
   "version": "0.1.0",
+  "dependencies": {
+    "earcut": "^2.2.4"
```
cazala, line 5: is this being used? If so, let's move it to `devDependencies`

cazala:
There seems to be an issue updating the engine when the EntityInsector is not open. When I add a MeshRenderer component to an entity, it is not rendered until I select it:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/a4aa8b85-e597-4482-b7fe-5a40345ecf3e

cazala:
Also, when I change values of the albedo, emissive and reflectivity colors, every time I change one, the previous one is reseted:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/96801e54-e07f-4ccd-b404-619423ba1bcf

cazala:
Is it okay that the default value for the emissive is 2? It brights a lot, this cube almost looks like an sphere:

Screenshot 2023-09-14 at 16 37 00

I know I can tweak it, but to me it's strange that this is the default value of a newly created material.

cazala:
This looks like an issue on the babylon renderer, we are always rendering the bounding mox of a MeshRenderer component when it is selected, even when it's not a cube. Like here this sphere shows a box around it:

Screenshot 2023-09-14 at 16 55 52

cazala:
I noticed that this PR is not being able to build the Cloudflare Pages deployment. This was fixed on master, you can pull to get it fixed 🙏

cazala:
There's something that I notice when I test this, and is that there are many components that require a texture (or a path to a texture to be precise). This is not an issue on VSCode because the user can access the file system to add the texture, but on the Web Editor this would not be possible. 

Our current import flow only allows the user to import GLB/GLTF files, so there's no way for the Web Editor users to use many of these components.

Im not saying we need to address that issue on this PR, but I think it's important to allow users to import other type of files as as soon as possible after merging this, otherwise they wont be able to fully use the feature

packages/@dcl/inspector/src/components/EntityInspector/MaterialInspector/Texture/Texture.tsx
nicoecheza: yeah, this are ok, just left them for when we have a spec for them

packages/@dcl/inspector/package.json
nicoecheza: removed 👌 (was for a test for 3D text in babylon)

nicoecheza:
@cazala 

1. MeshRenderer not shown on babylon ✅
2. Colors getting reset ✅
3. Emissive default value ✅
4. That's the current default behaviour, but we don't notice it since we only allowed for cubes, thus the boundingBox was "invisible" (but it was there). We can remove it, but I think we should keep it. 🤔
5. Rebased PR ✅
6. Agree on updating the Web Dditor

cazala:
There seems to be an issue with the `TextShape` component. When I apply it to an entity, I can't click anything else that is further away from the camara, as if it is blocking it (even tho they are not behind the text).

After inspecting the normals, you can see there is a huge plane attached to the entity with the TextShape component.

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/69089648-6c96-402d-ac93-dc407516a533

There are also a few CSS issues with the `TextShapeInspector`, some paddings missing:

Screenshot 2023-09-14 at 20 32 45

Also on some components like `MaterialInspector` the TextField when it is on error state, the red strip-lined outline is not centered:
Screenshot 2023-09-14 at 20 34 04

cazala (approved):

PR #722, opened 2023-09-06
feat: Update inspector components styles

cyaiox:
This PR updates the Trigger and Action component styles.

It also uses the new trigger schema defined in the package `@dcl/asset-pack to allow handling more than one action per trigger event.

Actions:
image

Triggers:
image


cazala (approved):
Looking great! I approved it but just a heads up on two tiny style issues:

1. When the new components are collapsed, they have an extra border on the bottom:

Screenshot 2023-09-07 at 09 50 11

2. The wrapper of the dropdown menu seems to have an extra margin on top and the bottom (see the darker gray margins vs the lighter gray menu):

Screenshot 2023-09-07 at 09 50 20

PR #733, opened 2023-09-14
feat: analytics for quests

lauti7:


packages/@dcl/sdk-commands/src/commands/quests/index.ts
```diff
@@ -145,14 +144,16 @@ async function executeCreateSubcommand(
...
       if (res.status === 201) {
-        const questsId = (await res.json()) as { id: string }
-        printSuccess(logger, `> Your Quest: ${quest!.name} was created successfully - ID: ${questsId.id}`, '')
+        const { id: questId } = (await res.json()) as { id: string }
+        components.analytics.track('Quest Created Success', { questId, questName: quest!.name })
+        printSuccess(logger, `> Your Quest: ${quest!.name} was created successfully - ID: ${questId}`, '')
       } else {
-        const error = (await res.json()) as { code: number; message: string }
+        const { code, message } = (await res.json()) as { code: number; message: string }
+        components.analytics.track('Quest Created Failure', { code, message })
```
gonpombo8, line 152: are we sure that we want to send this message ? isn't enough with the code ?
I dont know what this message could be but if we are already sending a code we can then map that failure with the message

gonpombo8 (commented):
my only concern is if its ok that we send a message to segment

gonpombo8 (approved):

PR #736, opened 2023-09-15
feat: Add Trigger Conditional component to the Inspector

cyaiox:
This PR adds the trigger conditional feature to the Trigger Component, allowing users to decide when the Trigger should be executed based on the entity States

image


packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerCondition/TriggerCondition.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <div className="TriggerConditionsContainer">
+      <div className="TriggerConditionsTitle">
+        <span>Trigger Condition</span>
+        <div className="RightContent">
+          <Button className="AddButton" onClick={onAddTriggerCondition}>
+            <AddIcon size={16} />
+          </Button>
+        </div>
+      </div>
+      {trigger.conditions!.map((condition, idx) => {
```
cazala, line 37: I would make sure that the `conditions` here is defined, because the type does not enforce it. It could happen for example that a smart item defined in the catalog has a trigger without the property `conditions` defined and it would break here.

cazala (changes requested):

packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerCondition/TriggerCondition.tsx
cyaiox: good catch! fixed it

cazala (approved):

PR #739, opened 2023-09-18
feat: add quests manager command

lauti7:


packages/@dcl/sdk-commands/package-lock.json
cazala, line 1: ⚠️⚠️⚠️

50k additions 😬 this was only by installing `@dcl/quests-manager`? By looking at the package it seems that it's an already built webapp: https://www.npmjs.com/package/@dcl/quests-manager?activeTab=code

So does it need so many dependencies? Could [some of these](https://github.com/decentraland/quests-manager/blob/main/package.json#L17-L43) be moved to `devDependencies` so we avoid having them installed over here?

cazala (approved):
LGTM

PR #746, opened 2023-09-22
chore: add deprecated flags to api methods

kuruk-mm:
Extract from https://github.com/decentraland/js-sdk-toolchain/pull/743

### DIFF
--- a/scripts/rpc-api-generation/index.ts
+++ b/scripts/rpc-api-generation/index.ts
@@ -108,7 +108,7 @@ async function internalCompile() {
 `
     apisDTsContent += readFileSync(moduleDTsPath).toString()
 
-    // rmSync(apiModuleDirPath, { recursive: true, force: true })
+    rmSync(apiModuleDirPath, { recursive: true, force: true })
   }
 
   writeFileSync(path.resolve(outModulesPath, 'index.d.ts'), apisDTsContent)
--- a/scripts/rpc-api-generation/index.ts
+++ b/scripts/rpc-api-generation/index.ts
@@ -70,7 +70,6 @@ async function internalCompile() {
     for (const [methodName, method] of Object.entries(api.def.methods)) {
       types.add(method.requestType)
       types.add(method.responseType)
-
       functions.push(
         `export declare function ${method.requestStream ? '*' : ''}${methodName}(body: ${
           method.requestType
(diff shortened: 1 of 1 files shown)


scripts/rpc-api-generation/index.ts
```diff
@@ -94,7 +108,7 @@ async function internalCompile() {
 `
     apisDTsContent += readFileSync(moduleDTsPath).toString()
 
-    rmSync(apiModuleDirPath, { recursive: true, force: true })
+    // rmSync(apiModuleDirPath, { recursive: true, force: true })
```
leanmendoza, line 111:
```suggestion
    rmSync(apiModuleDirPath, { recursive: true, force: true })
```

scripts/rpc-api-generation/index.ts
```diff
@@ -54,9 +65,12 @@ async function internalCompile() {
     if (NON_EXPOSED_LIST_NAMES.includes(api.name)) continue
     const types: Set<string> = new Set()
     const functions: string[] = []
+    const deprecatedMethods = deprecatedApiMethods[api.name] || []
+
     for (const [methodName, method] of Object.entries(api.def.methods)) {
       types.add(method.requestType)
       types.add(method.responseType)
+
```
leanmendoza, line 73:
```suggestion
```

scripts/rpc-api-generation/index.ts
```diff
@@ -22,6 +22,14 @@ type ProtoServiceDefinition = {
   >
 }
 
+const deprecatedApiMethods: Record<string, string[]> = {
+  CommunicationsController: ['send'],
+  EngineApi: ['sendBatch', 'subscribe', 'unsubscribe'],
+  EthereumController: ['requirePayment', 'signMessage', 'convertMessageToObject', 'getUserAccount'],
+  Players: ['getPlayerData', 'getPlayersInScene', 'getConnectedPlayers'],
```
pbosio, line 29: should we add some text with the deprecated tag? i.e: "@deprecated use <this other api> instead"?

scripts/rpc-api-generation/index.ts
```diff
@@ -22,6 +22,14 @@ type ProtoServiceDefinition = {
   >
 }
 
+const deprecatedApiMethods: Record<string, string[]> = {
```
gonpombo8, line 25: wait, why are we duplicating all the depecreated methods here ?
we already have this info on the .proto file. We can do some regex to find all the methods that are depcreated.
This way we have only one source of truth

gonpombo8 (changes requested):

PR #758, opened 2023-10-02
feat: Add tweens component

cyaiox:


packages/@dcl/inspector/src/components/EntityInspector/ActionInspector/ActionInspector.tsx
```diff
@@ -144,13 +153,18 @@ export default withSdk<Props>(
...
+      return actions.some((action) => action.type === ActionType.START_TWEEN)
+    }, [actions])
+
     // actions that may only be available under certain circumstances
     const conditionalActions: Partial<Record<string, () => boolean>> = useMemo(
       () => ({
         [ActionType.PLAY_ANIMATION]: () => hasAnimations,
-        [ActionType.SET_STATE]: () => hasStates
+        [ActionType.SET_STATE]: () => hasStates,
+        [ActionType.START_TWEEN]: () => hasTweens
```
cazala, line 165: mm I think the `start_tween` should not be conditional? You can always use the start tween on an entity. Also, the `hasTweens` will always be `false` for a brand new entity, so you would never be able to add tween actions to it. This seems to have slipped from the original version with the `Tweens` component.

cazala (changes requested):

packages/@dcl/inspector/src/components/EntityInspector/ActionInspector/ActionInspector.tsx
cyaiox: You're right, updated!

PR #759, opened 2023-10-02
feat: remove entity ids

cazala:
Use our own ids instead of relying on entity ids which could change between loads of the composite


packages/@dcl/inspector/src/lib/sdk/operations/add-child.ts
```diff
@@ -18,35 +22,72 @@ export function addChild(engine: IEngine) {
     Transform.create(child, { parent })
     // update Nodes component
     Nodes.createOrReplace(engine.RootEntity, { value: pushChild(engine, parent, child) })
+    if (components) {
+      const getComponent = <T>(componentName: ComponentName) =>
+        components[componentName] ? ({ ...components[componentName] } as T) : null
+      console.log('is smart')
```
cyaiox, line 28: wild `console.log` appeared!

cyaiox (approved):

PR #762, opened 2023-10-03
feat: Add Visibility action

cyaiox:
This PR adds the `VisibilityComponent` and the `set_visibility` action to make an entity visible or invisible.

It also adds the Triggers selectors:
- On Player Enters Area
- On Player Leaves Area

image

image


packages/@dcl/inspector/src/lib/sdk/operations/add-asset.ts
```diff
@@ -96,6 +97,11 @@ export function addAsset(engine: IEngine) {
             values.set(componentName, { ...audioSource, src: audioSource.audioClipUrl.replace('{assetPath}', base) })
             break
           }
+          case CoreComponents.VISIBILITY_COMPONENT: {
```
cazala, line 100: This case is not necessary, this step is to make changes to the values in the components (like replacing the {self} ids or the paths with {assetPath}) but for this component there’s no need to make modifications to the values

cazala (changes requested):

packages/@dcl/inspector/src/lib/sdk/operations/add-asset.ts
cyaiox: @cazala thanks for the review, updated!

cyaiox:
> @cyaiox can we make the default value of visible true when you add one?
> 
> Or maybe it's worth a debate, but I feel like it's potentially confusing. It would be great also to have a tooltip there, explaining that the item will remain visible in the editor, but not when running the scene.

@nearnshaw -- default behavior added :)

cazala (approved):

PR #765, opened 2023-10-04
fix: upgrade @dcl/asset-packs

cazala:
This new version of `@dcl/asset-packs` removes the dependency that seems to be causing the recent issues when building scenes with the SDK.


pravusjif (approved):
This solves the problem we were having when initializing a scene in preview mode

PR #777, opened 2023-10-10
fix: delete item after refresh

cazala:
Fixes https://github.com/decentraland/sdk/issues/946

### DIFF
--- a/packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
+++ b/packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
@@ -49,11 +49,13 @@ const Renderer: React.FC = () => {
         const sceneEntity = sdk.sceneContext.getEntityOrNull(entity)
         if (!sceneEntity) continue
 
-        if (!fileSet.has(value.src)) removeGltf(sceneEntity)
-        else loadGltf(sceneEntity, value.src)
+        removeGltf(sceneEntity)
+        if (fileSet.has(value.src)) {
+          loadGltf(sceneEntity, value.src)
+        }
       }
     }
-  }, [files])
+  }, [files, init])
 
   useEffect(() => {
     if (sdk) {
--- a/packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
+++ b/packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
@@ -8,7 +8,7 @@ import type { ComponentOperation } from '../component-operations'
 import { EcsEntity } from '../EcsEntity'
 import { SceneContext } from '../SceneContext'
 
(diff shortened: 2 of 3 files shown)


packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
```diff
@@ -8,7 +8,7 @@ import type { ComponentOperation } from '../component-operations'
 import { EcsEntity } from '../EcsEntity'
 import { SceneContext } from '../SceneContext'
 
-const sceneContextMap = new Map<string /*sceneId*/, WeakRef<SceneContext>>()
+let sceneContext: WeakRef<SceneContext>
```
cazala, line 11: This was converted into a single reference instead of a map, since the scene id is always the same for the inspector, this came from a copypaste from the babylon explorer

packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
```diff
@@ -49,11 +49,13 @@ const Renderer: React.FC = () => {
         const sceneEntity = sdk.sceneContext.getEntityOrNull(entity)
         if (!sceneEntity) continue
 
-        if (!fileSet.has(value.src)) removeGltf(sceneEntity)
-        else loadGltf(sceneEntity, value.src)
+        removeGltf(sceneEntity)
+        if (fileSet.has(value.src)) {
+          loadGltf(sceneEntity, value.src)
+        }
```
cazala, line 55: I changed this logic to also remove the GLTF before loading it, to match the logic in babylon

packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
```diff
@@ -125,6 +124,13 @@ async function tryLoadGltfAsync(sceneId: string, entity: EcsEntity, filePath: st
     (assetContainer) => {
       processGLTFAssetContainer(assetContainer, entity)
 
+      // remove old entities
+      const prevChildren = entity.getChildren()
+      for (const child of prevChildren) {
+        child.setEnabled(false)
+        child.dispose(false, true)
+      }
```
cazala, line 132: The issue was that when the scene is reloaded, the gltfs are loaded from two places, a side effect in the Renderer.tsx component, and the CRDT messages coming from the engine. I dunno what is the reason for the side effect in the Renderer component but by looking at the blame it came to fix an issue with the undo/redo, so I didn't remove it, but this caused GLTFs to be loaded twice when the scene is reloaded for every GTLFContainer. What I did was adding the piece above that empties an entity when loading a GLTF into it, so if it's loaded twice, the second time will clear the previous one. This previous load was the one that remained after deleting the entity causing the problem described in the issue.

cyaiox (approved):

PR #779, opened 2023-10-11
add polyfill only when we use ethereum provider

gonpombo8:
### What ?
Add text-encoding polyfill when using @dcl/sdk/etherum-provider

### WHY ? 
We need the TextEncoder & TextDecoder functions when we try to use some methods from a contract using the ethereum-provider that the sdk exports.


pravusjif (approved):
LGTM!

PR #782, opened 2023-10-11
feat: Add analytics

cyaiox:
This PR adds tracking to the drag&drop assets action and adds components.

Closes https://github.com/decentraland/sdk/issues/950


packages/@dcl/inspector/src/lib/logic/config.ts
```diff
@@ -16,6 +17,10 @@ export const CONTENT_URL = version.includes('commit')
   ? 'https://builder-items.decentraland.zone'
   : 'https://builder-items.decentraland.org'
 
+export const SEGMENT_KEY = version.includes('commit')
+  ? 'RmfAUVW7o1EEuvYJGqmVjK9q48BUpnCr'
+  : '8PrGcB0XsogabYgo02gXqfqVM20KKzva'
```
cazala, line 22: we should not have these defaults here, they should be passed from the parent app (builder or vscode), and if none is provided, the tracking should be disabled.

packages/@dcl/inspector/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,54 @@
...
+  [Event.REMOVE_ITEM]: {
+    name: string
+  }
+  [Event.ADD_COMPONENT]: {
+    type: string
+    parentEntityId: number
+  }
+  [Event.REMOVE_COMPONENT]: {
+    type: string
+    parentEntityId: number
```
cazala, line 27: We should remove the entity ids from events since that's not really relevant for the data team. 

I would rename the `type` property in the ADD_COMPONENT and REMOVE_COMPONENT to be `componentName` to match the name of the property in the component.

The x,y,z information is not necessary for the ADD_ITEM event, but we need to provide some information about the item itself, like the item id and item name.

The REMOVE_ITEM should be REMOVE_ENTITY, and it does not need to have any payload. We are also missing the ADD_ENTITY and RENAME_ENTITY events.

cazala (changes requested):

packages/@dcl/inspector/package.json
```diff
@@ -2,7 +2,8 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.2.5"
+    "@dcl/asset-packs": "^1.2.5",
+    "uuid": "^9.0.1"
```
cazala, line 6: Could this be a `devDependency`?

packages/@dcl/inspector/src/lib/logic/analytics.ts
```diff
@@ -0,0 +1,91 @@
...
+        traits = {
+          ...traits,
+          appId: config.segmentAppId
+        }
+      }
+
+      this.analytics.identify({
+        userId: this.userId,
+        traits
+      })
```
cazala, line 68: is this the correct signature? the docs show it like this:

```ts
analytics.identify(userId: string, traits: object)
```

https://segment.com/docs/connections/spec/identify/

packages/@dcl/inspector/src/lib/logic/analytics.ts
```diff
@@ -0,0 +1,91 @@
...
+    itemId: string
+    itemName: string
+  }
+  [Event.REMOVE_ITEM]: {
+    itemId: string
+    itemName: string
+  }
+  [Event.ADD_COMPONENT]: {
+    componentName: string
+    parentItemId: string
```
cazala, line 23: i don't think we need the `itemId` in the `ADD_COMPONENT` and `REMOVE_COMPONENT` events, we could simplify a lot by removing them.

packages/@dcl/inspector/src/lib/logic/analytics.ts
```diff
@@ -0,0 +1,91 @@
...
+  ADD_COMPONENT = 'Add Component',
+  REMOVE_COMPONENT = 'Remove Component'
+}
+
+export type Events = {
+  [Event.ADD_ITEM]: {
+    itemId: string
+    itemName: string
+  }
+  [Event.REMOVE_ITEM]: {
```
cazala, line 17: This `REMOVE_ITEM` event doesnt seem to be tracked anywhere

cazala (approved):

PR #783, opened 2023-10-11
fix: ensure `node_modules/**` is ignored to avoid delays trying to pr…

kuruk-mm:
…ocess it

The .dclignore from SDK7 Template has an incorrect format for `node_modules` ignore.

It's using `node_modules` and it should be `node_modules/**`.

This makes the `globSync` to take some seconds to process, doing the hot-reload very slow.

This PR ensures that the `node_modules/**` is ignored from the ContentMapping generation in the `globSync` function.

Required for acceptable performance on: https://github.com/decentraland/unity-renderer/pull/5782


nicoecheza:
I think it might be better if we add it to [here](https://github.com/decentraland/js-sdk-toolchain/blob/fe88e6c95b93e2eddcf1c3ace0dd3abeba5219e7/packages/@dcl/sdk-commands/src/logic/dcl-ignore.ts#L47) instead

nicoecheza (approved):

PR #792, opened 2023-10-17
feat: UITransform pointer filter property

pravusjif:
* Updated protocol
* Exposed new PointerFilter property in UiTransform

-------------------
Related PRs:
* https://github.com/decentraland/protocol/pull/180
* https://github.com/decentraland/unity-renderer/pull/5819

### DIFF
--- a/packages/@dcl/inspector/package-lock.json
+++ b/packages/@dcl/inspector/package-lock.json
@@ -30,6 +30,7 @@
         "@types/react-dom": "^18.0.10",
         "@types/react-modal": "^3.16.0",
         "@types/redux-saga": "^0.10.5",
+        "@types/uuid": "^9.0.5",
         "@vscode/webview-ui-toolkit": "^1.2.2",
         "@well-known-components/pushable-channel": "^1.0.3",
         "classnames": "^2.3.2",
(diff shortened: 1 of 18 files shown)


packages/@dcl/react-ecs/src/components/uiTransform/types.ts
```diff
@@ -124,4 +126,6 @@ export interface UiTransformProps {
   flexShrink?: number
   /** The overflow property controls what happens to content that is too big to fit into an area */
   overflow?: OverflowType
+  /** The pointer filter mode that specifies if the element should block the pointer or not (elements with pointer events always block the pointer regardless of this property) **/
+  pointerFilter?: PointerFilterMode | undefined
```
gonpombo8, line 130: LGTM but maybe we can create a helper to avoid using the enums here, like we do in the overflow line above.
instead of `pointerFilter: PointerFilterMode.PFM_Block` do something like, `pointerFioter: 'block' | 'none'`

pravusjif: Cool, sounds good to me, let me give it a try.

pravusjif: Done!

gonpombo8 (approved):

PR #796, opened 2023-10-20
feat: ui transform auto-size

pravusjif:
* Exposed 'auto' option for width/height of ui transform

Related PR: https://github.com/decentraland/unity-renderer/pull/5836


packages/@dcl/react-ecs/src/components/uiTransform/utils.ts
```diff
@@ -52,6 +52,10 @@ function parsePositionUnit(val?: PositionUnit): [number | undefined, YGUnit] {
     return [undefined, YGUnit.YGU_UNDEFINED]
   }
 
+  if (val === 'auto') {
+    return [Number(val), YGUnit.YGU_AUTO]
```
gonpombo8, line 56: Number('auto') will return NaN 🤔 .
 Maybe this needs to return [undefined, YGUint.YGU_AUTO] ?

gonpombo8 (approved):

PR #808, opened 2023-10-25
feat: Inspector UI Components

cyaiox:
This is an integration PR for the Editor UI Components initiative.

Project: https://github.com/orgs/decentraland/projects/53/views/1


cazala:
Looking great! 💅🏻✨

A just a couple things I noticed:

1) The dropdowns seem to have a "Select X" option at the top that is always selectable. Like "Select a state" in the `set_state` action, same in the "Animations" dropdown in `play_animation`, and so forth.

2) The "Path" property of the `play_sound` action is a file (path to a file) but it's not using the new FileUploader component, it's still a regular TextField with drag n drop. I dunno if that was part of the scope or not, but just wanted to give a heads up in case it was missed.

cazala (approved):

PR #819, opened 2023-11-01
fix: Trigger component conditions validation

nicoecheza:
closes https://github.com/decentraland/sdk/issues/982

### <samp>🤖 Generated by Copilot at 1cbe795</samp>

Improved the validation and filtering of trigger actions in the `TriggerInspector` component

### DIFF
--- a/packages/@dcl/inspector/package.json
+++ b/packages/@dcl/inspector/package.json
@@ -2,7 +2,7 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.2.5"
+    "@dcl/asset-packs": "^1.2.8"
   },
   "devDependencies": {
     "@babylonjs/core": "^6.18.0",
--- a/packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerInspector.tsx
+++ b/packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerInspector.tsx
@@ -71,16 +71,13 @@ export default withSdk<Props>(
     const hasCounter = useHasComponent(entityId, Counter)
 
     const areValidActions = useCallback(
-      (updatedActions: TriggerAction[]) =>
-        updatedActions.length > 0 && updatedActions.every((action) => action.id && action.name),
+      (updatedActions: TriggerAction[]) => updatedActions.every((action) => action.id && action.name),
       []
     )
 
     const areValidConditions = useCallback(
-      (updatedConditions: TriggerCondition[] | undefined) =>
-        updatedConditions && updatedConditions.length > 0
-          ? updatedConditions.every((condition) => condition.id && condition.value)
-          : true,
+      (updatedConditions: TriggerCondition[] = []) =>
+        updatedConditions.every((condition) => condition.id && condition.value),
       []
     )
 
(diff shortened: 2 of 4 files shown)


packages/@dcl/inspector/src/lib/sdk/operations/add-asset.ts
```diff
@@ -108,10 +108,12 @@ export function addAsset(engine: IEngine) {
                 ...condition,
                 id: mapId(condition.id)
               })),
-              actions: trigger.actions.map((action) => ({
-                ...action,
-                id: mapId(action.id)
-              }))
+              actions: trigger.actions
+                .filter((action) => action.id)
```
cazala, line 112: which asset contains actions without an id?

packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerInspector.tsx
```diff
@@ -71,17 +71,14 @@ export default withSdk<Props>(
...
 
     const areValidConditions = useCallback(
-      (updatedConditions: TriggerCondition[] | undefined) =>
-        updatedConditions && updatedConditions.length > 0
-          ? updatedConditions.every((condition) => condition.id && condition.value)
-          : true,
-      []
+      (updatedConditions: TriggerCondition[] = []) =>
+        updatedConditions.every((condition) => condition.id && condition.value),
+      [triggers]
```
cazala, line 81: why is it necessary to add `triggers` as a dependency if it's not used in the scope of the callbacks?

packages/@dcl/inspector/src/lib/sdk/operations/add-asset.ts
nicoecheza: trigger area smart item comes "empty"

Screen Shot 2023-11-01 at 10 42 40

packages/@dcl/inspector/src/components/EntityInspector/TriggerInspector/TriggerInspector.tsx
nicoecheza: good catch, was testing something and left this out, removed!

packages/@dcl/inspector/src/lib/sdk/operations/add-asset.ts
cazala: mm i think it would be better to remove those from the asset rather than filtering them here

cazala: https://github.com/decentraland/asset-packs/blob/main/packs/smart_items/assets/trigger_area/data.json#L23-L26

cazala: it would require a PR + release over there, and then upgrade the dependency over here

nicoecheza: removed empty actions from those components + added a validation rule for future cases:
https://github.com/decentraland/asset-packs/pull/48

packages/@dcl/inspector/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.2.5"
+    "@dcl/asset-packs": "^1.2.7-20231102125017.commit-3f47c4a"
```
cazala, line 5: please update to the `latest` version

PR #826, opened 2023-11-03
fix: Inspector component styles

cyaiox:
This PR fixes some style issues for the inspector components.

- Update LeftPanel default and min size
- Update RightPanel default and min size
- Increase the font size for the selected  entity name in the RightPanel
- Increase the font size for the component name in the RightPanel
- Adds a contextual menu for the components in the RightPanel


packages/@dcl/inspector/src/components/EntityInspector/ActionInspector/ActionInspector.tsx
```diff
@@ -61,14 +57,13 @@ const ActionMapOption: Record<string, string> = {
 }
 
 export default withSdk<Props>(
-  withContextMenu<Props & WithSdkProps>(({ sdk, entity: entityId, contextMenuId }) => {
+  withContextMenu<Props & WithSdkProps>(({ sdk, entity: entityId }) => {
```
cazala, line 60: do we need the `withContextMenu` higher order component if we are not using the `contextMenuId` prop?

cazala (approved):

PR #832, opened 2023-11-08
fix: Inspector FileUploadField

cyaiox:
This PR fixes the uploading of files using the file explorer.

https://github.com/decentraland/js-sdk-toolchain/assets/3170051/9e8ae907-5a3d-4f54-b746-5a819005d9b5


cazala (commented):
I'm seeing this error in console when importing this file (I had to zip it to post it here but unzip it before importing it):
[Floor.glb.zip](https://github.com/decentraland/js-sdk-toolchain/files/13299537/Floor.glb.zip)

```
2.16d4c2fe.chunk.js:2 Uncaught (in promise) LoadFileError: /images/0/uri: Failed to load 'Floor_Grass01.png.png': 0 
    at t [as constructor] (2.16d4c2fe.chunk.js:2:2676024)
    at new t (2.16d4c2fe.chunk.js:2:2676228)
    at 2.16d4c2fe.chunk.js:2:4001910
    at 2.16d4c2fe.chunk.js:2:2679270
    at XMLHttpRequest.r (2.16d4c2fe.chunk.js:2:2680421)
wearable-preview.decentraland.org/:1 Uncaught SyntaxError: Unexpected token 'g', "glTFt"... is not valid JSON
    at JSON.parse (<anonymous>)
    at e._parseJson (2.16d4c2fe.chunk.js:2:3930850)
    at 2.16d4c2fe.chunk.js:2:3925239
    at XMLHttpRequest.r (2.16d4c2fe.chunk.js:2:2680142)
```

The issue is related to that GLB using an external texture which is not present. To support this use case we would need to be able to upload multiple files (like a zip file with both glb + png). I know that's not a supported use case but just wanted to give a heads up because this fails silently and the button just doesn't do anything

PR #835, opened 2023-11-09
feat: Update Inspector components

cyaiox:
This is an integration PR to update the following component's style:

- [x] Gltf Inspector Component
- [x] Material Inspector Component
- [x] Text Shape Inspector Component
- [x] Mesh Collider Inspector Component
- [x] Mesh Renderer Inspector Component
- [x] Action Inspector Component
- [x] Trigger Inspector Component


packages/@dcl/inspector/src/components/Container/types.ts
```diff
@@ -4,5 +4,6 @@ export type Props = {
   rightContent?: JSX.Element
   initialOpen?: boolean
   indicator?: boolean | string | JSX.Element
+  borderer?: boolean
```
cazala, line 7: is this right? shouldn’t be “border”?

cyaiox: @cazala fixed!

cazala (approved):

PR #841, opened 2023-11-14
feat: Update inspector tracking events

cyaiox:
This PR updates the inspector tracking events, adding the property: `isSmart` when adding a new item to the scene and the property `projectId` to all the events, allowing track scenes easily

### DIFF
--- a/packages/@dcl/inspector/src/lib/logic/catalog.ts
+++ b/packages/@dcl/inspector/src/lib/logic/catalog.ts
@@ -57,11 +57,10 @@ export function getAssetsByCategory(assets: Asset[]) {
 export function isSmart(asset: Partial<Asset>) {
   const components = Object.keys(asset?.components ?? {})
   return (
-    (components.length > 1 &&
-      components.some((component) =>
-        [ComponentName.ACTIONS, ComponentName.TRIGGERS].find(($) => $.toString() === component)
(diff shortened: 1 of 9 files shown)


packages/@dcl/inspector/src/lib/logic/catalog.ts
```diff
@@ -54,9 +54,15 @@ export function getAssetsByCategory(assets: Asset[]) {
...
-  const components = Object.keys(asset.components)
-  return components.length > 1 || (components.length === 1 && components[0] !== CoreComponents.GLTF_CONTAINER)
+export function isSmart(asset: Partial<Asset>) {
+  const components = Object.keys(asset?.components ?? {})
+  return (
+    (components.length > 1 &&
+      components.some((component) =>
+        [ComponentName.ACTIONS, ComponentName.TRIGGERS].find(($) => $.toString() === component)
+      )) ||
+    (components.length === 1 && components[0] !== CoreComponents.GLTF_CONTAINER)
```
gonpombo8, line 64: i dont get this logic. If components.length === 1 and its not a gltf container its a smart item ?
what if you create an empty entity with a transform ? (sometimes they use this to agroup entities)

gonpombo8 (approved):

PR #842, opened 2023-11-16
New editor components: VideoPlayer, AudioStream, NftShape, Animator & PointerEvents 

nicoecheza:
[Shape up Cycle 3](https://www.notion.so/decentraland/New-Smart-Items-Videos-NFTs-more-5db47fab15c54b74b1126dafd26d7082)

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 79989a1</samp>

This pull request adds new inspector components for video, audio, NFT, animation and pointer events entities in the scene editor. It also updates the texture inspector component to support video textures. It includes the corresponding TypeScript types, utility functions and unit tests for each inspector component. It modifies the files `EntityInspector.tsx`, `Texture.tsx` and `types.ts` to import and render the new components and options.

### How to test
1. AudioStream, VideoPlayer & NftShape are smart components. So you can find them in the catalog + drag & drop them into the renderer
2. Animator component is automatically created whenever a dropped model has any animation defined on it. 
3. PointerEvents it's a normal component, you can add it through the context menu

### Assets/urls used for testing
AudioStream => https://icecast.ravepartyradio.org/ravepartyradio-192.mp3
VideoPlayer => 
* as file: https://github.com/decentraland/js-sdk-toolchain/assets/2366069/94f943fb-bf3b-4684-8e65-4cb26ba3281a 
* as URL: https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4

Animator => [shark.glb.zip](https://github.com/decentraland/js-sdk-toolchain/files/13515350/shark.glb.zip)
NftShape => `urn:decentraland:ethereum:erc721:0x06012c8cf97bead5deae237070f9587f8e7a266d:558536`


packages/@dcl/inspector/src/lib/sdk/operations/add-asset/index.ts
```diff
@@ -80,6 +83,36 @@ export function addAsset(engine: IEngine) {
...
+            break
+          }
+          case CoreComponents.AUDIO_STREAM: {
+            const audioStream = values.get(componentName) as PBAudioStream
+            values.set(componentName, audioStream)
+            break
+          }
+          case CoreComponents.NFT_SHAPE: {
+            const nft = values.get(componentName) as PBNftShape
+            values.set(componentName, nft)
```
cazala, line 113: the cases above `MESH_RENDERER`, `VISIBILITY_COMPONENT`, `AUDIO_STREAM` and `AUDIO_STREAM` are not making any modifications to the values, so they can be removed (they are getting the value and then setting it again without changes).

This section is only for making modifications to values.

packages/@dcl/inspector/src/lib/data-layer/host/fs-utils.ts
```diff
@@ -33,7 +33,18 @@ export const DIRECTORY = {
...
+  '.glb',
+  '.png',
+  '.composite',
+  '.composite.bin',
+  '.gltf',
+  '.jpg',
+  '.mp3',
+  '.ogg',
+  '.wav',
+  'mp4'
```
cyaiox, line 46:
```suggestion
  '.mp4'
```

cazala (changes requested):
I can confirm that the issue I reported above gets fixed by avoiding the creation of the `Animator` if the entity already has one. I sent a PR to `@dcl/asset-packs` to fix it: https://github.com/decentraland/asset-packs/pull/76.

I found another issue now.

The `AudioStream` component works fine when setting the URL in the component, but if you uncheck the `Start playing` property and instead use the `Play Audio Stream` action to start it, the `url` in the action's payload is `''`. 

empty

That throws an error saying that the url is empty:

Screenshot 2023-11-30 at 14 31 53

Also important, the volume is being sent as a string, that doesn't seem to be right either. The type says it should be number:

Screenshot 2023-11-30 at 14 43 48

cazala:
> I can confirm that the issue I reported above gets fixed by avoiding the creation of the `Animator` if the entity already has one. I sent a PR to `@dcl/asset-packs` to fix it: [decentraland/asset-packs#76](https://github.com/decentraland/asset-packs/pull/76).
> 
> I found another issue now.
> 
> The `AudioStream` component works fine when setting the URL in the component, but if you uncheck the `Start playing` property and instead use the `Play Audio Stream` action to start it, the `url` in the action's payload is `''`.
> 
> empty
> 
> That throws an error saying that the url is empty:
> 
> Screenshot 2023-11-30 at 14 31 53
> Also important, the volume is being sent as a string, that doesn't seem to be right either. The type says it should be number:
> 
> Screenshot 2023-11-30 at 14 43 48

cc @cyaiox I think this particular issue is not part of this PR but rather an issue with the `play_audio_stream` action, we might need to address this in a separate PR.

cazala:
Everything seems to be working fine but I think the NFTShape component is imposible to use for a regular user. The inputs should be something like a Network dropdown, a contract address input and a token id input.

Also the styles of the dropdown for the frame style is not the standard, it should use the `Dropdown` component instead of the `SelectField`:

Screenshot 2023-12-01 at 13 52 17

cyaiox (approved):

cazala (approved):

PR #855, opened 2023-12-07
Fix VideoPlayer + NftShape scaling in Babylon + URN split into different fields

nicoecheza:


cazala:
I tried the following to test the scaling fixes but I still have issues. I created a block, and put an nft shape on one side and a video player on another side, making it roughly match the width of the box:
Screenshot 2023-12-08 at 12 18 57
Screenshot 2023-12-08 at 12 18 38

Then on the preview, I see the video player way smaller than the face of the cube:
Screenshot 2023-12-08 at 12 18 13

And the nft is blank:
Screenshot 2023-12-08 at 12 19 52

I configured it like this:
Screenshot 2023-12-08 at 12 19 15

Attepmpting to point to this NFT in ethereum: https://opensea.io/assets/ethereum/0xd35147be6401dcb20811f2104c33de8e97ed6818/1683

packages/@dcl/inspector/src/components/EntityInspector/NftShapeInspector/utils.ts
```diff
@@ -124,3 +155,26 @@ export const NFT_STYLES = [
...
+  network?: number
+  contract?: string
+  token?: string
+}
+
+export const DEFAULT_NETWORK = NETWORKS[0]
+
+export function getUrn({ network, contract, token }: UrnTokens): string {
+  const _network = NETWORKS.find(($) => $.value === network)?.label || DEFAULT_NETWORK.label
+  return `urn:decentraland:${_network}:${contract}:${token}`
```
cazala, line 169: this is missing the contract standard. I don't know if we support multiple, if so we should add a dropdown to choose one (either erc721 or erc1155) or if we only support erc721 we could simply add it to the template string here.

cazala (changes requested):

cazala (approved):

PR #856, opened 2023-12-08
feat: more items

cazala:
This PR upgrades `@dcl/asset-packs` which includes several new smart items:
- Text
- Teleport
- Social Buttons/Links
- Invisible Walls/Shapes

### DIFF
--- a/packages/@dcl/inspector/package-lock.json
+++ b/packages/@dcl/inspector/package-lock.json
@@ -8,7 +8,7 @@
       "name": "@dcl/inspector",
       "version": "0.1.0",
       "dependencies": {
-        "@dcl/asset-packs": "^1.5.0"
+        "@dcl/asset-packs": "^1.6.0"
       },
       "devDependencies": {
         "@babylonjs/core": "^6.18.0",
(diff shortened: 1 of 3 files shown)


packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
```diff
@@ -114,13 +109,16 @@ const Renderer: React.FC = () => {
...
     // TODO: review this async/await
-    dispatch(
-      importAsset({
-        content: new Map(Object.entries(fileContent)),
-        basePath: withAssetDir(destFolder),
-        assetPackageName
-      })
-    )
+    const content = new Map(Object.entries(fileContent))
+    if (content.size > 0) {
```
cazala, line 113: prevents sending the importAsset operation on assets that doen't have any contents, like the Text smart item which nly has a TextShape component and no assets.

packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
```diff
@@ -89,10 +88,6 @@ const Renderer: React.FC = () => {
     const path = Object.keys(asset.contents).find(($) => isAsset($))
     let thumbnail: Uint8Array | undefined
 
-    if (!path && !asset.components[CoreComponents.MESH_RENDERER]) {
-      throw new Error('Invalid asset format: should contain at least one gltf/glb file or a mesh renderer component')
-    }
```
cazala, line 94: This exception here is not necessary, it is from the times that the only kind of assets where gltfs, but with smart items there are some assets that have no models and they are valid items, like Text smart item.

PR #868, opened 2024-01-15
feat: SceneInspector ability to edit all the scene.json metadata

cazala:
Allow editing the full scene.json metadata from the `@dcl/inspector`
Screenshot 2024-01-24 at 20 45 04
Screenshot 2024-01-24 at 20 45 21
Screenshot 2024-01-24 at 20 45 33


cyaiox (approved):
🚀

PR #869, opened 2024-01-15
feat: Renderer hotkeys and select multiple entities

cyaiox:
This PR allows the users to use the hotkeys in the editor:
- DEL `(backspace / del)`
- COPY `(ctrl + c)`
- PASTE `(ctrl + v)`
- ZOOM IN `=` or `+`
- ZOOM OUT `-` or `_`
- RESET CAMERA `space`

It also allows the users to select multiple entities while pressing the key: `ctrl` and interact with them, updating the transform component.  

Closes: https://github.com/decentraland/sdk/issues/1064
Closes: https://github.com/decentraland/sdk/issues/1065


packages/@dcl/inspector/package.json
```diff
@@ -2,7 +2,8 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.8.0"
+    "@dcl/asset-packs": "^1.8.0",
+    "hotkeys-js": "^3.13.3"
```
cazala, line 6: oops, this should be a `devDependency`

cazala:
Another thing that I think would be necessary for the multiselect is showing the other selected entities in the Hierarchy/Tree in the left sidebar. Also I think allowing multiselecting from there should be great to have, like just pressing CTRL and selecting from the tree instead of the editor.

cazala:
This is looking great! I found one issue with the undo though, when I undo some changes I did with multiselect, it only applies the undo to the originally selected entity:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/3371a0fa-33d7-446a-8f78-e18f12f4eaf8

Another minor thing, I think the warning about the ancestor is working correctly (it does not let me select a parent entity of the originally selected one) but the copy seems wrong to me. 

In this example I have a tree with entities "Ancestor" and "Child", which is nested to "Ancestor". When I start the multiselection from "Child" and I try to add later "Ancestor", the warning reads "An ancestor of this entity has already been selected". The "this entity" I understand it as the one I just clicked and made the warning appear, which is "Ancestor". So the entity that is "already selected" is not "An ancestor of this entity" but "A child of this entity".

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/62a01ecd-13f6-4c9c-b5b7-dbf09d2cf22a

cazala (approved):

PR #872, opened 2024-01-22
feat: Add component button

cyaiox:
This PR adds a toolbar in the entity's header, allowing the users to add components to the selected entity easily or delete the entity.

image
image

Closes: https://github.com/decentraland/sdk/issues/1066


nicoecheza:
seems there is a bug when adding `VisibilityComponent` on some entities (maybe it' also happening with other components, didn't try every one)

Jan-24-2024 11-03-07

cazala (approved):

PR #876, opened 2024-01-26
feat: Improve search logic and track items found

cyaiox:
This PR improves the asset catalog search logic and tracks the items found.

### DIFF
--- a/packages/@dcl/inspector/src/components/AssetsCatalog/AssetsCatalog.tsx
+++ b/packages/@dcl/inspector/src/components/AssetsCatalog/AssetsCatalog.tsx
@@ -1,6 +1,7 @@
-import React, { useCallback, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 
 import { AssetPack } from '../../lib/logic/catalog'
+import { analytics, Event } from '../../lib/logic/analytics'
 import { useAppDispatch } from '../../redux/hooks'
 import { selectAssetsTab } from '../../redux/ui'
 import { AssetsTab } from '../../redux/ui/types'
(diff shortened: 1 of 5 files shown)


packages/@dcl/inspector/src/components/AssetsCatalog/AssetsCatalog.tsx
```diff
@@ -33,28 +34,30 @@ const AssetsCatalog: React.FC<Props> = ({ catalog }) => {
...
+        results = assets.filter((asset) => asset.name.toLowerCase().includes(search.toLowerCase()))
       }
     }
 
     return results
   }, [catalog, selectedTheme, search])
 
+  useEffect(() => {
+    if (search) {
+      analytics.track(Event.SEARCH_ITEM, {
```
cazala, line 53: should this be debounced? otherwise it will track every key stroke

cazala: we alreaady have an implementation: 

https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/inspector/src/components/ui/utils.ts#L5-L11

cyaiox: the callback `onSearch => setSearch`: https://github.com/decentraland/js-sdk-toolchain/pull/876/files#diff-44779b63b16cc772140e7e09d5226f757e76b141dc7d6f2db194eca7f5646530R96

is debounced to `500ms` in this component: https://github.com/decentraland/js-sdk-toolchain/pull/876/files#diff-5d19b541c08926fc240ec0c07f2e80323946f4497f23c2205fca7044f148e2d8R57

packages/@dcl/inspector/src/components/AssetsCatalog/AssetsCatalog.tsx
```diff
@@ -33,28 +34,30 @@ const AssetsCatalog: React.FC<Props> = ({ catalog }) => {
...
-            if (asset.name.toLowerCase().includes(search.toLowerCase())) {
-              results.push(asset)
-            }
-          }
-        }
+      const assets = selectedTheme ? selectedTheme.assets : catalog.flatMap((theme) => theme.assets)
+      results = assets.filter((asset) => asset.name.toLowerCase().startsWith(search.toLowerCase()))
+
+      if (results.length === 0) {
+        results = assets.filter((asset) => asset.name.toLowerCase().includes(search.toLowerCase()))
```
nicoecheza, line 44: maybe we can do something like:

```typescript
const searchLower = search.toLowerCase()
const { starts, includes } = assets.reduce((results: { starts: AssetPack['assets'], includes: AssetPack['assets'] }, asset) => {
  const name = asset.name.toLowerCase()
  if (name.startsWith(searchLower)) acc.starts.push(asset)
  if (name.includes(searchLower)) acc.includes.push(asset)
  return acc
})

return starts.length ? starts : includes
```

to make it a bit more clear + avoid iterating the same list twice 

(reduce/for...of, etc)

nicoecheza (changes requested):

cazala (approved):

nicoecheza (approved):

PR #877, opened 2024-01-26
feat: Add shortcuts component

cyaiox:
This PR adds a toolbar in the render to show hotkeys help.

Also add the following new hotkeys:
- `M` to activate the positioning gizmo
- `R` to activate the rotating gizmo
- `X` to activate the scaling gizmo
- `ctrl` + `D` to duplicate entities

https://github.com/decentraland/js-sdk-toolchain/assets/3170051/854d57c1-61aa-4e30-8947-f42391be8079

Closes: https://github.com/decentraland/sdk/issues/1072
Fixes: https://github.com/decentraland/sdk/issues/1076

### DIFF
--- a/packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
+++ b/packages/@dcl/inspector/src/components/Renderer/Renderer.tsx
@@ -34,11 +34,14 @@ import {
   ZOOM_IN_ALT,
   ZOOM_OUT_ALT,
   ZOOM_OUT,
-  RESET_CAMERA
+  RESET_CAMERA,
+  DUPLICATE,
+  DUPLICATE_ALT
 } from '../../hooks/useHotkey'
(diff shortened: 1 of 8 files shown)


cazala:
i noticed one issue when i `ctrl` + `D` to duplicate it also moves the camera (because WASD moves the camera). Maybe we could check on the camera bindings if the ctrl key is also pressed and prevent the movement, or stop the propagation of the event from the duplicate handler to avoid reaching the camera handler, i dunno

cyaiox:
> i noticed one issue when i `ctrl` + `D` to duplicate it also moves the camera (because WASD moves the camera). Maybe we could check on the camera bindings if the ctrl key is also pressed and prevent the movement, or stop the propagation of the event from the duplicate handler to avoid reaching the camera handler, i dunno

Hey @cazala -- thanks for the review! issue fixed 🙌

cazala (approved):

PR #884, opened 2024-02-02
fix: Update linker dapp

fzavalia:


packages/@dcl/inspector/src/components/ui/Dropdown/Option/Option.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react'
```
fzavalia, line 1: Removed it as it was unused and the linter threw a warning

cazala (approved):

PR #890, opened 2024-02-06
feat: Add Metrics Limits overlay

cyaiox:
This PR adds the scene limits metrics.

image

https://docs.decentraland.org/creator/development-guide/sdk7/scene-limitations/#scene-limitation-rules

### DIFF
--- a/packages/@dcl/inspector/src/components/Renderer/Metrics/Metrics.tsx
+++ b/packages/@dcl/inspector/src/components/Renderer/Metrics/Metrics.tsx
@@ -57,9 +57,12 @@ const Metrics = withSdk<WithSdkProps>(({ sdk }) => {
 
   useChange(
     ({ operation, component }) => {
-      if (operation === CrdtMessageType.PUT_COMPONENT && component?.componentId === sdk.components.Nodes.componentId) {
-        handleUpdateMetrics()
-        handleUpdateSceneLayout()
+      if (operation === CrdtMessageType.PUT_COMPONENT) {
+        if (component?.componentId === sdk.components.Nodes.componentId) {
+          handleUpdateMetrics()
+        } else if (component?.componentId === sdk.components.Scene.componentId) {
+          handleUpdateSceneLayout()
+        }
       }
     },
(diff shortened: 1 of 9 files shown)


nearnshaw:
@cyaiox  amazing how fast you got this working ❤️ 

I'm testing in the link to inspector from this PR

it seems that when I change the parcel layout in the component of the Scene entity, this UI doesn't react. It always says 2x2 Land, with limits of a 2x2, even if I add or remove parcels from the list.

Another thing I'm noticing is that you reach the max textures very fast. Most of the items in the asset packs reuse 3 or 4 common textures. But even if I duplicate a same item many times, each copy adds to the number of textures. I should really only count unique textures, repeated textures should only be counted once

packages/@dcl/inspector/src/components/Renderer/Metrics/Metrics.tsx
```diff
@@ -0,0 +1,112 @@
...
+      {Object.values(limitsExceeded).length > 0 && (
+        <div className="LimitExceeded">
+          <AlertIcon />
+          Too many {Object.keys(limitsExceeded)[0].toUpperCase()}
+        </div>
+      )}
+      {showMetrics && (
+        <div ref={overlayRef} className="Overlay">
+          <h2 className="Header">
+            {scene.layout.base.x}x{scene.layout.base.y} LAND
```
cazala, line 90: the "COLSxROWS" layout only makes sense for the builder, but for scenes with irregular shapes it does not work. For example you could have a scene with the shape of an "S" and that is not translatable to a COLSxROWS format.

Also `layout.base` is not being used correctly in this context. The base is the corner of the scene where to originate the x,y,z coordinate system, but is not the number of columns or rows neither.

Also for Worlds the LAND part is not relevant.

So I would say we simply remove the COLSxROWS label from the UI.

cazala (changes requested):
I see the metrics do refresh when i add new entities (ie drag n drop or duplicate) but it does not when i delete an entity.

cazala (approved):

PR #891, opened 2024-02-09
Add scaleFontSize helper

nicoecheza:
Copy of CSS `calc` function but just for `vw/vh` units

### DIFF
--- a/packages/@dcl/react-ecs/src/components/Label/types.ts
+++ b/packages/@dcl/react-ecs/src/components/Label/types.ts
@@ -36,13 +36,19 @@ export type TextAlignType =
   | 'bottom-right'
 
 /**
- * unit value type. i.e. 'vw' || 'vh'
+ * unit value type. i.e. 'w' || 'h'
  * @public
  */
-export type ScaleUnit = 'vw' | 'vh'
+export type ScaleUnit = 'w' | 'h'
 
 /**
- * unit value specified. i.e. 10 || '10vw' || '10vh'
+ * unit value specified. i.e. 10 || '10w' || '10h'
  * @public
  */
 export type FontSizeScaleUnit = `${number}${ScaleUnit}` | number
+
+/**
+ * context for applying a scale
+ * @public
+ */
+export type ScaleContext = { width: number; height: number; ratio: number }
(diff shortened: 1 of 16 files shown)


packages/@dcl/react-ecs/src/components/Label/types.ts
```diff
@@ -34,3 +34,21 @@ export type TextAlignType =
   | 'bottom-left'
   | 'bottom-center'
   | 'bottom-right'
+
+/**
+ * unit value type. i.e. 'w' || 'h'
+ * @public
+ */
+export type ScaleUnit = 'w' | 'h'
```
gonpombo8, line 42:
```suggestion
export type ScaleUnit = 'vw' | 'vh'
```

gonpombo8 (commented):
It'll greate if we can use this values directly on all the properties that accept `%` and numbers like width, height, fontSize,etc.

<UiEntity uiTransform={{ width: '100vw', fontSize: '1vw'  }} />

packages/@dcl/react-ecs/src/components/utils.ts
```diff
@@ -15,3 +22,51 @@ export function parseProps(props: EntityPropTypes) {
...
+
+/**
+ * @internal
+ */
+function getScaleAndUnit(scaleUnit: ScaleUnit): [number, ScaleUnits] {
+  if (typeof scaleUnit === 'number') {
+    return [scaleUnit, 'vw']
+  }
+
+  // the following regex would match any valid CSS unit (i.e: 7rem || 7% || 7vw)
```
gonpombo8, line 34:
```suggestion
  // the following regex would match any valid CSS unit (i.e: 7px || 7% || 7vw)
```

packages/@dcl/react-ecs/src/components/utils.ts
```diff
@@ -15,3 +22,51 @@ export function parseProps(props: EntityPropTypes) {
...
+  const [_, value, unit] = scaleUnit.match(/(-?[\d.]+)([a-z%]*)/) as [string, number, ScaleUnits]
```
gonpombo8, line 35: maybe we should add both cases here instead of a regex ?
like 
```tsx
if (scaleUnity.endsWith('vw')) return [scaleUnity, 'vw']
if (scaleUnity.endsWith('vh')) return [scaleUnity, 'vw']
return [undefined, YGUnit.UNDEFINED]
```

This way if the user pass something that's not valid then we fallback to undefined. (A lot of users use any's) https://github.com/decentraland/js-sdk-toolchain/blob/main/test/react-ecs/transform.spec.tsx#L46

cazala (approved):

PR #892, opened 2024-02-09
feat: Add mesh bounding box

cyaiox:
This PR adds the behavior to detect when an entity is partially/totally outside the parcels.

https://github.com/decentraland/js-sdk-toolchain/assets/3170051/19b5304b-1f08-4434-91d6-0be5f9dbb7ef


nearnshaw:
Hey @cyaiox 
The different problems we discussed are now fixed 👏 

But I'm encountering this. It seems that we're detecting only the real visible mesh being out of bounds, but not the bounding box.
If only an invisible bounding box is out of bounds, we aren't detecting anything.
Although that's of course better, I believe it's not consistent with how the Unity engine checks. So that would lead to cases where the editor tells you everything is fine, but the models don't get shown in prod :/

https://github.com/decentraland/js-sdk-toolchain/assets/3507907/e8f83d87-c35a-4a28-a25a-f8b138740d8d

Here you see that the tree is marked as fine, even tho the bounding box is clearly sticking out

nearnshaw:
https://github.com/decentraland/js-sdk-toolchain/assets/3507907/2588f10a-cbef-483b-9839-bb914b820f04

Here's a clearer case, with a round plate that's near the ground

nearnshaw:
Another problem:
Are we checking out of bounds when we change values numerically via the Transform UI?

In this case, I'm trying to get the floor to match exactly w the 1 parcel scene, that is almost impossible moving the item visually. If I adjust the numbers to match this, it remains red

https://github.com/decentraland/js-sdk-toolchain/assets/3507907/0678e5d9-e796-4b1a-8d33-de02a58aba32


Similarly, if it's well placed visually and then I change the numbers in the UI, it doesn't get marked as out of bounds

https://github.com/decentraland/js-sdk-toolchain/assets/3507907/78906d48-ba24-4c6f-914b-f9929e78a1a9

cazala:
Hey @cyaiox this is looking great! I found some issues though when parenting entities, it seems that the status of the entity (either out of bounds or back in bounds) does not get updated when moving the parent entity:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/704fc43a-37eb-480a-8e78-c83a84311c3f


And also another minor issue but the entity does not turn red if you already drop it outside of the scene bounds until you move it:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/58683481-7b1f-4764-a88e-21c56e62d907

cyaiox:
Thanks @nearnshaw and @cazala for your review! 

with this latest push the issues will be fixed! 🤞

nearnshaw:
@cyaiox   perfect 👏 
I just checked and all of the issues we found are now fixed!  great job

cazala (approved):

cazala (approved):

PR #904, opened 2024-02-21
fix: BoundingBox mesh detection

cyaiox:
This PR addresses two issues.

1) The incorrect detection of the mesh bounding box when loading assets from the catalog
2) Fixes the misplaced bounding box position


cazala (approved):
awesome 🙌

PR #906, opened 2024-02-27
fix: Avoid showing the bounding box when the scene is still loading assets

cyaiox:
This PR adds validation to avoid showing meshes' bounding boxes when the scene is still loading out of the layout.

It also adds a new `SceneMetrics` RPC and store to consume the scene `metrics | limits | entitiesOutOfBoundaries` outside the inspector, like from the builder.


packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
```diff
@@ -244,22 +248,22 @@ async function validateEntityIsOutsideLayout(entity: EcsEntity) {
 
 function updateMeshBoundingBoxVisibility(entity: EcsEntity, mesh: BABYLON.AbstractMesh) {
   const scene = mesh.getScene()
+  if (scene.isLoading) return
```
cyaiox, line 251: Condition added to avoid showing bounding boxes while the scene is still loading assets

cazala (approved):

PR #909, opened 2024-03-01
auto install & import asset-packs for editor scenes

gonpombo8:
With this PR is not mandatory to import and install the @dcl/asset-pack lib for the editor scenes.
The sdk-commands does all this at build-time in the background using the asset-pack version that the inspector used to build the scene.
So this way we don't have mismatched of @dcl/asset-packs version. (The inspector already has the asset-packs lib, and the scene could have another version that could lead into mysterious bugs )

## Before (game.ts scene): 
```ts
import { engine } from '@dcl/sdk/ecs'
import { initAssetPacks } from '@dcl/asset-packs/dist/scene-entrypoint'

initAssetPacks(engine)

function main() {}
```

## Now
```ts
import { engine } from '@dcl/sdk/ecs'
function main() {}
```

### DIFF
--- a/packages/@dcl/inspector/package.json
+++ b/packages/@dcl/inspector/package.json
@@ -2,7 +2,7 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.12.2",
+    "@dcl/asset-packs": "^1.12.3-20240301140405.commit-8aef784",
     "ts-deepmerge": "^7.0.0"
   },
   "devDependencies": {
--- a/packages/@dcl/inspector/package.json
+++ b/packages/@dcl/inspector/package.json
@@ -2,7 +2,7 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.12.3-20240301140405.commit-8aef784",
+    "@dcl/asset-packs": "^1.13.0",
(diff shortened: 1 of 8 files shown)


packages/@dcl/inspector/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@dcl/inspector",
   "version": "0.1.0",
   "dependencies": {
-    "@dcl/asset-packs": "^1.12.2",
+    "@dcl/asset-packs": "^1.12.3-20240301140405.commit-8aef784",
```
gonpombo8, line 5: TODO: make a asset-packs release and use that stable versioin

packages/@dcl/sdk-commands/src/logic/bundle.ts
```diff
@@ -59,6 +60,16 @@ import * as sdk from '@dcl/sdk'
 import { compositeProvider } from '@dcl/sdk/composite-provider'
 import { compositeFromLoader } from '~sdk/all-composites'
 
+${
+  isEditorScene &&
+  `
+import { initAssetPacks } from '@dcl/asset-packs/dist/scene-entrypoint'
+initAssetPacks(engine)
+
+// TODO: do we need to do this on runtime ?
```
gonpombo8, line 69: move this to another issue

nicoecheza (approved):

cazala (approved):

PR #930, opened 2024-04-12
fix: hide GLTFs on entities nested under the Camera or Player

cazala:
Currently, when an entity is nested under the Player or Camera, it is hidden from the scene renderer (Babylon). That works fine when reparenting, but when the scene is loaded for the first time (like after refreshing the editor), all the GLTFContainers are loaded into the scene, no matter where the entity is nested under.

This PR changes the `loadGltf` function to check the root of the entity tree branch, and if it's not the RootEntity, it hides the model. We still load the model, otherwise if this entity is later nested under the RootEntity, the `entity.setVisibility(true)` whould not work, because the model is not loaded.


nicoecheza (approved):
🚀

PR #933, opened 2024-04-23
fix: prevent race condition between loading Nodes component and GLTFs

cazala:
The previous implementation was assuming that when the root could not be determined (ie. it was `null`) the model should be hidden. 

This was because when refreshing a scene, the `loadGLTF` method is called from two places: the `Renderer.tsx` and the `putGltfContainerComponent`, and the calls from the `Renderer.tsx` where prior to the composite being instantiated, thus the root could not be determined due to the Nodes component not being loaded into the engine, and the calls from the `putGltfContainerComponent` happend after the Nodes component was already loaded.

But that assumption was wrong, and sometimes the `putGltfContainerComponent` is called also before the Nodes component is loaded, so the root is never able to be determined (this depends on the order of the component keys in the composite file).

So this PR changes the `loadGLTF` to instead of hidding the models when the root can't be determined, it waits 100ms and tries again, until the Nodes are loaded and the root can be determined.

### DIFF
--- a/packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
+++ b/packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
@@ -78,8 +78,9 @@ export async function loadGltf(entity: EcsEntity, value: string) {
   }
 
   let root = entity.getRoot()
+  let attempts = 0
 
-  while (root === null) {
+  while (root === null && attempts++ < 10) {
     // waiting for nodes to be loaded...
     await new Promise((resolve) => setTimeout(resolve, 100))
     root = entity.getRoot()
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/src/lib/babylon/decentraland/sdkComponents/gltf-container.ts
```diff
@@ -77,23 +77,30 @@ export function loadGltf(entity: EcsEntity, value: string) {
     sceneContext = entity.context
   }
 
-  const root = entity.getRoot()
+  let root = entity.getRoot()
+  let attempts = 0
 
-  const shouldHide = root === PLAYER || root === CAMERA || root === null
+  while (root === null && attempts++ < 10) {
```
cazala, line 83: there could be a scenario where there are GLTFs and there a no Nodes (like if the user deliverately edits the `main.composite` and removes the Nodes component data), and in that case this would loop forever.

To avoid that, we kill the loop after 1s. What would happen if at such point the Nodes component is not loaded, the root will be `null`, and if the models were nested under the Camera or Player entities, they would NOT be hidden. It's a very specific edge case anyway.

nicoecheza (approved):

PR #938, opened 2024-04-26
fix: remove magic cube

cazala:
Remove "Magic Cube" from autogenerated `main.composite` on workspaces that don't have a composite when using the `@dcl/inspector`

This caused an issue that if you open a non-Editor project with the Editor, and then open the Visual Editor, the inspector would autogenerate a composite with the Magic Cube for that project.


AlejandroAlvarezMelucciDCL:
image
I'm not sure what that is, looks like it's not related to this PR, but in the Files changed list I see that. Maybe while you're working on this PR can you fix it?

cazala:
@AlejandroAlvarezMelucciDCL yep that one slipped thru! good catch :) fixed ✅

gonpombo8 (approved):

PR #945, opened 2024-05-09
Add new Scene Layout

nicoecheza:
Add new scene parcels editor

[Figma](https://www.figma.com/file/sXjsoASo7nc24UY6OZnR1d/%F0%9F%A7%B0-Editor-bug-fixes-%7C-Builder-%7C-Cycle-6?type=design&node-id=1502-21056&mode=design&t=JnomtHgeIzQ5w02W-4) (ask ezio for permissions)

Screenshot 2024-05-14 at 12 36 40 PM

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/SceneInspector/Layout/Layout.css
+++ b/packages/@dcl/inspector/src/components/EntityInspector/SceneInspector/Layout/Layout.css
@@ -1,20 +1,19 @@
-.SceneLayout {
-}
-
-.SceneLayout h2 {
-  margin-bottom: 10px;
-}
-
-.SceneLayout span {
-  text-align: center;
-  color: var(--base-10);
-  margin-bottom: 10px;
-}
-
 .SceneLayout .display {
   display: flex;
   align-items: center;
   flex-direction: column;
+  margin-bottom: 20px;
+}
+
+.SceneLayout .display h2 {
+  margin-bottom: 10px;
+}
+
+.SceneLayout .display span,
+.SceneLayout .Block .limitations span {
+  text-align: center;
+  color: var(--base-10);
+  margin-bottom: 10px;
 }
 
 .SceneLayout .Grid {
@@ -22,4 +21,22 @@
   flex-direction: column;
   width: 100%;
   padding: 0 10px 0 10px;
+}
+
+.SceneLayout .Block {
+  display: flex;
+  width: 100%;
+}
+
+.SceneLayout .Block button {
+  width: 100%;
+}
+
+.SceneLayout .Block .limitations {
+  justify-items: center;
+  align-items: center;
+}
+
+.SceneLayout .Block .limitations span {
+  margin: 0 3px 0 0;
 }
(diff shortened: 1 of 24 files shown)


packages/@dcl/inspector/src/components/EntityInspector/SceneInspector/Layout/Layout.tsx
```diff
@@ -0,0 +1,194 @@
...
+              Set coordinates (advanced)
+            </Button>
+            <Button type="blue" size="big" onClick={handleApplyClick} disabled={!!gridError}>
+              Apply layout
+            </Button>
+          </Block>
+          <Block className="limitations">
+            <span>About scene limitations</span>
+            <InfoTooltip
+              text=""
```
cazala, line 183: Is it okay that this tooltip has no text?

packages/@dcl/inspector/src/components/EntityInspector/SceneInspector/Layout/Layout.css
```diff
@@ -0,0 +1,67 @@
...
+.SceneLayout .Block .limitations span {
+  text-align: center;
+  color: var(--base-10);
+  margin-bottom: 10px;
+}
+
+.SceneLayout .Grid {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
```
cazala, line 28: Just a heads up, when the amount of parcels is wider than the width of the Grid (which is the width of the sidebar) some parecels are not reachable, even by using the scroll.

This seems like an edge case tho, and the sidebar can be widened to make them fit, just wanted to give a heads up.

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/5a5858ed-228e-4ace-bb84-69364ad7b948

cazala (changes requested):
Looking good! The basic view works as expected, but I'm having issues with the advanced view.

After clicking on the Confirm button, if I try to make further changes they are not applied, and the view seems to stop working until I go back to the basic view. But even after that, the preview of the selected parcel in the basic view stays out of sync with the actual layout in the renderer.

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/7f08ff2a-2d93-4c7f-83a1-83b06fcd27af

I also noted that the value of the Custom Parcels input does not get updated as I select or unselect parcels in the UI above it.

Another less important issue but I think is worth mentioning is that it seems the Custom Parcels input allows me to enter parcels that are already selected, i dunno if that is okay or if they should be removed after I clock on Confirm:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/053784c6-5d81-4da2-9215-c61a5a006f85

Another issue I find is that if I enter a ver distant parcel the whole editor just breaks with a `Maximum call stack size exceeded`

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/ded856c6-e9dc-4235-a7e3-cfaede0940d6

Similarly, if I try to delete all the parcels in the input, the UI will freeze when I try to delete the last one, as if the app was stuck in an infinite loop:

https://github.com/decentraland/js-sdk-toolchain/assets/2781777/75dc50a2-942a-482a-b12a-0d7a0bae752c

cazala (approved):

PR #947, opened 2024-05-13
feat: add more avatar attach points

AlejandroAlvarezMelucciDCL:
Updated protocol to enable new avatar attach points.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@actions/core": "^1.10.0",
-        "@dcl/protocol": "1.0.0-8758013256.commit-44aab53",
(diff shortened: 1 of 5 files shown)


leanmendoza (approved):
LGTM!
Note: you're updating the package.json version from `2` to `3`

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "js-sdk-toolchain",
   "version": "7.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
gonpombo8, line 4: which npm version are you using ?
there are a lot of changes for only updating a package version ? 
we need to verifiy that this update doesn't break anything, maybe its better to use npm 8 for now

leanmendoza (approved):

PR #950, opened 2024-05-15
DevExp improvements

gonpombo8:
- Add meshCollider check for entities with PointerEvent and print a warn if it cant be clickeable
- Use width as a default value for UiEntities (useful for UiText)
- Fix React.JSX type (Closes https://github.com/decentraland/sdk/issues/778)
- Warning when using syncEntity with components that can't be sync (Closes https://github.com/decentraland/sdk/issues/1105)
- Exclude invalid componentIds from syncEntity


gonpombo8:
Closes https://github.com/decentraland/sdk/issues/778

PR #952, opened 2024-05-16
feat: hide metrics

cazala:
This PR allows hiding the `<Metrics />` panel using the `UIClient` via RPC.


packages/@dcl/inspector/src/lib/rpc/ui/ui.spec.ts
```diff
@@ -70,4 +71,11 @@ describe('UiRPC', () => {
       )
     })
   })
+
+  describe('When using the toggleGroundGrid method of the client', () => {
+    it('should send a ui/toggleGroundGrid action in the server', async () => {
+      await expect(client.toggleGroundGrid(false)).resolves.not.toThrow()
+      expect(store.dispatch).toHaveBeenCalledWith(toggleGroundGrid({ enabled: false }))
+    })
+  })
```
cazala, line 80: This test is not related to this PR, but we were missing a test case for the `toggleGroundGrid` method from a previous PR so I added it.

PR #954, opened 2024-05-24
Some editor fixes

nicoecheza:
* Add unique names for new Nodes

Screenshot 2024-05-24 at 2 26 06 PM

* Add free movement gizmo to toolbar

Screenshot 2024-05-24 at 2 26 11 PM


cazala (approved):
Looking good! The unique name feature is working fine for assets that I drag and drop, but I noticed that if I duplicate an item, the name same name is used. I dunno if that is intended or if the name should also have the `_X` suffix on the duplicated entity.

PR #955, opened 2024-05-27
Fix/react ecs types

gonpombo8:


AlejandroAlvarezMelucciDCL (approved):
I think it looks nicer if null and undefined are the last elements, but it's fine as is

leanmendoza (approved):

PR #961, opened 2024-06-07
fix: Refactor get bounding box info and outside layout material generation

cyaiox:
This PR refactors the bounding box info and the outside layout material generation to avoid cases when nested entities behave unexpectedly.

Closes: https://github.com/decentraland/sdk/issues/1119
Closes: https://github.com/decentraland/sdk/issues/1101

### DIFF
--- a/packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
+++ b/packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
@@ -97,6 +97,66 @@ export class EcsEntity extends BABYLON.TransformNode {
     return new BABYLON.BoundingInfo(min, max)
   }
 
+  getTransformedVertices(mesh: BABYLON.AbstractMesh) {
+    const positions = mesh.getVerticesData(BABYLON.VertexBuffer.PositionKind)!
+
+    const worldVertices = []
+    const scaling = mesh.scaling
+    const rotationQuaternion = mesh.rotationQuaternion || BABYLON.Quaternion.Identity()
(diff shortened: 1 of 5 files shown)


packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
```diff
@@ -97,6 +97,66 @@ export class EcsEntity extends BABYLON.TransformNode {
...
+  }
+
+  getGroupMeshesBoundingBox() {
+    const children = this.getChildMeshes(false)
+    if (children.length === 0) {
+      return null // No children to calculate the bounding box
+    }
+
+    let groupMin = new BABYLON.Vector3(Number.MAX_VALUE, Number.MAX_VALUE, Number.MAX_VALUE)
+    let groupMax = new BABYLON.Vector3(-Number.MAX_VALUE, -Number.MAX_VALUE, -Number.MAX_VALUE)
```
nicoecheza, line 147: are we sure we want to use `Number.MAX_VALUE` instead of `Number.MAX_SAFE_INTEGER` & `Number.MIN_SAFE_INTEGER`?

nicoecheza (changes requested):

packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
cyaiox: good catch -- updating it!

nicoecheza (approved):

PR #974, opened 2024-07-02
[WIP] Map pin

gonpombo8:


test/ecs/components/MapPin.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  it('should serialize/deserialize move Tween', () => {
+    const newEngine = Engine()
+    const MapPin = components.MapPin(newEngine)
+
+    testComponentSerialization(MapPin, {
+      title: 'Title',
+      description: 'Description',
+      iconSize: 0.5,
+      position: { x: 1, y: 2 },
+      texture: {
```
gonpombo8, line 14: are we sure about this ? Do we need all this fields or just a string to a path ?
@davidejensen

packages/@dcl/playground-assets/etc/playground-assets.api.md
```diff
@@ -598,6 +598,7 @@ export const componentDefinitionByName: {
     "core::EngineInfo": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBEngineInfo>>;
     "core::GltfContainer": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBGltfContainer>>;
     "core::GltfContainerLoadingState": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBGltfContainerLoadingState>>;
+    "core::MapPin": LwwComponentGetter<LastWriteWinElementSetComponentDefinition<PBMapPin>>;
```
gonpombo8, line 601: If this is a LWW, in which Entity are we going to store this information ? 
engine.RootEntity ? 
You can ONLY have ONE MapPin ?

PR #999, opened 2024-09-16
Fix corruption of Transform data (NetworkParent + Tween)

Maironire:
This PR fixes **Transform** misaligned data reading (all data is shifted by 4 bytes) when entity has **NetworkParent**.

This bug appears as incorrect data produced by `Transform.get`, and causes more impact when one is using `Transform.getMutable` (or recurrent `parentEntity` as well, because it implies the former internally) after network parent was initially assigned - this way incorrect data is forwarded back to engine, and on the next step it will be shifted again.

It occurs only when engine sends **Transform** updates to a scene (i.e. when using **Tween**, but perhaps other cases exist), and doesn't show up if only scene is updating the data.

### DIFF
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -161,14 +161,6 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
 
         /* istanbul ignore else */
         if (component) {
-          if (
-            msg.type === CrdtMessageType.PUT_COMPONENT &&
-            component.componentId === Transform.componentId &&
-            NetworkEntity.has(entityId) &&
-            NetworkParent.has(entityId)
-          ) {
-            msg.data = networkUtils.fixTransformParent(msg)
-          }
           const [conflictMessage, value] = component.updateFromCrdt({ ...msg, entityId })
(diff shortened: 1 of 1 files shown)


gonpombo8 (commented):
You cant delete this line since it will break the syncronization of that entity (and its parent).

packages/@dcl/ecs/src/systems/crdt/index.ts
```diff
@@ -161,14 +161,6 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
 
         /* istanbul ignore else */
         if (component) {
-          if (
```
gonpombo8, line 164: You cant delete this line since it will break the syncronization of that entity (and its parent).

gonpombo8 (changes requested):

packages/@dcl/ecs/src/systems/crdt/index.ts
Maironire: Okay, I understood your point that we need it zero parent to avoid conflicts while syncing with other players.
Moreover, it was not the root cause of the bug, so this is not actually a 'fix' anymore, you've got a better one.
Nevertheless, I'm up to propose a more consistent way of doing this, I'll update this PR as soon as your fix got merged.

PR #1013, opened 2024-09-30
chore: 'about' response local scene parcels

pravusjif:
Added local scenes parcels info in 'about' reponse message, so that the Explorer can use that to generate the dynamic landscape for local scene development.

--------------------------------------

Related PRS:

* https://github.com/decentraland/unity-explorer/pull/2229

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@actions/core": "^1.10.0",
-        "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-11054456137.commit-a579c92.tgz",
+        "@dcl/protocol": "1.0.0-10955038136.commit-7fe9554",
         "@dcl/quickjs-emscripten": "^0.21.0-3680274614.commit-1808aa1",
         "@dcl/ts-proto": "1.153.0",
         "@types/fs-extra": "^9.0.12",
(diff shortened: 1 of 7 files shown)


gonpombo8 (approved):

leanmendoza (changes requested):
If the goal is set the corner for the landscape, I suggest to use the `map` field to specify the size (and compute the landscape inf function of it), [recently introduced in the protocol ](https://github.com/decentraland/protocol/pull/206).
See the response of the `genesis-city` one (I think it's still hardcoded in EA, but it's ready to be used!) https://peer.decentraland.org/about

pravusjif:
> If the goal is to set the boundaries for the landscape, I suggest using the `map` field to specify the size (and compute the landscape as a function of it), [recently introduced in the protocol ](https://github.com/decentraland/protocol/pull/206). See the response of the `genesis-city` one (I think it's still hardcoded in EA, but it's ready to be used!) https://peer.decentraland.org/about

Uhmm in the Explorer side I'm needing the local scene occupied parcels before actually loading any scene, for local scene development, so that based on those parcels the dynamic landscape can get generated during initialization, the same that the Explorer Alpha does for worlds but wasn't able to do it for local scenes because worlds have their collection of ipfs urn scenes.

leanmendoza:
> > If the goal is to set the boundaries for the landscape, I suggest using the `map` field to specify the size (and compute the landscape as a function of it), [recently introduced in the protocol ](https://github.com/decentraland/protocol/pull/206). See the response of the `genesis-city` one (I think it's still hardcoded in EA, but it's ready to be used!) https://peer.decentraland.org/about
> 
> Uhmm in the Explorer side I'm needing the local scene occupied parcels before actually loading any scene, for local scene development, so that based on those parcels the dynamic landscape can get generated during initialization, the same that the Explorer Alpha does for worlds but wasn't able to do it for local scenes because worlds have their collection of ipfs urn scenes in the realm.
> 
> Maybe in the future we can use your approach (would need a bigger refactor of all the landscape terrain generation logic in Explorer Alpha) but right now all the dynamic landscape complexity only needs a list of parcels (or urn scenes, as in the case of worlds).

I understand. Let's go ahead and merge this temporarily to avoid blocking anything (if there's urgency), but let's not commit to keeping it in the about section permanently just because of this (at least we agreed it's completely necessary).

leanmendoza (approved):

PR #1014, opened 2024-09-30
emit messages to specific address

gonpombo8:


packages/@dcl/sdk/src/network/message-bus-sync.ts
```diff
@@ -45,8 +62,9 @@ export function addSyncTransport(
           console.log(...Array.from(serializeCrdtMessages('[NetworkMessage sent]:', message, engine)))
         binaryMessageBus.emit(CommsMessage.CRDT, message)
       }
-      const messages = getMessagesToSend()
-      const response = await sendBinary({ data: messages })
+      const [broadcastMessages, messagesToAddress] = getMessagesToSend()
+
+      const response = await sendBinary({ data: broadcastMessages, messagesToAddress })
```
gonpombo8, line 67: maybe `peerMessages` its a better name ?

PR #1024, opened 2024-10-30
feat: pointer event toggle highlight

pravusjif:
* Implements support for new `PointerEvents` `showHighlight` property
* Fixed bug existent bug for the case of using `pointerEventsSystem.onPointerDown()` without any of the pointer event optional properties (those entities weren't clickable before this PR)

Related PRs:
- https://github.com/decentraland/protocol/pull/224
- https://github.com/decentraland/unity-explorer/pull/2651

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "bugs": "https://github.com/decentraland/js-sdk-toolchain/issues",
   "dependencies": {
     "@actions/core": "^1.10.0",
-    "@dcl/protocol": "1.0.0-11406954347.commit-ba19c4f",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-11593429245.commit-6dd6863.tgz",
     "@dcl/quickjs-emscripten": "^0.21.0-3680274614.commit-1808aa1",
     "@dcl/ts-proto": "1.153.0",
     "@types/fs-extra": "^9.0.12",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "bugs": "https://github.com/decentraland/js-sdk-toolchain/issues",
   "dependencies": {
     "@actions/core": "^1.10.0",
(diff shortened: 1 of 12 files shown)


pravusjif:
I'll update the protocol package to point to the latest published one after the protocol PR is merged.

package.json
```diff
@@ -6,7 +6,7 @@
   "bugs": "https://github.com/decentraland/js-sdk-toolchain/issues",
   "dependencies": {
     "@actions/core": "^1.10.0",
-    "@dcl/protocol": "1.0.0-11406954347.commit-ba19c4f",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-11593429245.commit-6dd6863.tgz",
```
gonpombo8, line 9: update

packages/@dcl/ecs/src/systems/events.ts
```diff
@@ -110,14 +111,15 @@ export function createPointerEventsSystem(engine: IEngine, inputSystem: IInputSy
   }
 
   function setPointerEvent(entity: Entity, type: PointerEventType, opts: EventSystemOptions) {
-    if (opts.hoverText || opts.showFeedback) {
+    if (opts.hoverText || opts.showFeedback || opts.showHighlight) {
```
gonpombo8, line 114: is this necessary ? not sure why we have this so asking just in case

pravusjif: great catch! by removing that check we solve an existent bug (when using `pointerEventsSystem.onPointerDown()` without any of the pointer event optional properties)

package.json
pravusjif: Done!

packages/@dcl/ecs/src/systems/events.ts
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1024/commits/0abea71304fb5c50ecd1ebbfc4af91a83673cd83

gonpombo8 (approved):

PR #1025, opened 2024-10-30
chore: update ecs-math package

AlejandroAlvarezMelucciDCL:
Updated ecs-math package


packages/@dcl/inspector/package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@dcl/inspector",
   "version": "0.1.0",
-  "lockfileVersion": 3,
```
gonpombo8, line 4: i think this can cause uknown issues.
Which version of npm are you using ? 
I would checkout this changes, and do this update again from skratch using npm v10 maybe ? (notsure) (also clean your node_modules just in case)

AlejandroAlvarezMelucciDCL: 8.19.4, I had to install that one a long ago to avoid having that issue in another PR... I'll see what can I do

gonpombo8 (approved):

PR #1029, opened 2024-11-06
Add gizmo centroid

nicoecheza:


packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -152,7 +152,7 @@ export function Tree<T>() {
       const quitInsertMode = () => setInsertMode(false)
 
       const handleSelect = (event: React.MouseEvent) => {
-        if (event.type === ClickType.CONTEXT_MENU && event.ctrlKey) {
+        if (event.type === ClickType.CONTEXT_MENU && (event.ctrlKey || event.shiftKey)) {
```
cazala, line 155: I tried using the `shift` key to multi select by clicking on the tree or elements on the renderer, and it doesn't seem to be working 🤔 `ctrl` is working fine.

nicoecheza: shift won't work since it's usual function is to select a range of items at once. Since we that's not required, I didn't want to have two different ways to do the same since it's confusing with the "normal" behaviour for selecting ranges

packages/@dcl/inspector/src/lib/babylon/decentraland/gizmo-manager.ts
```diff
@@ -309,24 +372,32 @@ export function createGizmoManager(context: SceneContext) {
...
+      if (areMultipleEntitiesSelected()) {
+        return repositionGizmoOnCentroid()
+      } else {
+        gizmoManager.attachToNode(entity)
+        lastEntity = entity
+        // fix gizmo rotation/position if necessary
+        const transform = getTransform()
+        fixRotationGizmoAlignment(transform)
+        fixPositionGizmoAlignment(transform)
+        events.emit('change')
```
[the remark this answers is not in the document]
nicoecheza: good catch

cazala (approved):

PR #1032, opened 2024-11-09
wip

leanmendoza:
this is a PoC to have
```typescript
// getter
console.log(Transform.get(entity).globalPosition)
// and setter
Transform.getMutable(entity).globalPosition = Vector3.create(8, 1, 8)
```
also this research is to have a test for performance and if it's a viable option to implementing further in more components (like the ones with oneOf and make their interactions seamless)


packages/@dcl/ecs/src/components/extended/transform-utils/globalPosition.ts
```diff
@@ -0,0 +1,114 @@
...
+export function getGlobalPositionHelper(transformDef: LastWriteWinElementSetComponentDefinition<TransformType>) {
+  const helperArray: Entity[] = []
+  function getGlobalTransform(transform: TransformType): TransformType {
+    const position = { ...transform.position }
+    const scale = { ...transform.scale }
+    const rotation = { ...transform.rotation }
+
+    helperArray.length = 0
+    let currentTransform = null
+    if (transform.parent !== undefined) {
```
gonpombo8, line 18: This is not going to work for network entities.

We should include some logic here for that, like if the entity is marked as sync, look for the network parentEntity instead of the transform.parent

leanmendoza: We can take two path for this, as it doesn't work with AvatarAttach nor Billboard neither:
a. We add the disclaimer comment to get noticed when this accessor is approachable or not.
b. We proper implement all the possibilities and get this feature reliable.


In case of a) we can move forward with the current implementation and add the comments

In case of b), my question is, is there any mechanism to get the globalPositon so far? Because what I saw in Utils library, this is not implemented. The current `getWorldPosition` and `getWorldRotation` doesn't take into consideration the `scale`.

PR #1035, opened 2024-11-12
chore: helper functions for InputModifier

AlejandroAlvarezMelucciDCL:
wip

### DIFF
--- a/packages/@dcl/ecs/src/components/extended/InputModifier.ts
+++ b/packages/@dcl/ecs/src/components/extended/InputModifier.ts
@@ -1,0 +1,35 @@
+import { ReadOnlyLastWriteWinElementSetComponentDefinition, IEngine } from "../../engine"
+import { InputModifier } from "../generated/index.gen"
+import { PBInputModifier_StandardInput, PBInputModifier } from "../generated/pb/decentraland/sdk/components/input_modifier.gen"
+
+export interface InputModifierHelper {
+  /**
+   * @returns a input modifier mode
+   */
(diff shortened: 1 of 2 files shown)


packages/@dcl/ecs/src/components/extended/InputModifier.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @public
+ */
+export interface InputModifierComponentDefinitionExtended
+  extends LastWriteWinElementSetComponentDefinition<PBInputModifier> {
+  /**
+   * InputModifier helper with constructor
+   */
+  InputModifier: InputModifierHelper
```
gonpombo8, line 23: I think this should be named in a different way because this way it generates the following code:
```ts
      mode: InputModifier.InputModifier.Standard({ disableAll: true })
```
and its redundant. Maybe it should be `InputModifier.Mode.Standard...`
```suggestion
  Mode: InputModifierModeHelper
```

packages/@dcl/ecs/src/components/extended/InputModifier.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Create or replace the component InputModifier in the entity specified
+   * @param entity - the entity to link the component
+   * @param inputModifier - the data for this input modifier
+   */
+  addStandardModifier: (entity: Entity, inputModifier: PBInputModifier_StandardInput) => void
```
gonpombo8, line 30: do we need this helper ? It seems that the Mode.Standard could do the trick and avoid generating extra methods

gonpombo8 (approved):

PR #1042, opened 2024-11-29
refactor: composite

cazala:
This PR refactor the `addAsset` operation to work with the new `@dcl/asset-packs` asset format, which uses composites instead of the custom format.

It also adds support for multiple entities per item, since the composite supports multiple entities. When multiple entities are present, a main entity will be created (this one will have the Transform in the position the user selected) and the item's entities will be added as children of that main entity. When a single entity is used and it's id is 0, then those components will be added to the main entities (to avoid having the main entity with a single child when only 1 entity is needed. which is the case of the current smart items).

Finally, it refactors the ProjectAssets and BuilderAssets into LocalAssets and CatalogAssets, which match better the current nomenclatures.

### DIFF
--- a/packages/@dcl/inspector/src/lib/sdk/drag-drop.ts
+++ b/packages/@dcl/inspector/src/lib/sdk/drag-drop.ts
@@ -14,7 +14,7 @@ export type CatalogAssetDrop = Drop<Asset>
 export type IDrop = LocalAssetDrop | CatalogAssetDrop
 
 export enum DropTypesEnum {
-  ProjectAsset = 'local-asset',
+  LocalAsset = 'local-asset',
   CatalogAsset = 'catalog-asset'
 }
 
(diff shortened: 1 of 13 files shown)


packages/@dcl/inspector/src/lib/sdk/drag-drop.ts
```diff
@@ -8,14 +8,14 @@ interface Drop<T, K = object> {
...
+export type LocalAssetDrop = Drop<string, { tree: Map<string, TreeNode> }>
+export type CatalogAssetDrop = Drop<Asset>
 
-export type IDrop = ProjectAssetDrop | BuilderAsset
+export type IDrop = LocalAssetDrop | CatalogAssetDrop
 
 export enum DropTypesEnum {
-  ProjectAsset = 'project-asset',
-  BuilderAsset = 'builder-asset'
+  ProjectAsset = 'local-asset',
```
cyaiox, line 17: should be here?
```suggestion
  LocalAsset = 'local-asset',
```

cazala: Yes! Updated ✅

PR #1051, opened 2025-01-16
Easy import asset flow

nicoecheza:
How to test?

* Drag & drop files to the asset explorer panel
* Enjoy

### DIFF
--- a/packages/@dcl/inspector/src/components/AssetPreview/AssetPreview.tsx
+++ b/packages/@dcl/inspector/src/components/AssetPreview/AssetPreview.tsx
@@ -1,10 +1,12 @@
-import { useCallback, useMemo } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import { PreviewCamera, PreviewProjection } from '@dcl/schemas'
+import cx from 'classnames'
 import { WearablePreview } from 'decentraland-ui'
 import { AiFillSound } from 'react-icons/ai'
 import { IoVideocamOutline } from 'react-icons/io5'
 import { FaFile } from 'react-icons/fa'
 
+import { Loading } from '../Loading'
 import { toWearableWithBlobs } from './utils'
 import { Props } from './types'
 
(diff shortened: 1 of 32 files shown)


nicoecheza:
QA check => https://github.com/decentraland/creator-hub/pull/392#issuecomment-2616484552

packages/@dcl/inspector/src/components/AssetPreview/AssetPreview.tsx
```diff
@@ -14,36 +17,58 @@ const WIDTH = 300
...
-      camera={PreviewCamera.STATIC}
-      onLoad={handleLoad}
-    />
+    <>
+      <div className={cx('GltfPreview', { hidden: loading })}>
+        <WearablePreview
+          id={value.name}
+          blob={toWearableWithBlobs(value, resources)}
+          disableAutoRotate
+          background="#3c3c3c"
```
cazala, line 64: mm i really dont think we should use a solid background for the image. We should have a transparent background png and set the background with css. Otherwise any change to the background or style later will look bad with a hardcoded background.

cazala:
The bottom-right buttons in the renderer viewport have a higher z-index than the import modal: 

https://github.com/user-attachments/assets/6f6f8151-4cca-43ed-bcc7-2554b4551a1f

cazala:
Switching between tabs makes the tab menu to jump a few pixels up and down:

https://github.com/user-attachments/assets/b7013fb3-492e-42c5-9ee0-a5803f01cab5

cazala:
The create custom item flow seems a bit broken:

1. The thumbnail has a double loader
2. The generated thumbnail is stretched
3. The final custom item has the wrong background due to the image not being transparent, causing a weird padding/frame, and the on-hover highlight not being properly applied.

https://github.com/user-attachments/assets/8604f2ed-d83a-4f3a-a37f-c9ace7876223

PR #1061, opened 2025-02-05
fix Animator component initialization

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/266

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
+++ b/packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
@@ -21,24 +21,33 @@ export function isValidSpeed(speed: string | undefined): boolean {
   return !isNaN(parseFloat(value)) && parseFloat(value) >= 0 && parseFloat(value) <= 200
 }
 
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
```diff
@@ -21,24 +21,33 @@ export function isValidSpeed(speed: string | undefined): boolean {
   return !isNaN(parseFloat(value)) && parseFloat(value) >= 0 && parseFloat(value) <= 200
 }
 
+export function mapAnimationGroupsToStates(animations: AnimationGroup[]): PBAnimationState[] {
+  return animations.map(($) => ({
+    clip: $.name,
+    playing: !!$.isPlaying,
+    weight: $.weight ?? 1,
```
cyaiox, line 28: wdyt about initializing this property as undefined and letting the user define the value in the UI?

nicoecheza: I remember talking with nico so about always having some defaults on the components values so user's can have their preview updated as soon as they add a component and have feedback. I think we should leave it as 1 (same value as before)

cyaiox (approved):

PR #1063, opened 2025-02-06
feat: Make createReactBasedUiSystem public

cyaiox:
This PR updates the access modifier for the class `createReactBasedUiSystem`, allowing us to instantiate the system in another library, such as `@dcl/asset-packs`.

It also passes the playerHelper object as a param to the initializer method: `initAssetPack`, allowing us to access the player data from the lib `@dcl/asset-packs`.


gonpombo8 (approved):
the only weird thing that i see here is that if the user has a UI defined in the code, then it would be overriden by the admin toolkit.

PR #1069, opened 2025-02-18
fix network entities for the editor

gonpombo8:


packages/@dcl/inspector/src/lib/data-layer/host/utils/migrations/fix-network-entity-values.ts
```diff
@@ -0,0 +1,35 @@
...
+import { INSPECTOR_ENUM_ENTITY_ID_START } from '../../../../sdk/enum-entity'
+
+/**
+ * Fixes duplicate entity IDs in NetworkEntity components within the engine.
+ *
+ * This function scans through all entities that have the NetworkEntity component
+ * and identifies any duplicate entity IDs. If duplicates are found, it reassigns
+ * unique entity IDs to each NetworkEntity component to ensure all entity IDs are unique.
+ */
+export function fixNetworkEntityValues(engine: IEngine) {
```
cyaiox, line 12: will this logic re-assign a new entityId for all the entities if one of them is duplicated?

packages/@dcl/inspector/src/lib/data-layer/host/utils/migrations/fix-network-entity-values.ts
```diff
@@ -0,0 +1,35 @@
...
+  const NetworkEntity = engine.getComponentOrNull(CoreComponents.NETWORK_ENTITY) as SdkComponents['NetworkEntity']
+
+  function needsFixing(): boolean {
+    const usedEntities: Set<Entity> = new Set()
+    for (const [_, value] of engine.getEntitiesWith(NetworkEntity)) {
+      if (usedEntities.has(value.entityId)) {
+        return true
```
cyaiox, line 19: based on the previous question, could we use the method `getNextEnumEntityId` here?

packages/@dcl/inspector/src/lib/data-layer/host/utils/migrations/fix-network-entity-values.ts
gonpombo8: YES

packages/@dcl/inspector/src/lib/data-layer/host/utils/migrations/fix-network-entity-values.ts
gonpombo8: Yes, that was my first approach but the algorithm for that is to iterate all the entities to look for entities not being used.
It was simpler to just re-generate them IMO

cyaiox (approved):

PR #1072, opened 2025-02-27
fix several issues

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/415
closes https://github.com/decentraland/creator-hub/issues/255
closes https://github.com/decentraland/creator-hub/issues/261
closes https://github.com/decentraland/creator-hub/issues/418
closes https://github.com/decentraland/creator-hub/issues/256

### DIFF
--- a/packages/@dcl/inspector/src/components/Renderer/Shortcuts/Shortcuts.tsx
+++ b/packages/@dcl/inspector/src/components/Renderer/Shortcuts/Shortcuts.tsx
@@ -16,6 +16,8 @@ import { Props } from './types'
 import './Shortcuts.css'
 
 const ICON_SIZE = 18
+const isMac = /Mac|iPhone|iPod|iPad/.test(navigator.userAgent)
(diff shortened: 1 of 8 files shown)


packages/@dcl/inspector/src/components/Renderer/Shortcuts/Shortcuts.tsx
```diff
@@ -16,6 +16,8 @@ import { Props } from './types'
 import './Shortcuts.css'
 
 const ICON_SIZE = 18
+const isMac = /Mac|iPhone|iPod|iPad/.test(navigator.userAgent)
```
cyaiox, line 19: wdyt about moving this validation to a utils/helper file?

nicoecheza: I don' think it's necessary for now. It's a one-liner check. We should do it if we keep using it in other places but for now I think it's ok

cyaiox (approved):

PR #1090, opened 2025-03-31
fix: Animation weight value

cyaiox:
This PR validates the animation's weight when mapping the value.

This will ensure the animation to be executed has a valid weight between 0 and 1.

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
+++ b/packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
@@ -12,18 +12,18 @@ export function toNumber(value: string | number, div: number = 100) {
 }
 
 export function isValidWeight(weight: string | undefined): boolean {
-  const value = (weight ?? 0).toString()
+  const value = (weight ?? '').toString()
   return !isNaN(parseFloat(value)) && parseFloat(value) >= 0 && parseFloat(value) <= 100
 }
 
 export function isValidSpeed(speed: string | undefined): boolean {
-  const value = (speed ?? 0).toString()
+  const value = (speed ?? '').toString()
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
```diff
@@ -22,14 +22,17 @@ export function isValidSpeed(speed: string | undefined): boolean {
...
-  return animations.map(($) => ({
-    clip: $.name,
-    playing: !!$.isPlaying,
-    weight: $.weight ?? 1,
-    speed: $.speedRatio ?? 1,
-    loop: $.loopAnimation ?? false,
-    shouldReset: false
-  }))
+  return animations.map(($) => {
+    const weight = isValidWeight($.weight.toString()) ? $.weight : 1
```
gonpombo8, line 26: the default in the validWeight is `0` but here is `1`

cyaiox: Hey @gonpombo8, thanks for the review! 

we are using the default value `1` here to keep consistency with the `PBAnimationState` component.

updated the `isValidWeight` method to consider `undefined` as an invalid value

packages/@dcl/inspector/src/components/EntityInspector/AnimatorInspector/utils.ts
```diff
@@ -12,24 +12,27 @@ export function toNumber(value: string | number, div: number = 100) {
 }
 
 export function isValidWeight(weight: string | undefined): boolean {
-  const value = (weight ?? 0).toString()
+  const value = (weight ?? '').toString()
```
gonpombo8, line 15: maybe its better to have an early return ? 

if (!weight) return false

PR #1107, opened 2025-04-22
feat: add type safe entity names file

gonpombo8:
# Entity Names Type Generation

## Description
This PR introduces a new feature to automatically generate TypeScript definitions for entity names in the scene. It simplifies referencing scene entities by name with full type safety and IDE autocompletion support.

## Implementation Details
- Added a new function `generateEntityNamesType` to create a TypeScript file with string literal types for entity names
- Automatically generates this file whenever the scene composite is saved
- Creates both a string literal union type and a constant object for better developer experience
- Entity names are properly sanitized for valid TypeScript identifiers

## Benefits
- Type safety when referencing entities by name in code
- Autocomplete support in IDE for all named entities in the scene
- Prevents runtime errors from typos in entity name strings
- Makes refactoring entity names easier as TypeScript will catch all references

## Example Usage
```typescript
import { EntityNames } from '../assets/scene/entity-names'

// Auto-completion support via constant object
const GroundBoedoOrNull = engine.getEntityOrNullByName(EntityNames.Ground_Boedo1) 
const GroundBoedo = engine.getEntityByName<EntityNames>(EntityNames.Ground_Boedo1)
console.log(GroundBoedo)

```

This implementation helps reduce runtime errors and improves the developer experience when working with named entities in Decentraland scenes.

### DIFF
--- a/packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
+++ b/packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
@@ -8,9 +8,12 @@ import {
   IEngine,
   LastWriteWinElementSetComponentDefinition,
   PutComponentOperation,
-  getCompositeRootComponent
+  getCompositeRootComponent,
+  Name,
+  NameType
 } from '@dcl/ecs'
 import { ReadWriteByteBuffer } from '@dcl/ecs/dist/serialization/ByteBuffer'
+import { FileSystemInterface } from '../../types'
 
 function componentToCompositeComponentData<T>(
   $case: 'json' | 'binary',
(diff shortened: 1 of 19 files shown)


packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
```diff
@@ -128,3 +130,85 @@ export function dumpEngineToCrdtCommands(engine: IEngine): Uint8Array {
...
+    let fileContent = `// Auto-generated entity names from the scene\n\n`
+    fileContent += `/**\n * Union type containing all entity names in the scene.\n * Use for type-safe references to entities by name.\n */\n`
+    fileContent += `export type I${typeName} = `
+
+    if (uniqueNames.length === 0) {
+      fileContent += `never\n`
+    } else {
+      fileContent += `\n`
+      for (const key in uniqueNames) {
+        const isLast = Number(key) === uniqueNames.length - 1
```
nicoecheza, line 194: is this casting to `Number` ok? Because it's casting strings 🤔

packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
```diff
@@ -128,3 +130,85 @@ export function dumpEngineToCrdtCommands(engine: IEngine): Uint8Array {
...
+        fileContent += `  | "${uniqueNames[key]}"${isLast ? '\n' : '\n'}`
```
nicoecheza, line 195: this one is the same for both cases: `${isLast ? '\n' : '\n'}`

packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
```diff
@@ -128,3 +130,85 @@ export function dumpEngineToCrdtCommands(engine: IEngine): Uint8Array {
...
+    for (const { original, valid } of validNames) {
+      fileContent += `  ${valid}: "${original}",\n`
+    }
+
+    fileContent += `} as const\n`
+
+    // Write to file
+    await fs.writeFile(outputPath, Buffer.from(fileContent, 'utf-8'))
+  } catch (e) {
+    console.error('Faield to generate entity names types', e)
```
nicoecheza, line 212: `Failed`

packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
```diff
@@ -128,3 +130,85 @@ export function dumpEngineToCrdtCommands(engine: IEngine): Uint8Array {
...
+    // Generate valid TypeScript identifiers
+    const validNames = uniqueNames.map((name) => {
+      // Create a valid TypeScript identifier
+      const validName = name
+        .replace(/[^a-zA-Z0-9_]/g, '_') // Replace non-alphanumeric chars with underscore
+        .replace(/^[0-9]/, '_$&') // Prepend underscore if starts with number
+
+      // Ensure no duplicate keys after validation
+      return { original: name, valid: validName }
+    })
```
nicoecheza, line 182: if the user creates two entities with names like: 'test-1' and 'test_1', the latter will be used as valid for both cases below on line 203. I think this may produce a bug?

tested it here: https://tsplay.dev/w2yyxW

nicoecheza (changes requested):

packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
gonpombo8: its a ` for key in ` loop in an array, so its returning the index key

packages/@dcl/inspector/src/lib/data-layer/host/utils/engine-to-composite.ts
gonpombo8: good catch, solved

nicoecheza (approved):

PR #1111, opened 2025-05-05
fix: Texture being flipped vertically

nicoecheza:
closes https://github.com/decentraland/creator-hub/issues/356


gonpombo8 (approved):
i guess this fixes it

PR #1114, opened 2025-05-07
Support jpg textures

alejandralevy:
Close: https://github.com/decentraland/creator-hub/issues/405

Support URLs on textures
Missing: there's a bug when the user adds a file that's not on the scene
Screenshot 2025-05-09 at 12 49 36


cyaiox (approved):
awesome work!!

PR #1119, opened 2025-05-13
Update global value on Audio Source component

alejandralevy:
The global property was not being saved on the audio source component. Now we are saving it but only when you select a valid audio source. If not, it does not keep the selection. This issue was already discussed with Ezio and it will be hanlded in another ticket, as it's something that it's happening in more than one component.

https://github.com/user-attachments/assets/5a991bdd-e287-43e4-b333-33ef3c467a87

### DIFF
--- a/packages/@dcl/inspector/src/components/EntityInspector/AudioSourceInspector/utils.ts
+++ b/packages/@dcl/inspector/src/components/EntityInspector/AudioSourceInspector/utils.ts
@@ -9,22 +9,26 @@ import { AudioSourceInput } from './types'
 export const fromAudioSource =
   (base: string) =>
   (value: PBAudioSource): AudioSourceInput => {
+    debugger
     return {
       audioClipUrl: removeBasePath(base, value.audioClipUrl),
       loop: value.loop,
       playing: value.playing,
-      volume: volumeFromAudioSource(value.volume)
+      volume: volumeFromAudioSource(value.volume),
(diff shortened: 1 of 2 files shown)


packages/@dcl/inspector/src/components/EntityInspector/AudioSourceInspector/utils.ts
```diff
@@ -9,22 +9,26 @@ import { AudioSourceInput } from './types'
 export const fromAudioSource =
   (base: string) =>
   (value: PBAudioSource): AudioSourceInput => {
+    debugger
```
gonpombo8, line 12:
```suggestion
```

packages/@dcl/inspector/src/components/EntityInspector/AudioSourceInspector/utils.ts
```diff
@@ -9,22 +9,26 @@ import { AudioSourceInput } from './types'
...
-      volume: volumeFromAudioSource(value.volume)
+      volume: volumeFromAudioSource(value.volume),
+      global: value.global
     }
   }
 
 export const toAudioSource =
   (base: string) =>
   (value: AudioSourceInput): PBAudioSource => {
+    debugger
```
gonpombo8, line 25:
```suggestion
```

cyaiox (approved):

PR #1134, opened 2025-06-16
Show modal when trying to upload an asset with an existing name

alejandralevy:
- Show warning to the user when is trying to upload a file that has a name that already exists
- User can confirm the replace or go back to change the name
- The thumbnail is updated after replacing the asset
- Includes a refactor of the code, moving the Slider logic to another component

Screenshot 2025-06-16 at 18 32 55
Screenshot 2025-06-16 at 18 33 42


DafGreco (approved):
✔️  PR reviewed and approved by QA on both platforms following instructions playing both happy and un-happy path.

Implementation correctly done in both platforms

PR #1141, opened 2025-06-25
Improve import emotes 

alejandralevy:
- Detecting whether a GLB/GLTF file is an emote or a regular model to show the correct preview and UI behavior.
- Ensures that emotes have filenames ending with _emote. If not, the suffix is automatically appended to the name during processing.

To test:
- Import a model, check that it's uploaded correctly
- Import an emote, check that the preview is working and that it's uploaded correctly
- If the emote has the _emote at the end of the name, it will be uploaded directly. Otherwise, it will be automatically added to the emote name

Screenshot 2025-06-25 at 17 29 42

### DIFF
--- a/packages/@dcl/inspector/src/components/ImportAsset/Slider/AssetSlides.tsx
+++ b/packages/@dcl/inspector/src/components/ImportAsset/Slider/AssetSlides.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect } from 'react'
+import { useCallback } from 'react'
 import { IoIosArrowBack, IoIosArrowForward } from 'react-icons/io'
 import cx from 'classnames'
 
@@ -8,10 +8,10 @@ import { Loading } from '../../Loading'
 
 import { getAssetSize, getAssetResources } from '../utils'
 import { Asset } from '../types'
-import { Thumbnails } from './types'
+import { AssetWithEmote, Thumbnails } from './types'
 
 interface AssetSlidesProps {
-  uploadedAssets: Asset[]
+  uploadedAssets: AssetWithEmote[]
   currentSlide: number
   screenshots: Thumbnails
   onSlideChange: (newSlide: number) => void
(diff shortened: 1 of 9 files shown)


packages/@dcl/inspector/src/components/ImportAsset/utils.ts
```diff
@@ -175,12 +176,14 @@ async function processFile(file: File): Promise<BaseAsset> {
...
+  const isEmoteFile = await isEmote(file)
+
   const sizeError = validateFileSize(file.size)
   if (sizeError) {
-    return { blob: file, name, extension, error: sizeError }
+    return { blob: file, name, extension, error: sizeError, isEmote: isEmoteFile }
   }
 
-  return { blob: file, name, extension }
+  return { blob: file, name, extension, isEmote: isEmoteFile }
```
nicoecheza, line 186: I think declaring if the asset `isEmote` isn't something the import-asset flow should care about, since this should only care about files types and not differentiation between usage of the files. 
Maybe we could move the check of `isEmote` to the `Slider` component, since that's where we want to differentiate between usage of the same file type

packages/@dcl/inspector/src/components/ImportAsset/Slider/AssetSlides.tsx
```diff
@@ -41,6 +41,19 @@ export function AssetSlides({
...
+  }
+
+  useEffect(() => {
+    uploadedAssets.forEach((asset, index) => {
+      if (asset.isEmote && !hasEmoteName(asset)) {
+        const newName = `${asset.name}_emote`
+        onNameChange(index)(newName)
+      }
+    })
+  }, [])
```
nicoecheza, line 55: following the same logic from the comment other comment, I think we should do this in `https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/inspector/src/components/ImportAsset/Slider/Slider.tsx#L20` altogether. This way we avoid the `useEffect` usage.

nicoecheza (changes requested):

nicoecheza (approved):

PR #1162, opened 2025-07-18
Feat: out of bounds entities changes

alejandralevy:
- Remove the "too many triangles" text.
- Modify the Scene optimization menu.
- Display out-of-scene entities only when selected.
- Adjust the style: eliminate the red box and use yellow borders exclusively on entities.
- Introduce a warning in the hierarchy tree for any entity (or its children) out of the scene.
- Only show the yellow borders for entities out of the scene if the entity is selected
- If a entity is otside the limits, when selecting a parent, also show the yellow borders

Screenshot 2025-07-18 at 15 29 05
Screenshot 2025-07-18 at 15 29 29
Screenshot 2025-07-18 at 15 30 04

### DIFF
--- a/packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
+++ b/packages/@dcl/inspector/src/lib/babylon/decentraland/EcsEntity.ts
@@ -264,17 +264,12 @@ function updateMeshBoundingBoxVisibility(entity: EcsEntity, mesh: BABYLON.Abstra
   const isParentSelected = hasSelectedParent(entity, context)
   const children = entity.gltfContainer ? entity.gltfContainer.getChildMeshes(false) : entity.getChildMeshes(true)
 
-  const shouldShowBoundingBox =
-    (isSelected && isEntityOutsideLayout(mesh)) || (isEntityOutsideLayout(mesh) && isParentSelected)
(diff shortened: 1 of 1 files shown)


packages/@dcl/inspector/src/components/Tree/Tree.tsx
```diff
@@ -284,10 +290,38 @@ export function Tree<T>() {
...
 
-      const isEntity = useMemo(() => {
+      const isValidEntity = useMemo(() => {
         return typeof value !== 'string'
       }, [value])
 
+      const isEntity = useMemo(() => {
+        if (typeof value === 'string') return false
+        return value !== ROOT && value !== PLAYER && value !== CAMERA
+      }, [value])
```
nicoecheza, line 300: this should be a check in the useTree hook. The Tree component should not know about `ROOT,etc` entities

nicoecheza (changes requested):

packages/@dcl/inspector/src/components/Tree/Tree.tsx
nicoecheza: there is an `isRoot` function in the `useTree` hook that we can use 😋

nicoecheza (approved):

PR #1185, opened 2025-07-31
chore: stringify console log arguments

pravusjif:
Added console log wrapper to stringify arguments, to avoid printing `Object object` on those cases.

With this change, if a scene code calls `console.log(whateverObject)` it doesn't print `Object object` anymore.

When putting the arg in an inline string, it will still need the `JSON.Stringify()` because as as soon as the object is included as: `'print object: ${whateverObject}'` it gets automatically casted to String by Javascript and Javascript always stringifies objects as "Object object".


pravusjif:
Will do the same on Explorer side

PR #1333, opened 2026-02-26
feat: Add getCompositeProvider/setCompositeProvider methods in the engine

cyaiox:
## Summary

Expose composite-provider hooks on the engine so scenes can resolve and load `.composite` files at runtime without going through the SDK barrel.

- **`engine.setCompositeProvider(provider)` / `engine.getCompositeProvider()`** — new `IEngine` methods. The SDK now registers its built-in `compositeProvider` on boot (`packages/@dcl/sdk/src/index.ts`), so `engine.getCompositeProvider()` is non-null in any scene that imports `@dcl/sdk`.
- **`CompositeProvider.loadComposite(src)`** — new optional async method on the provider interface. The SDK implementation reads the file via `~system/Runtime.readFile`, auto-detects JSON (`{`-prefixed) vs protobuf-binary `.composite.bin`, decodes accordingly, caches the resource, and returns it. If JSON content is encountered without a `TextDecoder` available, the error message points users at `@dcl/sdk/ethereum-provider` to install the polyfill.
- **`InstanceCompositeOptions.useRootComponent`** — opt-out flag for `Composite.instance()`. When set to `false`, the call skips creating/reading the `CompositeRootComponent` on the destination entity. Defaults to `true` (existing behavior).
- **Runtime polyfill import path fix** — `@dcl/sdk/ethereum-provider`'s text-encoder shim now imports `setGlobalPolyfill` from `@dcl/ecs/dist/runtime/globals` directly, so the polyfill installs reliably even when the main `@dcl/ecs` entrypoint is tree-shaken.

CRDT snapshots were regenerated because the provider is now registered at boot.

## Test plan

- [ ] `npm test` — full suite green (snapshots match, `loadComposite` mocked-runtime tests pass).
- [ ] Bundle a scene with `@dcl/sdk` and confirm `engine.getCompositeProvider()` returns a non-null provider at runtime.
- [ ] In a scene, call `await engine.getCompositeProvider()!.loadComposite!('some.composite')` against a JSON `.composite` file and verify the resource is returned and cached (second call returns it via `getCompositeOrNull`).
- [ ] Repeat with a `.composite.bin` file and confirm the binary branch is taken.
- [ ] Trigger the missing-`TextDecoder` path in a runtime without the polyfill and confirm the error message points to `@dcl/sdk/ethereum-provider`.
- [ ] Call `Composite.instance(engine, resource, provider, { useRootComponent: false })` and confirm no `CompositeRootComponent` is created on the root entity, while child entities and other components still spawn correctly.
- [ ] Smoke-test an existing scene that relies on the default `Composite.instance()` path to confirm no regression (the `useRootComponent` default stays `true`).


packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -28,16 +35,47 @@ export async function getAllComposites(
...
+  // Explicitly exclude `node_modules` so a transitive dependency that ships
+  // a `composite.json` cannot inject a component-schema definition that the
+  // scene's auto-generated entrypoint then pre-registers (which would let
+  // it shadow any non-`core::` component name).
+  const additionalCompositeFiles = globSync('**/composite.json', {
+    cwd: workingDirectory,
+    ignore: ['node_modules/**', '**/node_modules/**']
+  })
+  components.logger.info(
+    `[composite] scanning '${workingDirectory}': ${watchFiles.length} .composite, ${additionalCompositeFiles.length} composite.json file(s)`
```
RocioCM: Are these logs expected to be kept?

RocioCM (approved):

packages/@dcl/sdk-commands/src/logic/composite.ts
cyaiox: good catch. removing it

nicoecheza (changes requested):
I have some concerns about this approach. Feel like we are adding a lot of code that we might not need to achieve this behaviour.
Just throwing this since it makes sense in my head, but what do you think about this approach?:

## `packages/@dcl/ecs/src/composite/`

1. **index.ts**: change `Composite.instance` return type from `void` to `Entity`. forward `instanceComposite` return value
2. **instance.ts**: only write `CompositeRootComponent` on the spawned root entity when the composite actually contains nested child composites (i.e. when the recursion phase processed a `composite::root`-named component). Leaf composites emit no bookkeeping component. No new option, no caller-side flag
3. **instance.ts**: handle schema registration inline: ~when a composite carries a non-`core::` component whose schema isn't yet defined on the engine, register it via a seal-bypassing path (new internal `defineComponentFromSchemaForComposite`, scoped to this call site only). Removes the need for build-time pre-registration~. Register at setCompositeProvider() time. The SDK's boot-time `setCompositeProvider(compositeProvider)` runs pre-seal. Extend the provider with a schemas: `Array<{name, jsonSchema}>` field. `setCompositeProvider` walks it and calls `engine.defineComponentFromSchema` immediately. Pre-seal, deterministic, no build-time scanner, no runtime bypass. *asset-packs* ship their schemas as data alongside the composite and the provider exposes them, the engine registers them at boot
4. **instance.ts**: emit a warning (don't throw) when `getCompositeOrNull(childCompositePath)` returns `null` during recursion
5. **New file `composite/provider-registry.ts**: module-level slot:
   ```ts
   let currentProvider: CompositeProvider | null = null
   export function setCompositeProvider(p: CompositeProvider) { currentProvider = p }
   export function getCompositeProvider(): CompositeProvider | null { return currentProvider }
   ```
   re-export from `composite/index.ts`
6. **instance.ts**: add optional `loadComposite?: (src: string) => Promise<CompositeResource>` to the `CompositeProvider` type.

## `packages/@dcl/sdk/src/`

7. **composite-provider.ts**: replace the array-based cache with a `Map<string, Composite.Resource>`; extract `decodeFromLoader` and `decodeFromBytes` helpers => implement `loadComposite(src)` using `~system/Runtime.readFile` + `decodeFromBytes` + cache
8. **ethereum-provider/text-encoder.ts**: remove the `/* @__PURE__ */` annotation on `polyfillTextEncoder`
9. **index.ts**: one-line `setCompositeProvider(compositeProvider)` at boot

usage:

```ts
import { engine, Composite, getCompositeProvider } from '@dcl/ecs'

const provider = getCompositeProvider()!
const resource = provider.getCompositeOrNull(src) ?? await provider.loadComposite!(src)
const rootEntity = Composite.instance(engine, resource, provider, opts)
```

we keep `IEngine` and `sdk-commands` untouched

gonpombo8 (commented):
can i ask why we need this ? composites at runtime ? whats the use case ?

nearnshaw:
Hey @gonpombo8!
We always thought of composites as something comparable to prefabs in Unity. Every smart item has its own composite, with this you could dynamically instance smart items as part of a game loop. Today most smart items are just one entity, but keep in mind that a smart item can be a lot more complex.

Lately I've been experimenting with expanding the Creator Hub to also support switching scope, so that instead of editing the whole scene, you're editing a single composite for an item. Inside your project you can define a complex hierarchy of entities that make up a composite, setting relative positions and other configurations visually, and then via code instance them dynamically.
You could also use this for editing various levels for a game, and swap those dynamically.
Here's a video of my latest experiment w that.

https://github.com/user-attachments/assets/3cb50579-92a7-497f-8382-1487313d9505

packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -28,16 +35,44 @@ export async function getAllComposites(
...
+  // and defeat on-demand spawning via `engine.addEntityFromComposite`).
+  // Scan asset-pack-style composites for component-schema pre-registration.
+  // Explicitly exclude `node_modules` so a transitive dependency that ships
+  // a `composite.json` cannot inject a component-schema definition that the
+  // scene's auto-generated entrypoint then pre-registers (which would let
+  // it shadow any non-`core::` component name).
+  const additionalCompositeFiles = globSync('**/composite.json', {
+    cwd: workingDirectory,
+    ignore: ['node_modules/**', '**/node_modules/**']
+  })
```
nicoecheza: i think we can combine this 2 right?

packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -50,6 +85,27 @@ export async function getAllComposites(
...
+      additionalComposites.push(Composite.fromJson(json))
+    } catch (err: any) {
+      // Not every `composite.json` in the tree is a valid composite definition
+      // (some are unrelated config). Log at debug to keep the build output
+      // quiet for legitimate non-composite JSON, but visible when investigating.
+      const message = err?.message ? String(err.message).slice(0, 200) : String(err).slice(0, 200)
+      components.logger.debug(`[composite] skipped '${file}': ${message}`)
+    }
+  }
+
```
nicoecheza: same as previous comment (merge with existing loop)

packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -103,10 +164,68 @@ export async function getAllComposites(
...
+      componentRegistrationsByName.set(component.name, {
+        name: component.name,
+        jsonSchema: component.jsonSchema
+      })
+    }
+  }
+  const componentRegistrations = Array.from(componentRegistrationsByName.values())
+  if (componentRegistrations.length > 0) {
+    components.logger.debug(`[composite]   components: ${componentRegistrations.map((r) => r.name).join(', ')}`)
+  }
```
nicoecheza: dont think we need this log, but in case we do, i think `componentRegistrations.size` should be enough

packages/@dcl/sdk-commands/src/logic/composite.ts
```diff
@@ -103,10 +164,68 @@ export async function getAllComposites(
...
+    .join('\n')
+
+  return `import { engine, Schemas } from '@dcl/ecs'
+import { getCompositeRootComponent } from '@dcl/ecs'
+// Pre-register composite::root so runtime addEntityFromComposite recursion
+// (via getCompositeRootComponent inside instanceComposite) doesn't trip the
+// seal for scenes that have no main.composite to instance at scene boot.
+getCompositeRootComponent(engine)
+${registrationLines}
+`
```
nicoecheza: is this needed?

test/ecs/composites/data-binary.composite.scene-snapshot.json
```diff
@@ -124,41 +124,6 @@
...
-      "src": "empty.composite",
-      "entities": [
-        {
-          "src": 0,
-          "dest": 513
-        }
-      ]
-    },
-    "514": {
-      "src": "one-transform.composite",
```
nicoecheza, line 137: is it ok to be deleting all this from the tests?

nicoecheza (changes requested):

packages/@dcl/sdk-commands/src/logic/composite.ts
cyaiox: good catch! not required with the new design

test/ecs/composites/data-binary.composite.scene-snapshot.json
cyaiox: yep, these are regenerated snapshots. with the new `Composite.instance` behavior, `composite::root` is written only when a composite instance's nested child composites, so for the ones without children, they aren't emitted.

nicoecheza (approved):

PR #1381, opened 2026-04-21
fix(sdk-commands): make asset-packs bundling conditional on composite content

nachomazzara:
## Summary

- `isEditorScene` previously returned `true` whenever `main.composite` existed, unconditionally injecting `initAssetPacks` even for pure-static scenes with no smart items
- Now reads and parses the composite JSON, injecting `initAssetPacks` only when the composite contains `asset-packs::*` components other than `asset-packs::Script` (which is build-time only)
- Also removes the pre-existing dead `setSyncEntity` import and documents the watch-mode caveat

### Root Cause

`isEditorScene` in `project-validations.ts` performed only a file-existence check. The asset-packs runtime (`initAssetPacks`) is only needed when the composite contains runtime `asset-packs::*` components (Actions/Triggers/States). It is **not** needed for:
- Pure static scenes (`core::Transform`, `core::GltfContainer`, etc.)
- Scenes with only `asset-packs::Script` components (handled by `generateInitializeScriptsModule` separately)

### Behavior Change

| Flow | Composite content | Before | After |
|---|---|---|---|
| No composite | File absent | `false` | `false` (unchanged) |
| Pure static | Only `core::*` | `true` ❌ | `false` ✅ |
| Smart items | `asset-packs::Trigger`, etc. | `true` | `true` (unchanged) |
| Script only | Only `asset-packs::Script` | `true` ❌ | `false` ✅ |
| Script + smart items | `asset-packs::Script` + `asset-packs::Trigger` | `true` | `true` (unchanged) |
| Malformed JSON | Invalid JSON | `true` ❌ | `false` ✅ |

### Key implementation notes
- `Composite.fromJson` does not throw on malformed input — try/catch around `JSON.parse` is essential
- `getAllComposites` is called after `isEditorScene` in `bundleSingleProject`, so composite data can't be reused — the function reads `main.composite` independently (one extra file read, minimal overhead)
- `editorScene` is baked into esbuild's `stdin` at context-creation time and is NOT re-evaluated during watch-mode rebuilds — this is a pre-existing limitation, now documented with a `// NOTE:` comment

## Changes

```
packages/@dcl/sdk-commands/src/logic/project-validations.ts  — isEditorScene: content-based detection
packages/@dcl/sdk-commands/src/logic/bundle.ts               — remove dead setSyncEntity import, add watch-mode note
test/sdk-commands/logic/project-validations.spec.ts          — 11 unit tests for all edge cases (new file)
```

## Testing

New test file `test/sdk-commands/logic/project-validations.spec.ts` covers all 10 cases:
- File absent → `false`
- File exists, no `components` key → `false`
- Empty `components: []` → `false`
- Only `core::*` components → `false`
- Only `asset-packs::Script` → `false`
- One non-Script `asset-packs::` component → `true`
- `asset-packs::Script` + `asset-packs::Trigger` → `true`
- Invalid JSON → `false`
- `components` is not an array → `false`
- Component items without `name` field → `false`

## Closes

https://github.com/decentraland/creator-hub/issues/1304

---
🤖 Created via Slack with Claude
Requested by Nico Earnshaw via Slack

### DIFF
--- a/packages/@dcl/sdk-commands/src/logic/bundle.ts
+++ b/packages/@dcl/sdk-commands/src/logic/bundle.ts
@@ -165,10 +165,6 @@ type SingleProjectOptions = CompileOptions & {
 
 export async function bundleSingleProject(components: BundleComponents, options: SingleProjectOptions) {
   printProgressStep(components.logger, `Bundling file ${colors.bold(options.entrypoint)}`, 1, MAX_STEP)
-  // NOTE: editorScene is evaluated once and baked into stdin at esbuild context-creation
-  // time. In watch mode it is NOT re-evaluated on context.rebuild(). If a scene's
-  // editor status changes (e.g. first smart item added to an empty composite) the
-  // watch process must be restarted for initAssetPacks to be injected correctly.
   const editorScene = await isEditorScene(components, options.workingDirectory)
 
   // Pre-compute composite data so we can inject maxCompositeEntity via esbuild define.
--- a/packages/@dcl/sdk-commands/src/logic/bundle.ts
+++ b/packages/@dcl/sdk-commands/src/logic/bundle.ts
@@ -83,12 +83,8 @@ ${
   `
 import { syncEntity } from '@dcl/sdk/network'
 import players from '@dcl/sdk/players'
-import { initAssetPacks, setSyncEntity } from '@dcl/asset-packs/dist/scene-entrypoint'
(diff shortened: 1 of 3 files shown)


RocioCM (approved):

RocioCM:
Evidence: index.js size for an scene with main.composite without components from asset-packs
| BEFORE | AFTER |
| --- | --- |
| image | image |

PR #1386, opened 2026-05-03
feat(react-ecs): add ScreenInsetArea component

kuruk-mm:
## Summary

Adds a React-ECS layout component bound to a new field of `UiCanvasInformation` in the protocol — replacing the previous `SafeAreaContainer` (which conflated two distinct concepts):

- **`ScreenInsetArea`** — reads `UiCanvasInformation.screenInsetArea` (the device safe margins: notch, status bar, home indicator, rounded corners on mobile; `(0,0,0,0)` on desktop). Use it when you need to keep critical UI out of hardware-reserved regions.

The component is absolutely positioned with the reported insets, so a child sized `100%x100%` fills the safe region exactly and reactively updates as the renderer reports new values.

> A separate `InteractableArea` component (reading `UiCanvasInformation.interactableArea`, the region not covered by the explorer's HUD) will land in a follow-up PR.

## API

```tsx
import { ScreenInsetArea } from '@dcl/sdk/react-ecs'

<ScreenInsetArea>     {/* lays out inside the device safe margins */}
  <MyHud />
</ScreenInsetArea>
```

`positionType` and `position` on `uiTransform` are owned by the container — overrides there are ignored both at the type level and at runtime. All other `EntityPropTypes` (events, `uiBackground`, layout) are forwarded.

## Why split out of `SafeAreaContainer`

The previous iteration exposed a single `SafeAreaContainer` that read `interactableArea`. That conflated two distinct concepts:

- **Hardware insets** (notch, home indicator) — a property of the device, not the explorer.
- **HUD insets** (chat, minimap) — a property of the explorer's UI state, can change every frame.

The protocol now exposes both as separate fields. This PR introduces the hardware-insets variant; the HUD-insets variant ships in a follow-up so each can be reviewed and adopted independently.

## Implementation

Same pattern as `getUiScaleFactor` / `setUiScaleFactor`:

- `system.ts`'s `UiScaleSystem` writes `canvasInfo.screenInsetArea` to a module-level variable each tick, guarded by an owner symbol so unrelated systems can't stomp it.
- The component reads from that variable during render. `destroy()` resets it.

This avoids the engine-context problem that would otherwise prevent the component from being testable: a JSX function component has no engine reference, and the global `engine` import isn't the same instance as the one returned by `Engine()` in unit tests.

## Protocol dependency

Depends on the published `@dcl/protocol@next` build (`1.0.0-26120507017.commit-56ef588`), which adds `screen_inset_area` (field 5) to `PBUiCanvasInformation`.

## Test plan

- [x] `screen-inset-area.spec.tsx` covers: positions absolutely with reported insets, reacts to mid-flight changes, falls back to zeros before `UiCanvasInformation` exists, forwards `uiTransform` props but ignores `positionType` / `position` overrides
- [x] `make build` regenerates `ui_canvas_information.gen.ts` with the `screenInsetArea` field from the bumped protocol
- [x] `make test` passes on the new spec and the existing `react-ecs` suite
- [x] `make lint` clean for `@dcl/react-ecs`
- [x] Refresh `test/snapshots/production-bundles/ui.ts.crdt` if numbers shift after the protocol bump


packages/@dcl/react-ecs/src/components/utils.ts
```diff
@@ -90,6 +96,32 @@ export function resetUiScaleFactor(owner?: symbol): void {
...
+}
+
+/**
+ * @internal
+ */
+export function resetScreenInsetArea(owner?: symbol): void {
+  if (owner && screenInsetAreaOwner !== owner) return
+  screenInsetAreaOwner = undefined
+  screenInsetArea = { ...ZERO_INSETS }
+}
```
RocioCM: Claude misunderstood the usage of owner here at first, worth adding some explanatory comments of the owner mechanism:

> Looking at the full code, the owner is not a write-protection mechanism — it's a reset-protection mechanism for the multi-instance case.
> 
> The real subtleties worth a comment in the code (neither setter has one today) are:
> - "last writer claims ownership"
> - "reset is no-op for non-owners, but no-owner reset always wins" — that second part is what makes the test path work and is the least obvious from the code.

packages/@dcl/react-ecs/src/components/utils.ts
```diff
@@ -12,6 +12,12 @@ import { parseUiTransform } from './uiTransform'
 let uiScaleFactor = 1
 let uiScaleOwner: symbol | undefined = undefined
 
+const ZERO_INSETS = { top: 0, left: 0, right: 0, bottom: 0 } as const
+type Insets = { top: number; left: number; right: number; bottom: number }
```
RocioCM: The local `Insets = { top, left, right, bottom }` shape duplicates `BorderRect` from the protocol. Since the setter is fed directly from `canvasInfo.screenInsetArea` (already a `BorderRect`), importing and reusing that type would avoid a parallel definition.

RocioCM (commented):
Hey Mateo, the code looks good, I share some suggestions from Claude in the comments. Additionally:
1. Tests are not passing, please fix them before merging.
2. When testing in mobile, I added a green overlay of 100% x 100% inside ScreenInsetArea, but I get this uncovered area despite not having native UI there (even when the floating status bar or navigation is open, it's on top or right, not on left). Is it expected?

Image

kuruk-mm:
> 2\. When testing in mobile, I added a green overlay of 100% x 100% inside ScreenInsetArea, but I get this uncovered area despite not having native UI there (even when the floating status bar or navigation is open, it's on top or right, not on left). Is it expected?

That is the expected behavior. The phones reports the borders and all content that is inside ScreenInsetArea, should be out of the margins... similar how Safe Areas works in React Native (https://reactnavigation.org/docs/handling-safe-area/)

RocioCM (approved):

PR #1388, opened 2026-05-04
feat(react-ecs): expose createContext and useContext

pabloes:
Re-export React's Context API from `@dcl/react-ecs` so scenes can use the standard Provider/Consumer pattern. The underlying `react-reconciler` already supports context at runtime; this PR only adds the public surface plus typings:

- `Provider<T>`, `Consumer<T>`, `Context<T>` interfaces
- `createContext<T>(defaultValue)` factory
- `useContext<T>(context)` hook

Enables ergonomic theming / layout-context flows in scenes (e.g. a single `<LayoutContext.Provider value={CONTEXT.DIALOG}>` wraps a popup subtree without prop-drilling).

Example: https://github.com/dcl-regenesislabs/bevy-ui-scene/commit/0548e9a0284cf3aaa73b9c025678b837430c6170


pabloes:
not necessary as we can reference react from scene

PR #1422, opened 2026-06-16
fix(react-ecs): don't drop an optional prop set for the first time

robtfm:
## Problem

`propsChanged` (the reconciler diff) only iterated the keys present in `prevProps`:

```ts
for (const k in prevProps) { ... }
```

So a prop that appears for the **first time** — present in `nextProps` but absent from `prevProps` — is never compared, and its change is silently dropped. Any optional UI prop that goes from `undefined` to a value on a later render hits this: the first change is lost, every subsequent one works, and seeding a non-null initial value avoids it.

(On `protocol-squad` this manifests as the reported "first scroll-to-target is dropped" bug, because `scrollPosition` is omitted from the parsed transform while undefined. That branch has the same fix in its own PR with a `scrollPosition` integration test; `scrollPosition` doesn't exist on `main`, so here the regression is covered by a direct `propsChanged` unit test.)

## Fix

Add a second loop over `nextProps` for keys not already in `prevProps`. This is **allocation-free** (no `Object.keys`/`Set`) to keep this per-frame hot path cheap — the first loop still handles changed/removed keys, the second only catches first-appearance additions. It only *adds* previously-missed change detection, so it cannot introduce false negatives.

## Test

`test/react-ecs/reconciler-utils.spec.ts` asserts a key present only in `nextProps` is detected, a changed existing key still is, and an unchanged one returns undefined. The first case fails without the fix and passes with it.

### DIFF
--- a/packages/@dcl/react-ecs/src/reconciler/utils.ts
+++ b/packages/@dcl/react-ecs/src/reconciler/utils.ts
@@ -38,9 +38,22 @@ export function propsChanged<K extends keyof EntityComponents>(
   }
 
   const changes: Partial<EntityComponents[K]> = {}
+  // Iterate prevProps (this also catches removed keys, where nextProps[k] is undefined)...
   for (const k in prevProps) {
     const propKey = k as keyof typeof prevProps
     if (!isEqual(prevProps[propKey], nextProps[propKey])) {
+      changes[propKey] = nextProps[propKey]
+    }
+  }
+  // ...then catch keys present only in nextProps (an optional prop set for the first time,
+  // e.g. an undefined value becoming defined). Iterating prevProps alone would drop these.
+  // Two allocation-free loops are used instead of a Set union to keep this hot path cheap.
+  // These keys aren't in prevProps, so emitting them whenever they're defined is both cheaper
+  // than (and avoids the falsy-value trap of) an isEqual(undefined, ...) comparison.
+  for (const k in nextProps) {
+    if (k in prevProps) continue
+    const propKey = k as keyof typeof nextProps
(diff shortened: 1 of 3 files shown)


packages/@dcl/react-ecs/src/reconciler/utils.ts
```diff
@@ -38,12 +38,23 @@ export function propsChanged<K extends keyof EntityComponents>(
...
       changes[propKey] = nextProps[propKey]
     }
   }
+  // ...then catch keys present only in nextProps (an optional prop set for the first time,
+  // e.g. an undefined value becoming defined). Iterating prevProps alone would drop these.
+  // Two allocation-free loops are used instead of a Set union to keep this hot path cheap.
+  for (const k in nextProps) {
+    if (k in prevProps) continue
+    const propKey = k as keyof typeof nextProps
+    if (!isEqual(prevProps[propKey], nextProps[propKey])) {
```
nicoecheza: nit: i think we could just do (a bit cheaper than `isEqual` call):
```typescript
if (nextProps[propKey] !== undefined) {
  changes[propKey] = nextProps[propKey]
}
```

nicoecheza (approved):

packages/@dcl/react-ecs/src/reconciler/utils.ts
robtfm: agreed and amended, thx. can i merge it? not sure of process here

nicoecheza: yeah, go ahead. Already approved (it needs a rebase + snapshots)

PR #1433, opened 2026-07-01
fix: virtual screen missing device pixel ratio

pravusjif:
## Problem

The automatic UI scale factor derived from `setUiRenderer`'s `virtualWidth`/`virtualHeight` rendered UI too large on high-DPI screens (mobile, and HiDPI/Retina desktop). Scenes that computed and applied the scale manually looked correct, which pointed at the built-in calculation.

## Cause

`UiScaleSystem` computed `Math.min(width / virtualWidth, height / virtualHeight)` using the raw `UiCanvasInformation` dimensions, which are in **physical** pixels. The renderer interprets `UiTransform` px values as **logical** (CSS) pixels — it multiplies them by `devicePixelRatio`. The `vw`/`vh` path already accounts for this (it divides by `devicePixelRatio`), but the virtual-screen factor did not, so px-based UI was oversized by a factor of `devicePixelRatio`.

## Fix

`packages/@dcl/react-ecs/src/system.ts` — divide the computed scale by `devicePixelRatio` (falling back to `1` when missing/zero):

```ts
const ratio = devicePixelRatio || 1
const nextScale = Math.min(width / virtualWidth, height / virtualHeight) / ratio
```

This makes px-based sizing consistent with the `vw`/`vh` path and with the renderer's logical-px model.

## Tests

`test/react-ecs/virtual-scale-array.spec.tsx` — added two cases:
- scale factor is normalized by `devicePixelRatio` (dpr 2 → factor halved)
- a missing/zero `devicePixelRatio` falls back to `1` (no divide-by-zero)

Existing scale tests use `devicePixelRatio: 1`, so they are unaffected. Full react-ecs suite passes.

## Notes

- This is **not** a desktop no-op. HiDPI/Retina desktops report `devicePixelRatio > 1`, so the fix also corrects a real ~`devicePixelRatio`× oversize there. Scenes that visually compensated for the old (too-big) px behavior will render smaller — this is the correct behavior, but it is a behavior change worth calling out.

## QA TEST INSTRUCTIONS

During these tests IGNORE the UIcimage with yellow border at the center of the screen.

1. Enter `pravus.dcl.eth` with a DESKTOP explorer version
2. Confirm that you see the GREEN rectangle completely inside the screen, at the top-left
3. Enter the same world using the MOBILE explorer version
4. Confirm that you also see the GREEN rectangle completely inside the screen, at the top-left

Screenshot 2026-07-01 at 3 24 21 AM

Screenshot_20260701-032428


pravusjif:
`ui.tsx` used in test scene (deployed at `pravus.dcl.eth`):
```
import { Color4 } from '@dcl/sdk/math'
import ReactEcs, { ReactEcsRenderer, UiEntity } from '@dcl/sdk/react-ecs'

const VIRTUAL_WIDTH = 1920
const VIRTUAL_HEIGHT = 1080

const IMAGE = 'images/pravs-avatar.jpg'

export function setupUi() {
    ReactEcsRenderer.setUiRenderer(uiComponent, { virtualWidth: VIRTUAL_WIDTH, virtualHeight: VIRTUAL_HEIGHT })
}

const uiComponent = () => (
    // Full-screen container that centers its child.
    <UiEntity
        uiTransform={{
            width: '100%',
            height: '100%',
            positionType: 'absolute',
            position: { top: 0, left: 0 },
            justifyContent: 'center',
            alignItems: 'center'
        }}
    >
        {/* SDK-fix QA reference: outline of the 1920x1080 virtual canvas (px-sized).
        With the fix it fits inside the screen (edges flush on the limiting axis);
        without it, it is ~devicePixelRatio× too big and its right/bottom edges run
        off-screen (clipped, not visible). */}
        <UiEntity
            uiTransform={{
                positionType: 'absolute',
                position: { top: 0, left: 0 },
                width: VIRTUAL_WIDTH,
                height: VIRTUAL_HEIGHT,
                borderWidth: 4,
                borderColor: Color4.create(0, 1, 0, 1)
            }}
        />

        {/* Bordered image for Explorer-side QA (background-vs-border alignment). */}
        <UiEntity
            uiTransform={{ width: 500, height: 500, borderWidth: 6, borderColor: Color4.Yellow() }}
            uiBackground={{ texture: { src: IMAGE }, textureMode: 'stretch' }}
        />
    </UiEntity>
)
```

nicoecheza (approved):

DafGreco (approved):
✔️  PR reviewed and approved by QA on both platforms following instructions playing both happy and un-happy path

Regressions for this ticket had been performed in order to verify that the normal flow is working as expected:

Image
Image
Image

PR #1436, opened 2026-07-01
chore: explorer start multi-instance by default

pravusjif:
### WHY

When working on a local scene creators either:
- Don't care about multi-instance
- WANT multi-instance (to be able to test multiplayer functionalities)

It makes absolutely no sense to make a creator go out of their way to toggle it ON, losing time with that.

### WHAT

Made multi-instance param for the explorer deep link TRUE by default.


pravusjif:
```
When working on a local scene creators either:
* **Don't care about multi-instance**
```

Aparently that case is not 100% true since there have been cases of creators opening multiple instances without realizing, and having more than 1 explorer consuming battery and processing power Inadvertently.

PR #1447, opened 2026-07-09
chore: export unique avatar mask enum

pravusjif:
## Problem
`triggerSceneEmote({ mask })` (`~system/RestrictedActions`) takes an `AvatarMask`, but there was **no single `AvatarMask` a scene could use**:

- `@dcl/ecs` / `@dcl/sdk/ecs` did **not** export `AvatarMask` at all — `avatar_mask.proto` is an API-only common type, so the ecs codegen (which only runs protoc over *component* protos) never generated it.
- The js-runtime generator inlined its **own** `export enum AvatarMask` into `apis.d.ts`, nominally distinct from any other.

Result: importing `AvatarMask` from `~system/RestrictedActions` type-checked but was `undefined` at runtime (`TypeError: Cannot read properties of undefined (reading 'AM_UPPER_BODY')`).

## Fix
Make `@dcl/ecs` the single source of `AvatarMask` and have the API reference it.

- **`scripts/protocol-buffer-generation/generateProtocolBuffer.ts`** — add `common/avatar_mask.proto` to the ecs protoc compile list, so `AvatarMask` is generated into `@dcl/ecs` and re-exported (as a `const enum`) through `@dcl/sdk/ecs`.
- **`scripts/rpc-api-generation/index.ts`** — drop the inlined `AvatarMask` enum from `apis.d.ts` and rewrite the `mask` field to `import('@dcl/ecs').AvatarMask`. An *inline* import type is used deliberately: a top-level import would turn `apis.d.ts` into a module and break the ambient `declare module "~system/*"` declarations.

## Result
One `AvatarMask`, imported from `@dcl/sdk/ecs`, usable directly with `triggerSceneEmote` — no cast, real runtime value:

```ts
import { AvatarMask } from '@dcl/sdk/ecs'
import { triggerSceneEmote } from '~system/RestrictedActions'

triggerSceneEmote({ src, loop: true, mask: AvatarMask.AM_UPPER_BODY })

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,15 @@ export DEVELOPER_MODE=true
 install:
 	npm i --silent
 	make install-protobuf
+	make link-ecs-into-js-runtime
+
+# The generated packages/@dcl/js-runtime/apis.d.ts references `import('@dcl/ecs').AvatarMask`
+# (see scripts/rpc-api-generation). This repo has no root-hoisted node_modules, so make
+# @dcl/ecs resolvable from the js-runtime package dir for every build/test tsconfig.
+# @dcl/ecs has no dependencies, so a plain symlink is enough.
+link-ecs-into-js-runtime:
+	mkdir -p packages/@dcl/js-runtime/node_modules/@dcl
+	ln -sf ../../../ecs packages/@dcl/js-runtime/node_modules/@dcl/ecs
 
 update-protocol:
 	npm i --save-exact @dcl/protocol@next
--- a/packages/@dcl/playground-assets/etc/playground-assets.api.md
+++ b/packages/@dcl/playground-assets/etc/playground-assets.api.md
@@ -191,6 +191,12 @@ export const AvatarEquippedData: LastWriteWinElementSetComponentDefinition<PBAva
 export const AvatarLocomotionSettings: LastWriteWinElementSetComponentDefinition<PBAvatarLocomotionSettings>;
 
 // @public (undocumented)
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Review by nicoecheza — APPROVED** [review]

File: Makefile [hunk 1]
```diff
@@ -34,6 +34,15 @@ export DEVELOPER_MODE=true
...
 	make install-protobuf
+	make link-ecs-into-js-runtime
+
+# The generated packages/@dcl/js-runtime/apis.d.ts references `import('@dcl/ecs').AvatarMask`
+# (see scripts/rpc-api-generation). This repo has no root-hoisted node_modules, so make
+# @dcl/ecs resolvable from the js-runtime package dir for every build/test tsconfig.
+# @dcl/ecs has no dependencies, so a plain symlink is enough.
+link-ecs-into-js-runtime:
+	mkdir -p packages/@dcl/js-runtime/node_modules/@dcl
+	ln -sf ../../../ecs packages/@dcl/js-runtime/node_modules/@dcl/ecs
```
cyaiox, line 45: was this added for testing purposes, or is it required for building the package?

pravusjif: It's not test-only, but it doesn't affect the published packages or scenes.

pravusjif:
I'm trying to get a simpler better approach suggested by @cyaiox in a different PR.

pravusjif:
closing in favour of https://github.com/decentraland/js-sdk-toolchain/pull/1458

PR #1451, opened 2026-07-12
chore: warning-free, vulnerability-free install; jest 30 without ts-jest

eordano:
## What

`npm install` on this repo greeted every contributor with **22 vulnerabilities (2 critical, 10 high)**, 5 engine warnings, and (on a fresh resolve) 13 deprecation warnings. After this PR: **0 vulnerabilities, 0 engine warnings, and a single residual deprecation** that cannot be fixed from this repo (see below).

| | before | after |
|---|---|---|
| audit vulnerabilities | 22 (2 critical) | **0** |
| EBADENGINE warnings | 5 | **0** |
| deprecation warnings | 13 | **1** (`string-similarity`, see below) |

## How

- **Security**: `overrides` pin patched transitives — `protobufjs@^7.6.3` (critical prototype pollution, reached via `@dcl/protocol`/`ts-proto`), `semver`, `minimatch`, `zod`, `ajv`, `undici`, `test-exclude`; plus `npm audit fix` for the rest. Lockfile regenerated.
- **Engines**: `"node": ">=20.0.0"` instead of exactly `24.16.0` — the exact pin warned on every node that wasn't byte-identical (CI keeps 24.16.0 via `.node-version`).
- **Deprecations**: jest 30 (glob@7/inflight chain), glob 13 (only non-deprecated line), rimraf 6, syncpack 13, `@dcl/eslint-config` 3.
- **ts-jest removed**: tests compile through babel-jest (bundled with jest) + `@babel/preset-typescript`. Jest 30's removed matcher aliases are codemodded (`toBeCalledWith` → `toHaveBeenCalledWith` etc., 25 files). Type-only imports that ts-jest elided through type-checking are now explicit `import type` (they pulled raw `.ts` from `@dcl/protocol/out-ts` into the runtime under babel).
- **eslint 9**: `.eslintrc.json` + `.eslintignore` fold into `eslint.config.mjs` via FlatCompat (the shared config still publishes eslintrc shape).
- **Snapshots regenerated**: protobufjs 7.2.4→7.6.3 legitimately shifts bundle sizes and VM stats.

## Verification

Full suite collects **1097 tests (943 before** — suites that previously died at import now run). The only failures are the pre-existing ones that fail identically on `main`: the `init` blackbox network flake, the `help over all commands` timeout, and the ethereum-fixture cross-suite interference (each passes in isolation). `make build` green end to end, including proto generation.

## The one residual warning

`string-similarity@4.0.4` (deprecated, no successor) is a hard dependency of `eslint-plugin-autofix`, which is a hard dependency of `@dcl/eslint-config` — not overridable from here without breaking the plugin. Getting to literal zero needs a one-line fix in `@dcl/eslint-config` (drop or vendor the plugin's similarity helper); happy to send that PR too.


eordano:
Second commit: dependency replacement with node builtins. @dcl/sdk-commands — the tree every scene template installs — goes from 36 to 29 dependencies: uuid → crypto.randomUUID, glob → fs.globSync, colorette → util.styleText, portfinder → net, fp-future → Promise.withResolvers, open/prompts → small local modules over child_process and readline/promises; undici moves to the current major and remains only for the linker Agent (self-signed TLS), with the fetch component wrapping global fetch. Root drops rimraf (fs.rmSync), @actions/core (six-line GITHUB_STEP_SUMMARY append), buffer and mitt (unused). Fresh npm ci: 673 packages (was 685), still 0 vulnerabilities, full build green, suite failures unchanged from main (the two known flakes).

PR #1459, opened 2026-07-13
fix(sdk-commands): prevent preview path traversal

LautaroPetaccio:
## Why

The preview server accepted request-derived paths and used string-prefix checks to decide whether they belonged to a project directory. String prefixes do not represent filesystem boundaries: a sibling such as scene-secret matches scene, and resolved traversal segments can point outside the directory being served. This could allow preview HTTP routes to read files beyond their intended project or static-content roots.

## What changed

- Added a shared resolvePathInside helper based on path.resolve and path.relative.
- The helper only returns paths whose relative form remains within the configured root.
- Applied the containment check to generated b64 content paths and static preview routes before any filesystem lookup or stream is created.
- Added coverage for valid nested paths, parent-directory traversal, and absolute sibling paths that merely share the same prefix.
- Added Node types to the test TypeScript configuration because the regression suite uses the Node path API.

## Verification

- Focused path-containment Jest suite passes.
- The sdk-commands TypeScript check passes.
- make lint passes.

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/start/server/endpoints.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/server/endpoints.ts
@@ -8,6 +8,7 @@ import { v4 as uuidv4 } from 'uuid'
 import { PreviewComponents } from '../types'
 import { fetchEntityByPointer } from '../../../logic/catalyst-requests'
 import { CliComponents } from '../../../components'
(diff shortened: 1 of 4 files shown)


nicoecheza (changes requested):
maybe we can move this to `packages/@dcl/sdk-commands/src/logic/workspace-validations.ts`?

PR #1460, opened 2026-07-13
fix(ecs): preserve ByteBuffer offsets during growth

LautaroPetaccio:
## Why

`ReadWriteByteBuffer` can be initialized from a `Uint8Array` slice whose `byteOffset` is non-zero. When the buffer grew, it allocated a new standalone `Uint8Array` but reused the old slice offset when creating its `DataView`. The new allocation starts at offset zero, so retaining the old offset made subsequent reads and writes address the wrong region and could also create an invalid view.

The constructor had the mirror problem: it built the `DataView` from the backing buffer's `byteOffset` with no length, so the view spanned to the end of the underlying `ArrayBuffer` rather than stopping at the buffer's own region. Over a subarray that let the unchecked `set*`/`get*` proxies address memory belonging to whatever else shares the `ArrayBuffer`.

## Reachability: latent, not live

Worth stating so reviewers can size this correctly: **no current code path can trigger it, and a scene cannot reach it through public API.**

Triggering needs a buffer built over a non-zero-offset slice *and* that buffer growing, and growth only happens on write via `#woAdd`. Across the four scene-runtime packages there are 23 `ReadWriteByteBuffer` constructions: 16 start empty (`byteOffset` 0 — all the write paths), and the 7 that wrap an existing buffer are all read-only deserialization (`network/utils.ts:77`, `composite/instance.ts:65`, `parseChunkMessage`, `lww-element-set-…:136`, `grow-only-value-set-…:156`, `logger.ts:6` and `:31`). The two sets don't intersect.

The precondition is nonetheless everywhere: `msg.data` comes from `readBuffer()` (a `subarray`), and the CRDT path builds `messageBuffer: buffer.buffer().subarray(...)`. Those slices are only ever read for `.byteLength` or passed as *source* data to `writeBuffer(...)`, never used as a backing buffer. One refactor — appending to a received buffer, or wrapping a `messageBuffer` in a writable one — makes this reachable, and the failure mode is silent CRDT data loss with no exception and no log. Reproduced that way locally: two `uint32`s written after growth came back as `0x0`, and both round-trip with this change.

`ReadWriteByteBuffer` is not exported from `@dcl/ecs` or `@dcl/sdk/ecs`, so reaching it from a scene requires deep-importing `@dcl/ecs/dist/serialization/ByteBuffer` (which is how `logger.ts` uses it internally).

## What changed

- Rebuild the `DataView` from the newly allocated buffer's own `byteOffset` and `byteLength` after growth.
- Stop carrying the source slice offset into the replacement allocation.
- Pass `byteLength` in the constructor too, so both views are scoped identically.
- Added two tests in `test/ecs/byteBuffer.spec.ts`: one starting from a sliced backing buffer, forcing growth and verifying the next value lands at its logical offset; one pinning the constructor's bound, where a raw proxy write aimed outside the slice must throw `RangeError` and leave the neighbouring byte untouched.

`byteBuffer.spec.ts` already had `'should create a buffer with subArray offset'` and two `'... after a grow'` tests next to each other, and they missed this because each covers one axis alone: the grow tests start from a standalone array whose `byteOffset` is 0, so a stale offset of 0 is indistinguishable from a correct one, and the subarray test only reads, so it never grows. The bug needed both at once. Both new tests are mutation-checked — reverting either half of the fix fails them.

## Behavior change

Callers of the raw `set*`/`get*` proxies with an offset outside the buffer's own region now get a `RangeError` instead of silently reading or writing out of bounds. Nothing in this repo is affected: every write path either reserves capacity through `#woAdd` first or goes through `_buffer.set`/`utf8.write`, which were already `Uint8Array`-bounded. The reserve-then-back-patch callers in `serialization/crdt/*` and `components/manual/Transform.ts` only write within the range they reserved.

Previously the two paths disagreed: the same out-of-range call threw once the buffer had grown, and corrupted neighbouring memory before it had.

## Verification

- `node_modules/.bin/jest --forceExit --runInBand test/ecs test/sdk test/react-ecs` — 1161 passed
- `node_modules/.bin/jest --forceExit --runInBand test/snapshots.spec.ts` — 17 passed, after regenerating the golden files (only `MALLOC_COUNT` and `MEMORY_USAGE_COUNT` move, each down slightly)
- `npx tsc --noEmit -p tsconfig.json` in `packages/@dcl/ecs`
- `make lint`


packages/@dcl/ecs/src/serialization/ByteBuffer/index.ts
nicoecheza, line 38: can we add `byteLength` here too?

nicoecheza (approved):

packages/@dcl/ecs/src/serialization/ByteBuffer/index.ts
LautaroPetaccio: Good catch — done in 9b766264.

It turned out to be more than symmetry. Without the length the constructor's view spans to the end of the underlying `ArrayBuffer`, so for a buffer built over a subarray the unchecked `set*`/`get*` proxies could address memory outside the slice. A `ByteBuffer` over `shared.subarray(8, 16)` plus a raw `setUint32(24, ...)` wrote straight into byte 32 of the shared buffer — silently clobbering whatever else lives there. With `byteLength` it's a `RangeError` instead.

It also removed an inconsistency this PR would otherwise have introduced: the growth path passes `byteLength`, so the same out-of-range call threw once the buffer had grown but corrupted memory before it had.

PR #1463, opened 2026-07-13
fix(sdk): await network profile initialization

LautaroPetaccio:
## Why

Network setup loaded the local profile as an unawaited side effect. Entity synchronization and transport ticks could run before userId and networkId were populated, causing profile-not-initialized errors or filtering messages with undefined identity data. Failures from getUserData were also thrown inside an unobserved promise chain, producing unhandled rejections without a readiness signal for callers.

## What changed

- Make fetchProfile await getUserData and return a Promise that rejects with a descriptive error when no user ID is available.
- Add a ready promise to the network transport initialization result and export it as networkReady from the public network module.
- Queue syncEntity, parentEntity, and removeParent operations until profile initialization completes.
- Await readiness before the transport sends or processes network responses.
- Attach handling for module-level initialization failures while preserving rejection on ready for callers that explicitly await it.
- Added regression coverage for an early syncEntity call and for an identity response without a user ID.

## Verification

- The profile-initialization regression suite passes.
- Existing network transport, engine synchronization, and CRDT state suites pass.
- The sdk TypeScript check passes.
- make lint passes.

### DIFF
--- a/packages/@dcl/sdk/src/network/index.ts
+++ b/packages/@dcl/sdk/src/network/index.ts
@@ -4,7 +4,26 @@ import { addSyncTransport } from './message-bus-sync'
 import { getUserData } from '~system/UserIdentity'
 
 // initialize sync transport for sdk engine
-const { getChildren, syncEntity, parentEntity, getParent, myProfile, removeParent, getFirstChild, isStateSyncronized } =
-  addSyncTransport(engine, sendBinary, getUserData)
+const {
+  getChildren,
+  syncEntity,
+  parentEntity,
+  getParent,
+  myProfile,
+  removeParent,
+  getFirstChild,
+  isStateSyncronized,
+  ready: networkReady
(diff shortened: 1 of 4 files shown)


LautaroPetaccio:
Addressed the two correctness-related P2 notes in ea996824:

- Queued operations are now isolated so one failure is reported without discarding later operations.
- Profile initialization has an explicit failed state, and transport ticks return without sending instead of rejecting repeatedly after identity loading fails.
- Operations invoked after initialization has already failed are no longer retained indefinitely.

I kept the explicit queue rather than replacing it with ready.then wrappers because it preserves synchronous execution for calls made after initialization, as noted in the review.

Added regression coverage for queue continuation, successful engine ticks after identity failure, and suppression of network sends. The focused and existing network suites pass (25 tests), the SDK build/typecheck passes, and the SDK package has no lint issues.

PR #1469, opened 2026-07-13
fix: cache tween systems by engine identity

LautaroPetaccio:
## Why

Tween systems were stored in a process-wide `Map` keyed by `engine._id`, while engine IDs are generated with `Date.now()`. Engines created in the same millisecond could therefore receive the same tween system even though its components and systems belonged to only one engine. The strong global map also retained every engine-specific tween system for the lifetime of the process.

## What changed

- Replace the numeric-ID map with a `WeakMap` keyed by the engine object itself.
- Continue returning a single tween system for repeated requests from the same engine.
- Keep systems for distinct engines isolated even if their numeric IDs collide.
- Allow unused engines and their tween systems to be garbage-collected.
- Add explicit collision and same-engine cache regression tests.

## Verification

- `node_modules/.bin/jest --colors --forceExit --runInBand --testPathPattern=test/ecs/events/tween(-cache)?.spec.ts`
- `npx tsc --noEmit -p tsconfig.json` in `packages/@dcl/ecs`


eordano (changes requested):
This seems to break scenes in my local testing

PR #1475, opened 2026-07-13
perf: index CRDT network entities

LautaroPetaccio:
## Why

Resolving an inbound network CRDT message to its local entity scanned every `NetworkEntity` component. That lookup runs per received message, so the update path cost `messages x synchronized entities` — quadratic in scene population for a busy multiplayer scene.

## What changed

- Keep a per-engine map keyed by `(networkId, entityId)` and resolve in constant time.
- Track the reverse local-entity key, so a replacement or deletion drops only its own mapping.
- Maintain the index on every path that can change it: locally emitted `NetworkEntity` changes, newly received mappings, accepted remote updates, component deletion, and entity cleanup.

## Measured

Inbound messages targeting the last synced entity — worst case for the scan, and the realistic one when a scene addresses whichever peer just moved:

| synced entities | messages/frame | main | this branch |
| --- | --- | --- | --- |
| 100 | 20 | 1.12 ms/frame | 0.27 ms/frame |
| 500 | 50 | 7.82 ms/frame | 0.34 ms/frame |
| 1 000 | 100 | **31.81 ms/frame** | **0.49 ms/frame** |

The scaling confirms the cost model rather than a tuned constant: quadrupling the work (500→1000 entities, 50→100 messages) takes `main` from 7.8 to 31.8 ms, almost exactly 4x, while this branch goes 0.34 to 0.49 ms. At the top row `main` spends about two 60 fps frames per frame on entity resolution alone.

## Tests

`test/ecs/crdt-network-entity-index.spec.ts` — moved from `test/ecs/network/`, which was a new directory holding one file; `test/ecs` already keeps `crdt-rules-lww`, `crdt-rules-gset`, `crdtMessage` and `ecs.crdt` flat, and `test/sdk/network/` covers the SDK-side transport layer rather than this system.

An index is only as good as its invalidation, so the two original happy-path tests are joined by three that exercise it, each mutation-checked:

- **re-pointing** a mapped entity at a different remote entity — the old key must stop resolving
- **deleting** `NetworkEntity` from a mapped entity — same
- **ownership transfer**: a second local entity claims the same remote mapping, then the first is removed; the mapping must survive, which is what the `localEntityId` guard in `removeIndexedNetworkEntity` exists for

Verified by mutation: dropping the stale-mapping removal fails 2, dropping the ownership guard fails 1, and dropping the local-change indexing fails 2. Only that last one was caught by the original pair.

## Scene-level test

[`crdt-network-entity-index`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-network-entity-index) — 400 entities registered with `syncEntity`, 40 mutated per frame, measuring the per-frame cost of the SDK's systems while checking peer entities still arrive and track. Resolution only runs for inbound messages, so it needs two clients (`npm start -- --multi-instance`); with one it reports `SOLO` rather than a number that measures nothing.

## Verification

- `jest test/ecs test/sdk test/react-ecs` — 1175 passed
- `jest test/snapshots.spec.ts` — 17 passed after regeneration
- `make build`, `make lint`

`Generated EngineInfo ProtoBuf › should serialize/deserialize EngineInfo` fails here and on a clean `main` checkout, from the protocol bump in ffb26183. Unrelated to this change.

### DIFF
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -50,6 +50,35 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
   const receivedMessages: ReceiveMessage[] = []
   // Messages already processed by the engine but that we need to broadcast to other transports.
   const broadcastMessages: ReceiveMessage[] = []
+  const networkEntityIndex = new Map<string, { localEntityId: Entity; network: INetowrkEntityType }>()
+  const networkEntityKeyByLocalEntity = new Map<Entity, string>()
+
+  function networkEntityKey(networkId: number, entityId: Entity): string {
+    return `${networkId}:${entityId}`
+  }
+
+  function removeIndexedNetworkEntity(localEntityId: Entity): void {
+    const previousKey = networkEntityKeyByLocalEntity.get(localEntityId)
+    if (!previousKey) return
+
+    if (networkEntityIndex.get(previousKey)?.localEntityId === localEntityId) {
+      networkEntityIndex.delete(previousKey)
+    }
+    networkEntityKeyByLocalEntity.delete(localEntityId)
+  }
+
+  function indexNetworkEntity(localEntityId: Entity, network?: INetowrkEntityType): void {
+    removeIndexedNetworkEntity(localEntityId)
+    if (!network) return
+
+    const key = networkEntityKey(network.networkId, network.entityId)
+    const indexedNetwork = {
+      localEntityId,
+      network: { networkId: network.networkId, entityId: network.entityId }
+    }
+    networkEntityIndex.set(key, indexedNetwork)
+    networkEntityKeyByLocalEntity.set(localEntityId, key)
+  }
 
   /**
    *
@@ -120,10 +149,9 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
     const hasNetworkId = 'networkId' in msg
 
     if (hasNetworkId) {
-      for (const [entityId, network] of engine.getEntitiesWith(NetworkEntity)) {
-        if (network.networkId === msg.networkId && network.entityId === msg.entityId) {
-          return { entityId, network }
-        }
+      const indexedNetwork = networkEntityIndex.get(networkEntityKey(msg.networkId!, msg.entityId))
+      if (indexedNetwork) {
+        return { entityId: indexedNetwork.localEntityId, network: indexedNetwork.network }
       }
     }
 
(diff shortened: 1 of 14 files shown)


LautaroPetaccio:
Checked against `auth-server`: this patch can't apply — `findNetworkId` was deleted from the engine along with the whole network-entity subsystem. The identical linear scan does still exist, relocated to `findExistingNetworkEntity` in `packages/@dcl/sdk/src/network/server/index.ts`, now running on the server for every peer.

**I did not port it, on measurement.** I built the index there and benchmarked it against the scan through `processServerMessages`:

| N entities / N messages | scan | index |
| --- | --- | --- |
| 1,500 | 37 ms | 47 ms |
| 3,000 | 90 ms | 73 ms |

Noise, and slower at the smaller size. Unlike the engine's receive path, the scan isn't the bottleneck on the server — `validateMessagePermissions` does a full `deserialize` plus a dry-run CRDT per message and dominates. A cache with staleness handling isn't worth adding for no measurable win.

Worth re-measuring if scenes grow much past a few thousand synced entities, or if that per-message validation gets cheaper.

LautaroPetaccio:
Closing: the code this patches does not exist on `auth-server`.

**What is superseded.** `findNetworkId` was deleted from the engine along with the whole network-entity subsystem, so there is no scan left on the engine's receive path to index. The equivalent lookup now lives in `findExistingNetworkEntity` in `packages/@dcl/sdk/src/network/server/index.ts`, and **I measured it there rather than porting on assumption**:

| N entities / N messages | scan | with an index |
| --- | --- | --- |
| 1,500 | 37 ms | 47 ms |
| 3,000 | 90 ms | 73 ms |

Noise, and slower at the smaller size. Unlike the engine's receive path, the scan is not the bottleneck on the server — `validateMessagePermissions` runs a full `deserialize` plus a dry-run CRDT per message and dominates. A cache with staleness handling is not worth adding there for no measurable win, so there is genuinely nothing to port.

**What closing costs on `main`, stated plainly.** The problem this fixes is real and measured there — 31.81 ms/frame at 1,000 synced entities and 100 messages/frame, against 0.49 ms with the index. It is also a security mitigation, not only a performance one: the quadratic resolution is what makes entity flooding cheap. On `main` a single inbound message costs ~27 ms once 8,000 entities exist, so roughly 2,000 injected entities are enough to make a scene unplayable. With constant-time resolution an attacker has to actually exhaust the range instead of getting an amplifier for free.

So closing is right if `auth-server` is what ships. If `main` continues to ship first, this cost stays on it, and the entity-flooding issue is meaningfully worse there without this change.

PR #1479, opened 2026-07-13
perf: index entity tree children instead of rescanning

LautaroPetaccio:
## Why

Entity-tree traversal scanned every remaining component entry once per visited entity, so removing a subtree cost `O(k x n)` — k entities in the subtree, n entities carrying the parenting component. The recursive implementation could also overflow the call stack on deep trees, and network-parent removal repeated the full scan for every descendant.

## What changed

- Build a parent-to-children index in one pass over the parenting component, then traverse it iteratively. Cost becomes `O(n + k)`.
- `getComponentEntityTree` keeps child-first order, sibling order and cycle protection.
- `removeNetworkEntityChildrens` builds its index once and removes **parent-first**, unlike the Transform tree. That difference is load-bearing: `test/sdk/network/sync-engines.spec.ts` asserts the resulting message order, so this one is a pre-order walk while `getComponentEntityTree` stays child-first.

Note for anyone verifying locally: `sync-engines.spec.ts` resolves `@dcl/ecs` through the built package, so it needs `make build` after switching branches — without a rebuild it silently exercises the previously built code.

## Measured

Cost goes from `O(k x n)` to `O(n + k)`, so the win scales with how much of the scene the removed subtree covers — in a 10 000-entity scene:

| Subtree removed | main | this branch |
| --- | --- | --- |
| 3 | 7.3 ms | 3.3 ms |
| 2 000 | 194.6 ms | 3.0 ms |
| 9 999 | 544.4 ms | 3.5 ms |

That it is a complexity change and not a tuned constant shows in the scaling: as N doubles, `main` costs x3.98 while this branch costs x1.33.

Both versions keep an `O(n)` floor, so removing a single leaf still walks every entity carrying the component (~3 ms at n = 10 000). The deep-tree case is a correctness win rather than a speed one — `main` cannot complete the 12 001-entity test at all.

## Tests

Coverage for `tree.ts` traversal is consolidated in `test/ecs/entity-tree-traversal.spec.ts`, including the three tree tests that previously sat in the 1272-line `engine.spec.ts`, restructured with every assertion preserved. `getEntitiesWithParent` stays there, being a direct-children lookup this PR does not touch.

Mutation testing drove the coverage, since several guards had nothing defending them:

- **sibling order** — flipping the push order left all 752 tests green: a 12k chain has one child per node, and the multi-child test in `engine.spec.ts` asserted with order-insensitive `arrayContaining`.
- **entity 0 as a parent key** — weakening `parent !== undefined` to `parent` left everything green, though `Schemas.Entity` defaults to `0` and the recursive `value.parent === entity` comparison indexed it.
- **cycle protection in `removeNetworkEntityChildrens`** — dropping its `visited` guard left `sync-engines` at 17/17, and without it a cyclic `NetworkParent` graph never terminates. That function is `/* istanbul ignore next */`, so coverage says nothing about it either way; the new test reaches it through the exported helper, because `engine.removeEntityWithChildren` passes its own closure-scoped `removeEntity`.

The multi-child, cyclic and entity-0 expectations were verified against `main`'s implementation as well, so they pin preserved behavior rather than this implementation's.


LautaroPetaccio:
Closing it due to small performance improvements.

PR #1497, opened 2026-07-22
feat: storage cache, unchanged-write skipping, memoized storage URL

pravusjif:
## Problem

Every `Storage` operation was a signed network round-trip. Authoritative-server scenes read and flush state on tick loops, so the storage service absorbed a `GET` per read and a `PUT` per write even when nothing changed, and `getStorageServerUrl()` resolved the realm on every single operation.

Separately, `wrapSignedFetch` — the transport under every `Storage`/`EnvVar` call — could not survive a non-JSON response body: a storage server returning an HTML error page or a truncated body crashed the caller instead of degrading to the documented `null`/`false` results.

## Changes

`Storage` and `Storage.player` now share a per-scope value cache (TTL'd, bounded, in-memory) that serves reads and suppresses no-op writes. The authoritative server is the sole writer of its keys (`signedFetch` identity is host-selected; write authority is enforced by the storage service), so the cache is coherent within its TTL.

### Read caching (default on)

`get()` serves values known from a network round-trip within the last `cacheMaxAgeMs` without hitting the service:

- Populated by successful `get()`s, `set()` write-through, and `getValues()` pages (returned entries seed only keys with no fresher per-key state — a live cache entry or a pending write always wins over the page snapshot, whose request started earlier; the list request itself is never served from cache).
- Cached bodies are re-parsed per hit, so callers never share mutable object references.
- **Negative caching**: a confirmed `404` (or a successful `delete()`) caches "absent", so hot-looping on a missing key doesn't hammer the service. Only `status === 404` qualifies — transport errors and 5xx are never cached. `get()` no longer logs `console.error` for 404s (it's a first-class "not found" outcome now).
- **In-flight coalescing**: concurrent `get()`s for the same key share one request (player keys coalesce case-insensitively per address, never across addresses). Active even with read caching disabled — a live request is never stale.
- **Race guard**: a `set()`/`delete()` completing while a `GET` is in flight detaches it, so the stale response can't overwrite the newer cache entry.

Escape hatches: `get(key, { fresh: true })` forces an authoritative network read (and refreshes the cache); `Storage.configure({ cacheReads: false })` disables read serving module-wide.

### `skipIfUnchanged` (default on)

`set()` skips the `PUT` (resolving `true`) when the serialized value matches the last value confirmed stored for that key. A skip only suppresses a **proven no-op** — the new serialized body is compared for exact equality against the body confirmed by a previous successful round-trip within the TTL — so writes are never lost: every `set()` resolves once its value (or the newer value that superseded it, see write serialization below) is durably applied. A dedup skip never refreshes the entry's TTL, so a constantly-rewritten value still re-asserts itself with one real `PUT` per TTL window. A failed `PUT` invalidates the cache entry (the write may have reached the server). Opt out per call (`{ skipIfUnchanged: false }`) or via `configure()`.

### Per-key write serialization

Overlapping writes to the same key are serialized with latest-value coalescing: at most one `PUT`/`DELETE` per key is on the network, and writes issued meanwhile collapse into a single queued "latest value" slot. Without this, two fire-and-forget `set()`s from consecutive ticks race twice — the service keeps whichever request it processes last (it's last-write-wins with no version tokens), and the cache keeps whichever response arrives last — so either could disagree with the last `set()` issued, and a wrong cache entry would then be served (and defended by `skipIfUnchanged`) for up to the TTL. With serialization, the server's final state always matches the last write issued, and N rapid writes to a key cost at most 2 network calls — a further win for the tick-loop workload. Semantics:

- Issue order is commit order, per key (`delete()` participates in the same queue).
- A `set()` identical to the pending write joins it (no extra `PUT`) — unless the caller passed `skipIfUnchanged: false`, which always issues its own write.
- Superseded writers resolve with the outcome of the write that replaced theirs (their value was overwritten before it could ever be observed).
- Write dedup against the cache only applies while no write is pending for the key (the cache is momentarily behind the pending write).

### Cache mechanics & configuration

- Entries hold the serialized `{ value }` body, which serves both read hits (re-parsed per hit) and write dedup (exact string comparison). `cacheMaxEntries` (default 512 per scope) bounds entry **count**, not bytes — a large value stays pinned for up to the TTL.
- `cacheMaxAgeMs` default is **1 minute**, bounding both read staleness and write-dedup trust.
- Staleness caveat: out-of-band writers (e.g. the CLI `storage` commands against a live world) may not be visible for up to the TTL — use `{ fresh: true }` where that matters.
- Player cache keys are address-lowercased, so mixed checksum casing shares entries (and the write queue) correctly.
- Misconfiguration guards: non-finite `cacheMaxEntries`/`cacheMaxAgeMs` fall back to the defaults, negative bounds degrade safely, and `configure()` ignores explicitly-`undefined` options (e.g. from a forwarded partial options object), keeping the configured defaults.

New exports from `@dcl/sdk/server`: `GetOptions`, `SetOptions`, `StorageOptions`.

### Memoized `getStorageServerUrl()`

The resolved URL promise is cached: concurrent first calls collapse into a single `getRealm` request, later operations skip it entirely, and failed resolutions are not cached (transient errors retry). `EnvVar` benefits automatically.

### Fix: `wrapSignedFetch` rejects on malformed response bodies

`wrapSignedFetch` parsed responses with `tryCatch(JSON.parse(response.body || '{}'))`. `JSON.parse` throws synchronously while the argument is evaluated — before `tryCatch` can wrap it — so an invalid body rejected the whole `wrapSignedFetch` promise and the `'Failed to parse response'` branch was dead code. Every `Storage.get/set/delete` and `EnvVar.get` then threw instead of returning `null`/`false` as designed. The parse now runs inside a plain try/catch, restoring the intended error tuple (`['Failed to parse response', null, status]`).

## Behavior changes to note

- `get()` may serve cached data by default (bounded by the 1-minute TTL; `{ fresh: true }` / `cacheReads: false` opt out).
- `set()` skips proven-unchanged writes by default (`skipIfUnchanged: false` opts out).
- `get()` no longer logs `console.error` on 404.

## Testing

91 tests across the server suites. Read caching: cache hits (with per-hit object freshness), TTL expiry, `fresh` bypass-and-refresh, coalescing (including `fresh` joining in-flight and per-address isolation), negative 404 caching, non-404 errors never cached, set/delete invalidation and overwrite of negative entries, `getValues` seeding (reads + write dedup, scoped per address, plus the race guard: pages never overwrite state confirmed or pending while the list was in flight), the stale-GET-vs-set race guard, ambiguous 200s not cached, stored `null` as a positive entry, and `configure({ cacheReads: false })` reaching both scopes. Write dedup: default-on skip, per-call/configured overrides, failed-PUT retry, max-age expiry, per-address isolation, case-insensitive addresses. Write serialization: overlapping sets hold until the in-flight PUT lands and preserve issue order, rapid writes coalesce to the latest value in ≤2 PUTs, identical concurrent sets join one PUT (except with `skipIfUnchanged: false`), deletes serialize behind sets, failed coalesced flushes fail their writers and invalidate the entry, and case-insensitive per-address queueing. Config guards: non-finite `cacheMaxEntries`/`cacheMaxAgeMs` fallbacks, negative bounds, `configure()` ignoring explicit `undefined`. Plus cache-primitive tests (eviction, TTL, negative entries), URL memoization (sequential, concurrent, failure-retry), and `wrapSignedFetch` coverage (success, empty body, malformed-JSON regression, non-ok statuses, fetch rejection).

### DIFF
--- a/packages/@dcl/sdk/src/server/storage/index.ts
+++ b/packages/@dcl/sdk/src/server/storage/index.ts
@@ -1,8 +1,9 @@
+import { createStorageConfig, StorageOptions } from './constants'
 import { createSceneStorage, ISceneStorage } from './scene'
 import { createPlayerStorage, IPlayerStorage } from './player'
 
 // Re-export interfaces and types
-export { GetValuesOptions, GetValuesResult } from './constants'
+export { GetValuesOptions, GetValuesResult, SetOptions, StorageOptions } from './constants'
 export { ISceneStorage } from './scene'
 export { IPlayerStorage } from './player'
 
@@ -12,14 +13,22 @@ export { IPlayerStorage } from './player'
 export interface IStorage extends ISceneStorage {
   /** Player-scoped storage for key-value pairs */
   player: IPlayerStorage
+
+  /**
+   * Sets module-wide defaults for both scene-scoped and player-scoped storage.
+   * Merges the given partial options into the current configuration.
+   * Note: cacheMaxEntries applies per scope (scene and player each keep their own cache).
+   */
+  configure(options: StorageOptions): void
 }
 
 /**
  * Creates the Storage module with scene-scoped and player-scoped storage.
  */
 const createStorage = (): IStorage => {
-  const sceneStorage = createSceneStorage()
-  const playerStorage = createPlayerStorage()
+  const config = createStorageConfig()
+  const sceneStorage = createSceneStorage(config)
+  const playerStorage = createPlayerStorage(config)
 
   return {
     // Spread scene storage methods at top level
@@ -28,7 +37,10 @@ const createStorage = (): IStorage => {
     delete: sceneStorage.delete,
     getValues: sceneStorage.getValues,
     // Keep player as nested property
-    player: playerStorage
+    player: playerStorage,
+    configure(options: StorageOptions): void {
+      Object.assign(config, options)
+    }
   }
 }
 
@@ -38,6 +50,7 @@ const createStorage = (): IStorage => {
  *
  * - Use Storage.get/set/delete/getValues for scene-scoped storage
  * - Use Storage.player.get/set/delete/getValues for player-scoped storage
+ * - Use Storage.configure to change module-wide defaults (e.g. skipIfUnchanged)
  *
  * This module only works when running on server-side scenes.
  */
--- a/packages/@dcl/sdk/src/server/storage/index.ts
+++ b/packages/@dcl/sdk/src/server/storage/index.ts
@@ -3,7 +3,7 @@ import { createSceneStorage, ISceneStorage } from './scene'
 import { createPlayerStorage, IPlayerStorage } from './player'
 
 // Re-export interfaces and types
-export { GetValuesOptions, GetValuesResult, SetOptions, StorageOptions } from './constants'
+export { GetOptions, GetValuesOptions, GetValuesResult, SetOptions, StorageOptions } from './constants'
 export { ISceneStorage } from './scene'
 export { IPlayerStorage } from './player'
 
@@ -17,7 +17,9 @@ export interface IStorage extends ISceneStorage {
   /**
    * Sets module-wide defaults for both scene-scoped and player-scoped storage.
    * Merges the given partial options into the current configuration.
-   * Note: cacheMaxEntries applies per scope (scene and player each keep their own cache).
+   * Note: cacheMaxEntries applies per scope (scene and player each keep their
+   * own cache) and bounds entry count, not bytes (entries hold serialized
+   * value bodies).
    */
   configure(options: StorageOptions): void
 }
@@ -50,7 +52,18 @@ const createStorage = (): IStorage => {
  *
  * - Use Storage.get/set/delete/getValues for scene-scoped storage
  * - Use Storage.player.get/set/delete/getValues for player-scoped storage
- * - Use Storage.configure to change module-wide defaults (e.g. skipIfUnchanged)
+ * - Use Storage.configure to change module-wide defaults (e.g. skipIfUnchanged, cacheReads)
+ *
+ * Reads are cached by default: get() serves values known from a network
+ * round-trip within the last cacheMaxAgeMs (including confirmed "not found"
+ * results) without hitting the service, and concurrent gets for the same key
+ * share one request. Writes are never deferred — set()/delete() always reach
+ * the service; by default (skipIfUnchanged) an unchanged set is skipped, but
+ * only when a previous confirmed round-trip proved the exact value is already
+ * stored. Out-of-band writers (e.g. CLI storage commands)
+ * may not be visible for up to cacheMaxAgeMs; use get(key, { fresh: true })
+ * for an authoritative read or Storage.configure({ cacheReads: false }) to
+ * disable read caching.
  *
  * This module only works when running on server-side scenes.
  */
--- a/packages/@dcl/sdk/src/server/storage/index.ts
+++ b/packages/@dcl/sdk/src/server/storage/index.ts
@@ -41,7 +41,9 @@ const createStorage = (): IStorage => {
     // Keep player as nested property
     player: playerStorage,
     configure(options: StorageOptions): void {
-      Object.assign(config, options)
+      for (const [key, value] of Object.entries(options)) {
+        if (value !== undefined) (config as unknown as Record<string, unknown>)[key] = value
+      }
     }
   }
 }
@@ -57,10 +59,12 @@ const createStorage = (): IStorage => {
  * Reads are cached by default: get() serves values known from a network
  * round-trip within the last cacheMaxAgeMs (including confirmed "not found"
  * results) without hitting the service, and concurrent gets for the same key
- * share one request. Writes are never deferred — set()/delete() always reach
- * the service; by default (skipIfUnchanged) an unchanged set is skipped, but
- * only when a previous confirmed round-trip proved the exact value is already
- * stored. Out-of-band writers (e.g. CLI storage commands)
+ * share one request. Overlapping writes to the same key are serialized so the
+ * service commits them in issue order, and rapid writes coalesce to the
+ * latest value; every set()/delete() resolves once its value (or the newer
+ * value that superseded it) is durably applied. By default (skipIfUnchanged)
+ * an unchanged set is skipped, but only when a previous confirmed round-trip
+ * proved the exact value is already stored. Out-of-band writers (e.g. CLI storage commands)
  * may not be visible for up to cacheMaxAgeMs; use get(key, { fresh: true })
  * for an authoritative read or Storage.configure({ cacheReads: false }) to
  * disable read caching.
--- a/packages/@dcl/sdk/src/server/storage/scene.ts
+++ b/packages/@dcl/sdk/src/server/storage/scene.ts
@@ -1,6 +1,14 @@
 import { getStorageServerUrl } from '../storage-url'
 import { assertIsServer, wrapSignedFetch } from '../utils'
-import { GetValuesOptions, GetValuesResult, MODULE_NAME } from './constants'
+import {
+  createStorageConfig,
+  GetValuesOptions,
+  GetValuesResult,
+  MODULE_NAME,
+  SetOptions,
+  StorageConfigState
+} from './constants'
+import { createValueCache, fingerprint } from './value-cache'
 
 /**
  * Scene-scoped storage interface for key-value pairs from the Server Side Storage service.
@@ -18,8 +26,9 @@ export interface ISceneStorage {
    * Stores a value in scene storage in the Server Side Storage service.
    * @param key - The key to store the value under
    * @param value - The value to store (will be JSON serialized)
+   * @param options - Optional { skipIfUnchanged } to skip the network write when the value is already stored
    */
-  set<T = unknown>(key: string, value: T): Promise<boolean>
+  set<T = unknown>(key: string, value: T, options?: SetOptions): Promise<boolean>
 
   /**
    * Deletes a value from scene storage in the Server Side Storage service.
@@ -42,7 +51,9 @@ export interface ISceneStorage {
  * scene-specific key-value pairs from the Server Side Storage service.
  * This module only works when running on server-side scenes.
  */
-export const createSceneStorage = (): ISceneStorage => {
+export const createSceneStorage = (config: StorageConfigState = createStorageConfig()): ISceneStorage => {
+  const cache = createValueCache(config)
+
   return {
     async get<T = unknown>(key: string): Promise<T | null> {
       assertIsServer(MODULE_NAME)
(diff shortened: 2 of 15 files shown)


packages/@dcl/sdk/src/server/storage/scene.ts
```diff
@@ -42,27 +59,80 @@ export interface ISceneStorage {
...
+  const cache = createValueCache(config)
+  // Each in-flight GET is tracked by a wrapper object whose identity marks
+  // ownership: set()/delete() drop the wrapper, detaching the pending GET so
+  // its stale response cannot overwrite the newer cache entry.
+  const inflightGets = new Map<string, { promise: Promise<unknown> }>()
+
   return {
-    async get<T = unknown>(key: string): Promise<T | null> {
+    async get<T = unknown>(key: string, options?: GetOptions): Promise<T | null> {
       assertIsServer(MODULE_NAME)
```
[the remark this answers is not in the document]
pravusjif: for now it favours readability in the current way.

packages/@dcl/sdk/src/server/storage/value-cache.ts
```diff
@@ -0,0 +1,96 @@
...
+    entries.delete(key)
+    entries.set(key, { ...entry, storedAt: Date.now() })
+
+    while (entries.size > config.cacheMaxEntries) {
+      entries.delete(entries.keys().next().value!)
+    }
+  }
+
+  return {
+    get(key: string): CacheEntry | undefined {
```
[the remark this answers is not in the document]
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/e6b466f3bcea8980689ea0bd57c9e7f440dde232

regenesis-claw (approved):
## Security-focused Review

Thorough, well-engineered PR. The caching layer is coherent, the race guards are sound, and the `wrapSignedFetch` fix addresses a real bug. 77 tests with excellent edge-case coverage.

### Security Analysis ✅

- **No injection vectors** — all URL parameters properly `encodeURIComponent`'d, values JSON-serialized
- **Player isolation** — per-address cache keys use lowercased addresses with NUL separator (`\u0000`), preventing cross-address cache poisoning and key collision
- **No unbounded growth** — cache bounded by `cacheMaxEntries` (default 512) with LRU-style eviction; the negative `cacheMaxEntries` infinite loop was already fixed in e6b466f
- **No auth bypass** — cache only serves values previously retrieved through `signedFetch`; cache hits never skip authorization for _new_ data
- **No secrets in code** — clean
- **Write-dedup safety** — a skipped write only suppresses a proven no-op (fingerprint from a confirmed server response within TTL); failed PUTs correctly invalidate the cache entry so the next write retries

### Code Quality

- The in-flight coalescing pattern (object-identity ownership tracking) is clean and handles all critical races correctly: SET completing before pending GET → stale GET detached; GET completing before SET → SET overwrites cache
- FNV-1a is appropriate for QuickJS (no native crypto); hash+length fingerprint provides strong collision resistance for same-key comparisons
- `wrapSignedFetch` fix is correct: `JSON.parse()` throws synchronously before `tryCatch` can intercept — the explicit try/catch restores the intended error tuple
- URL memoization pattern (memoize the promise, not the value) correctly collapses concurrent first calls and clears on failure

### CI

- CLI E2E failure is infrastructure-related (missing `scene/package.json` in CI runner) — also failing on unrelated PR #1498. Not caused by this PR.
- All relevant checks pass: build ✅, lint ✅, test ✅, docs ✅

No issues found. Clean work.

---
Requested by Gon via Slack

packages/@dcl/sdk/src/server/storage/value-cache.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+  return hash >>> 0
+}
+
+/**
+ * Cheap fingerprint of a serialized body: fnv1a hash plus length. The length
+ * guards against hash collisions between same-key values.
+ * @internal
+ */
+export function fingerprint(serialized: string): string {
```
gonpombo8: **Suggestion: drop the fingerprint and compare the cached body directly.**

Every cache entry that carries a `print` also retains the full serialized `body` (it's needed to serve read hits), so the write-dedup check in `set()` can simply be:

```ts
if (skipIfUnchanged && cache.get(key)?.body === body) return true
```

That makes `fnv1a`, `fingerprint`, and the `print` field removable (in both scene and player storage, plus the `getValues` seeding).

Rationale:

- **Correctness**: FNV-1a is a 32-bit non-cryptographic hash. A collision — even with the `:length` guard — means `set()` silently skips writing a genuinely different value, which is data loss, the worst failure mode for a storage API. Direct string equality has zero false positives by construction.
- **No perf win from hashing**: `set()` has to serialize the value either way. Hashing then walks the full string unconditionally, while `===` does an O(1) length check first and exits early on the first mismatch. String compare ties or wins in every case.
- **Negligible memory at stake**: `print` is ~16 bytes, so this isn't about memory — it's about removing a second concept (hash function + collision mitigation + its tests) that answers a question `body` already answers exactly.

The one future scenario where a fingerprint would earn its keep is capping retained body size (e.g. "don't keep bodies over N KB for read caching, but remember their hash for write dedup" — `CacheEntry`'s independently-optional `print?`/`body?` shape hints at this). Nothing in the PR exercises that path today, and reintroducing a fingerprint later is a small, well-understood change — YAGNI says don't pre-build it.

pravusjif: sounds good! tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/2d827f50533c491524da61e2f7dff17483d11af0

packages/@dcl/sdk/src/server/storage/scene.ts
```diff
@@ -73,37 +142,56 @@ export const createSceneStorage = (): ISceneStorage => {
...
+
       if (error) {
+        // The PUT may have reached the server, so the cached body is no
+        // longer reliable.
+        cache.delete(key)
         console.error(`Failed to set storage value '${key}': ${error}`)
         return false
       }
 
+      cache.set(key, { body })
```
cyaiox: [P1] `cache.set(key, { body })` runs when **this PUT's response** arrives, but with two overlapping `set()`s for the same key (e.g. fire-and-forget writes from a tick loop), client response order can differ from server commit order — the cache can end up asserting the value the server did **not** keep, for up to `cacheMaxAgeMs`. Reads then serve the wrong value and `skipIfUnchanged` suppresses the write that would correct it. The in-flight guard covers GET-vs-SET but not SET-vs-SET.

Suggested fix: serialize PUTs per key with latest-value coalescing — at most one in-flight PUT per key; a `set()` issued while one is in flight replaces a single queued "latest value" and resolves when its own PUT lands. That guarantees server final state == last `set()` issued (today overlapping PUTs leave the kept value to commit-order chance even ignoring the cache — the service is last-write-wins with no version tokens), makes write-through trivially correct, and collapses N rapid writes into ≤2 PUTs — a further win for exactly the tick-loop workload this PR targets.

Same applies to `player.ts` `set()`.

packages/@dcl/sdk/src/server/storage/scene.ts
```diff
@@ -137,6 +225,17 @@ export const createSceneStorage = (): ISceneStorage => {
...
 
       const data = response?.data ?? []
+
+      // Seed the per-key cache so subsequent get()/set() on returned keys can
+      // skip the network. Absence is never seeded (prefix/pagination make it
+      // non-authoritative). A page larger than cacheMaxEntries churns the
+      // cache; entries repopulate lazily.
+      for (const entry of data) {
+        if (entry.value !== undefined) {
+          cache.set(entry.key, { body: JSON.stringify({ value: entry.value }) })
```
cyaiox, line 257: [P2] `getValues()` seeding lacks the race guard `get()` has: a `set()`/`delete()` completing while the list request is in flight gets overwritten here by the stale page snapshot, which is then served/deduped for up to `cacheMaxAgeMs`. Suggested fix: per-key write generation (bumped in `set()`/`delete()`), snapshotted before the fetch — skip seeding keys whose generation changed. Same in `player.ts`.

packages/@dcl/sdk/src/server/storage/value-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return {
+    get(key: string): CacheEntry | undefined {
+      const entry = entries.get(key)
+      if (!entry) return undefined
+
+      // Lazy max-age expiry: storedAt is never refreshed on hits, so the age
+      // bounds the time since the last actual network confirmation.
+      if (Date.now() - entry.storedAt > config.cacheMaxAgeMs) {
```
cyaiox: [P2] `cacheMaxAgeMs` isn't guarded the way `cacheMaxEntries` is above: after `configure({ cacheMaxAgeMs: NaN })` (easy to hit via arithmetic on an undefined input), `Date.now() - entry.storedAt > NaN` is always false — entries never expire and write-dedup trusts them forever. Mirror the `Number.isFinite` fallback used for `cacheMaxEntries`. (Negative values already degrade safely — everything is expired.)

packages/@dcl/sdk/src/server/storage/index.ts
```diff
@@ -28,7 +39,10 @@ const createStorage = (): IStorage => {
     delete: sceneStorage.delete,
     getValues: sceneStorage.getValues,
     // Keep player as nested property
-    player: playerStorage
+    player: playerStorage,
+    configure(options: StorageOptions): void {
+      Object.assign(config, options)
```
cyaiox: [P2] `Object.assign` copies explicitly-undefined properties: `configure({ cacheReads: undefined })` — easy to produce when forwarding a partial options object — sets `config.cacheReads = undefined`, silently disabling read caching (`skipIfUnchanged` likewise). Suggest dropping `undefined` values before assigning.

cyaiox (commented):
Reviewed the caching layer against the world-storage-service implementation — mechanics are solid and test coverage is excellent. Inline comments below: one write-path race (P1) and three reliability guards (P2).

packages/@dcl/sdk/src/server/storage/index.ts
cyaiox: this one is a nitpick

packages/@dcl/sdk/src/server/storage/scene.ts
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/d12e1a72b85aa2a8f236d0082c601f0a9dea52b2

packages/@dcl/sdk/src/server/storage/scene.ts
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/d12e1a72b85aa2a8f236d0082c601f0a9dea52b2

packages/@dcl/sdk/src/server/storage/value-cache.ts
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/d12e1a72b85aa2a8f236d0082c601f0a9dea52b2

packages/@dcl/sdk/src/server/storage/index.ts
pravusjif: tackled at https://github.com/decentraland/js-sdk-toolchain/pull/1497/commits/d12e1a72b85aa2a8f236d0082c601f0a9dea52b2

cyaiox (approved):

PR #1498, opened 2026-07-22
feat: serve asset bundles in preview through an abgen sidecar

eordano:
## What

`sdk-commands start --asset-bundles` runs an [abgen](https://github.com/decentraland/abgen) sidecar: an ab-cdn-compatible server that JIT-converts the scene being previewed into Unity asset bundles, reading it through the preview's own `/content` endpoints. The CLI prints the sidecar URL and adds `optimized-assets-url` to the Desktop Explorer deep link. **Off by default** — the flag opts in.

- Binary resolution: `$ABGEN_BIN` → cached copy of the pinned release (v0.11.3) → `abgen` on the PATH → only then a one-time, sha256-verified download → graceful degradation with a warning.
- Converted bundles persist in `.dcl-optimized-assets/` next to `scene.json` (inside the default dcl-ignore).
- Preview content ids switch to base64url so they stay safe in URL paths and asset-bundle directory names.

## Review findings addressed since #1448

- **exec() double settlement**: added a `settled` guard so `error` + `close` can't both settle the promise.
- **Concurrent-download race**: releases extract into a per-call staging dir and publish via atomic rename; a losing resolver detects the winner and proceeds.
- **tar hardening**: `--no-absolute-names` was suggested but is rejected by both GNU tar 1.35 and bsdtar 3.8.8 (verified locally — it would break extraction everywhere). Both tars already strip absolute paths and refuse `..` members unless `-P` is passed; documented in place.
- **Round-trip coverage**: extracted `b64UrlHashDecodingFunction` (now used by the `/content/contents/:hash` route) with encode→decode round-trip tests covering base64url-specific characters (`-`, `_`) and unicode paths.
- **Weak negative assertions**: `expect.arrayContaining([expect.not.stringContaining(…)])` → `expect.not.arrayContaining([expect.stringContaining(…)])` in explorer-alpha.spec.ts (all three occurrences).
- **Lint**: fixed every outstanding lint finding across packages (repo lint is now fully clean).

## Cross-component P1s (companion PRs)

The base64url switch needs consumers to accept `-`/`_` in `b64-` hashes:
- **bevy-explorer**: decentraland/bevy-explorer#995 — decode accepts both alphabets, re-encode uses base64url.
- **hammurabi-headless**: branch [`fix/accept-base64url-content-hashes`](https://github.com/decentraland/hammurabi-headless/compare/main...fix/accept-base64url-content-hashes) is pushed with the regex fix + tests (PR pending).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,7 @@ For deeper context on a specific subsystem (component serialization, CRDT suppre
 
 ## Packages
 
-Six packages under `packages/@dcl/` (npm workspaces):
+Six packages under `packages/@dcl/` (a monorepo layout, but **not** npm workspaces):
 
 | Package | Purpose |
 | --- | --- |
@@ -26,6 +26,8 @@ Six packages under `packages/@dcl/` (npm workspaces):
 | `@dcl/playground-assets` | Browser-compatible SDK bundle |
 
 Internal references use `file:../` paths during development; published versions get pinned semver ranges (managed via syncpack — see `make sync-deps`).
+
+The root `package.json` has no `workspaces` field — root `make install` only installs root-level tooling deps. Each package keeps its own standalone `package-lock.json`, and per-package `node_modules/` are populated by `make build`, which runs `npm i` inside each package (`scripts/build.spec.ts`). Don't expect package deps to be hoisted to the root `node_modules/`.
 
 ## Development commands
 
@@ -72,6 +74,21 @@ grep -rn "ERR!" test/snapshots/        # must return no output
 An `ERR!` line means the QuickJS eval threw mid-execution (commonly a missing mock in `test/snapshots.spec.ts`'s `require` shim — e.g. a new `~system/Runtime` method the SDK started calling at module-load). The resulting snapshot captures a *truncated* run, hides real behavior, and would mask future regressions.
 
 **Anti-pattern:** committing snapshots containing `ERR! Error: Unknown module ...`, `ERR! TypeError: ... is not a function`, or any other `ERR!` trace. Treat them as broken artifacts — fix the mock (or the underlying scene-load failure), regenerate, and re-verify before committing.
+
+### Bumping `@dcl/inspector` (the vehicle for `@dcl/asset-packs`)
+
+`@dcl/asset-packs` is not a direct dependency of this repo — it ships nested inside `@dcl/inspector`, and `packages/@dcl/sdk-commands/src/logic/bundle.ts` resolves it from the inspector's `node_modules`. To bump (prior art: commits `04270ca5`, `8b6bd63d`):
+
+1. `cd packages/@dcl/sdk-commands && npm i --save-exact @dcl/inspector@<x.y.z>` — updates the pin, the package's standalone lockfile, and its `node_modules/` in one step (same pattern as the Makefile's `update-protocol` target).
+2. `make build` from the repo root.
+3. Regenerate snapshots with a `UPDATE_SNAPSHOTS=true` scoped Jest run — `test/snapshots/package-lock.json` updates itself during this run. Verify no `ERR!` lines (see above).
+
+Asset-packs injection is gated behind `isEditorScene` (requires `assets/scene/main.composite` — see `packages/@dcl/sdk-commands/src/logic/project-validations.ts`), so inspector bumps no longer change `.crdt` snapshots. An empty snapshot diff is expected, not a stale artifact.
+
+## Committing hygiene gotchas
+
+- A newer local npm rewrites committed lockfiles with `"peer": true` / `"dev": true` metadata churn during installs (root and per-package lockfiles). Revert lockfile changes unrelated to your dependency change before committing.
+- `make format` runs prettier over the whole repo, including paths CI's `make lint` does not check (`test/`, `scripts/`, dot-files), where HEAD may carry drift — a blind `make format` can dirty dozens of unrelated files. Check your own files, revert the rest.
 
 ## Code conventions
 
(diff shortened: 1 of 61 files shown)


dalkia:
Pushed the first-load-timeout work discussed in review, plus fixes it surfaced:

- **`730f4825` / `9657c221`** — pre-conversion on start: `--asset-bundles` now converts the scene *before* printing the deeplink (the sidecar holds a manifest request until conversion completes, so this is the corpus build over HTTP — no `abgen-corpus` needed). Explorer's manifest fetch is always warm; the first-load timeout is gone. Dropped connections retry and re-attach to the in-flight build; a dead sidecar bails to the raw-GLTF degrade instead of burning the 15-minute budget. Measured on Genesis Plaza central-plaza (1009 files): 91s cold, 3s warm.
- **`99361e9c` / `46e1fa59`** — readable output: the sidecar now runs quiet by default so its per-file telemetry doesn't drown the progress lines (exporting `RUST_LOG` restores full sidecar logs). Silencing exposed a latent `exec` bug: silent children's pipes were never drained, so a chatty child filled the 64KB pipe buffer and blocked at 0% CPU (readyz still green). `exec` now drains silent pipes; regression-tested.
- **`e3ed9c8d`** — per-asset progress: polls the sidecar's `/progress/{entity}` route (decentraland/abgen#25) and prints `[340/1008] converting <file>` instead of a blind heartbeat; sidecars without the route fall back gracefully.
- **`257aca2c` / `c7f7e5c1`** — hub sessions never self-open the client: Creator Hub always captures the printed `Desktop client:` deeplink, adjusts it, and fires it itself, so the sdk's own launch under `--hub` was at best a double open with unadjusted params — and with CH's background conversions (decentraland/creator-hub#1396) it opened the explorer nobody asked for. Gated on `--hub` (first attempted via `--no-browser`, reworked in review since that flag should keep meaning the web browser); the deeplink line still prints, non-hub CLI behavior unchanged.

nicoecheza:
## Review

Reviewed together with decentraland/creator-hub#1396. This half is in good shape: pinned + sha256-verified download (checked before anything touches disk, with a test asserting no write/extract on mismatch), staged extract with atomic rename and loser-detection, shell-less spawn, loopback-only bind. The `exec()` fixes are real bug fixes beyond this feature — the silent-pipe drain (64KB backpressure deadlock) and the `settled` guard — both tested. The base64url switch is decode-back-compatible: Node's `base64url` decoder accepts standard base64 (`+`, `/`, padding), verified empirically, so old-style `b64-` hashes still resolve in `/content/contents/:hash`. Leaving the analytics/`export-static` `b64HashingFunction` callers untouched is the right call.

### P1 — old hub + new sdk: first Preview press opens nothing

The `isHub` no-self-open change assumes the hub fires the deeplink itself — but **shipped Creator Hub versions only fire it on the reuse path**, never after a fresh spawn (they relied on sdk-commands self-opening). A scene that bumps to this sdk while the user runs an older hub: first Preview press boots the server and no client opens; only the second press works. Hub auto-update narrows the window, but this should be a conscious accept — or key the no-self-open on a new hub-passed flag instead of `isHub` (`--no-client` doesn't work as that marker since it also suppresses the deeplink *print* the hub needs).

(The mirror case — new hub + old sdk double-launch — is flagged on the hub PR.)

### P2 — orphaned sidecar on ready-timeout

If `/readyz` doesn't answer within 15s, `runAssetBundlesSidecar` warns and returns `undefined`, but the spawned abgen keeps running — converting and holding the port until CLI exit — and `exec()` returns no handle to kill it. Slow first boots (e.g. Windows Defender scanning a freshly downloaded binary) hit exactly this. Consider a kill handle on `exec` for `shell: false`, or a longer timeout.

### P3 — `pruneStaleReleases` vs concurrent versions

Two scenes pinning different sdk-commands (hence different abgen pins) previewing concurrently: each resolve prunes the other's release dir, yanking `template/`/`shader/` assets out from under a running sidecar (POSIX: binary survives, assets vanish). Rare but confusing when it hits; an mtime guard (don't prune recently-touched dirs) would cover it.

### Nit — cross-repo log-format contract

Creator Hub parses `printProgressStep`'s `[n/total] <log>` shape and the literal `still converting... (Ns)` string for its progress UI. The hub side documents the coupling; this side doesn't. A one-line comment at the heartbeat/`printProgressStep` call sites would prevent an innocent reformat from silently degrading hub progress (it fails open, but silently).

### Verdict

Approve with the above as follow-ups; only the version-skew point needs an explicit decision before merge.

nicoecheza (approved):
Re-reviewed at `b402ade2`.

The two reverts (`4d270edf`, `b402ade2`) resolve the version-skew P1: `runApp` self-opens the client unconditionally again, so this PR no longer changes who launches the explorer for any hub or non-hub session. Combined with creator-hub#1396 dropping its fresh-spawn `dclDeepLink` fire, the launch contract is now exactly what shipped before — no double-launch, no first-press-opens-nothing, in either direction of version skew. That was the only blocker.

Everything else I checked before still holds up: sha256 verified before anything touches disk, staged extract + atomic rename, shell-less spawn, loopback bind, the `exec()` pipe-drain and `settled` fixes, and base64url decode staying back-compatible with existing `b64-` hashes.

The three follow-ups from my earlier pass are unchanged and still worth doing at some point, none blocking:

- **Orphaned sidecar on ready-timeout** — after the 15s `/readyz` deadline, `runAssetBundlesSidecar` warns and returns `undefined`, but the spawned abgen keeps converting and holding its port until CLI exit, since `exec()` hands back no kill handle. A slow first boot (Windows Defender scanning a freshly downloaded binary) is the realistic trigger.
- **`pruneStaleReleases` vs concurrent versions** — two scenes on different sdk-commands pins previewing at once will each prune the other's release dir, pulling `template/`/`shader/` out from under a running sidecar. An mtime guard would cover it.
- **Log-format coupling** — Creator Hub parses `printProgressStep`'s `[n/total] <log>` shape and the literal `still converting... (Ns)` string. A comment at those call sites would stop an innocent reformat from silently degrading hub progress.

Approving.

dalkia:
Deprecated in favour of https://github.com/decentraland/unity-explorer/pull/9704

PR #1504, opened 2026-07-29
feat: serve preview asset bundles through the preview server itself (single origin)

dalkia:
Implements the sdk-commands side of [decentraland/abgen#21](https://github.com/decentraland/abgen/issues/21) (filed there because issues are disabled in this repo). Stacked on #1498.

## What

Mounts the abgen sidecar behind the preview server as a thin streaming reverse proxy:

```
GET  {realm}/optimized-assets/manifest/{id}_{platform}.json  →  sidecar
GET  {realm}/optimized-assets/v49/{id}/{bundle}              →  sidecar
POST {realm}/optimized-assets/entities/active                →  sidecar
```

Clients now only ever know **one origin — the realm they already have**. The sidecar's port becomes fully private to sdk-commands, and the deep link carries no assets URL at all: `local-ab=true` alone, with the explorer deriving the `/optimized-assets` base from the realm.

## How

- **New `/optimized-assets/:path+` route** (`server/asset-bundles-proxy.ts`), following the existing `/lambdas/:path+` proxy pattern: streams every method through to the sidecar (the explorer's registry POSTs included) and forwards status and headers both ways. The target is late-bound — the sidecar boots *after* the preview server (it reads the scene through the preview's own `/content` endpoints) — so the route answers 503 until the sidecar is up.
  - **Request headers ride along** minus the per-connection ones (`host`, `connection`, `content-length`, `accept-encoding`): without their `content-type`, the sidecar 415s the explorer's JSON registry POSTs, which broke wearable/catalyst entity resolution. Dropping `connection` (rather than forcing `close`) keeps undici's keep-alive pool to the sidecar warm across the hundreds of bundle GETs of a scene load.
  - **Response headers are filtered, not mutated**: `fetch()` responses carry immutable `Headers` in undici — `headers.delete()` threw `TypeError: immutable` on every proxied request — so `content-encoding`/`content-length` (stale after undici's transparent decompression) are dropped from a plain-object copy instead. The same pre-existing bug on `main`'s `/content/entities` deploy proxy is split out as #1506 (with its test); the fix commit is also cherry-picked here so branch builds keep deploying — it resolves to a no-op once #1506 lands.
- **Sidecar port moves back to `getPort(0)`**: the 5147 preferred-port convention existed only so a deeplink-less Unity Editor could guess the sidecar's address; with the proxy, everything derives from the realm origin instead.
- **`optimized-assets-url` deeplink param is gone**: `runExplorerAlpha` takes a plain `assetBundles` boolean and emits only `local-ab=true`. The explorer infers the assets base from the realm, so there is no URL to pass.

## Follow-up (other repos)

- **unity-explorer** (after decentraland/unity-explorer#9459): when `local-ab` is set, derive the assets base from the realm — note the base carries a path prefix (`/optimized-assets`), which the URL-building code must handle. This is now required, not optional: sdk-commands no longer sends an explicit URL.
- **abgen**: nothing — this only changes who fronts it.

## Test plan

- New `asset-bundles-proxy.spec.ts`: mount path, 503 before the sidecar is up, prefix-strip + query-string forwarding + body streaming on GET, method/body/status forwarding on POST, request-header forwarding (client `content-type` kept, per-connection headers replaced), and response-header scrubbing against a fetch-faithful mock whose `Headers.delete` throws `TypeError: immutable`.
- Updated `explorer-alpha.spec.ts`: `local-ab=true` appears exactly when asset bundles are enabled; `optimized-assets-url` never does.
- Validated live against a running preview: wearable URN resolution through `POST /optimized-assets/entities/active` returns the full entity with complete bundle status.
- `npx tsc --noEmit` clean on sdk-commands; all tests under `test/sdk-commands/commands/start` pass.

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/start/asset-bundles.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/asset-bundles.ts
@@ -1,5 +1,4 @@
 import * as path from 'path'
-import portfinder from 'portfinder'
 
 import { CliComponents } from '../../components'
 import { printProgressInfo, printProgressStep } from '../../logic/beautiful-logs'
@@ -17,11 +16,6 @@ const PREWARM_FALLBACK_EVERY_TICKS = 4 // seconds-heartbeat cadence when /progre
 const PREWARM_TIMEOUT_MS = 15 * 60_000
 const PREWARM_RETRY_DELAY_MS = 2_000
 
-// abgen's default port, and the Explorer's default `optimized-assets-url`:
-// preferring it lets a connected Unity Editor find the sidecar with zero
-// configuration. Scans upward when taken (the deeplink carries the real URL).
-const PREFERRED_SIDECAR_PORT = 5147
-
 /**
  * Boots an `abgen` sidecar: an ab-cdn-compatible server that JIT-converts the
  * scene being previewed into asset bundles, reading it through the preview's
@@ -29,6 +23,10 @@ const PREFERRED_SIDECAR_PORT = 5147
  * undefined — with a warning — when the binary is missing or never comes up.
  * The binary resolves from $ABGEN_BIN, then a cached copy of the pinned abgen
  * release, then `abgen` on the PATH, and downloads the release only when none exist.
+ *
+ * The sidecar binds an ephemeral port that stays private to sdk-commands:
+ * clients reach it through the preview server's /optimized-assets proxy, so
+ * they only ever see the realm origin they already have.
  */
 export async function runAssetBundlesSidecar(
   components: Pick<CliComponents, 'fetch' | 'logger' | 'spawner' | 'config' | 'fs'>,
@@ -36,7 +34,7 @@ export async function runAssetBundlesSidecar(
   projectRoot: string
 ): Promise<string | undefined> {
   const bin = await resolveAbgenBin(components)
-  const port = await portfinder.getPortPromise({ port: PREFERRED_SIDECAR_PORT }).catch(() => getPort(0))
+  const port = await getPort(0)
   const url = `http://127.0.0.1:${port}`
   const catalystUrl = await getCatalystBaseUrl(components)
   // next to scene.json: converted bundles survive preview restarts (never
--- a/packages/@dcl/sdk-commands/src/commands/start/explorer-alpha.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/explorer-alpha.ts
@@ -37,12 +37,12 @@ export async function runExplorerAlpha(
     baseCoords: { x: number; y: number }
     isHub: boolean
     args: Result<typeof startArgs>
-    assetBundlesUrl?: string
+    assetBundles?: boolean
   }
 ) {
-  const { cwd, realm, baseCoords, isHub, assetBundlesUrl } = opts
+  const { cwd, realm, baseCoords, isHub, assetBundles } = opts
 
-  if (await runApp(components, { cwd, realm, baseCoords, isHub, args: opts.args, assetBundlesUrl })) {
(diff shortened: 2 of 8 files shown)


nicoecheza (commented):
Single-origin is the right design, and the route implementation follows the existing `router.all('/lambdas/:path+')` pattern closely (I confirmed `.all` + `:path+` are already proven here at `endpoints.ts:109`/`:153`). Late-binding the target through a getter is well-commented, and the test that makes `headers.delete` throw to mimic a real immutable `Headers` guard is a genuinely good touch. One blocking issue though.

## P1 — `optimized-assets-url` is dropped entirely, and that breaks creator-hub#1396

The description says the param "now carries `{realm}/optimized-assets`" and "keeps being emitted until explorers derive the assets base from the realm". The code does the opposite: `explorer-alpha.ts` only sets `local-ab=true`, never the url, and the updated spec asserts its absence. Either the description or the implementation is stale — and the evidence says implement the description:

1. **No shipped explorer derives the base yet.** Your own follow-up section says unity-explorer needs that change *after* #9459 — so today's explorers get `local-ab=true` with no base and fall back to their default, which was the 5147 convention this same PR just abandoned. All asset-bundle traffic (wearables and avatar included) lands on a dead port. That's verbatim the failure mode #1498's own comment in `updateDeepLinkWithOpts` warns about.

2. **creator-hub#1396 detects the sidecar purely by the presence of `optimized-assets-url`** in the captured deeplink, so dropping it breaks it two ways:
   - `hasSidecarUrl = urlParams.has('optimized-assets-url')` → false, so the hub **strips `local-ab` on every re-focus** and logs "previewing with raw GLTFs". Optimized assets silently stop working after the first launch.
   - `previewHasSidecar` is then always false while `wantsSidecar` is true, so the "toggle disagrees with the running preview" branch fires on **every Preview press — killing and respawning the preview every time**, forever.

Emitting `optimized-assets-url={realm}/optimized-assets` fixes both, is backwards compatible with current explorers, and is exactly the migration path the description already describes. If dropping it really is intended, creator-hub#1396 needs a coordinated change to detect the sidecar some other way before either merges.

## P2 — `connection: close` kills keep-alive to the sidecar

I checked: undici does forward this header, so every proxied request opens and tears down a fresh TCP connection to the sidecar. A scene load is potentially hundreds of bundle GETs on this exact path. Nothing needs it — just `delete requestHeaders['connection']` alongside `host` and let undici manage its own leg.

## P3 — a mid-session sidecar death becomes a 500

The 503 only covers "not up yet". If abgen crashes or gets killed after boot, `components.fetch.fetch` rejects with ECONNREFUSED and the handler has no `try`/`catch`, so the client gets a 500 with a stack instead of a clean 502/503 — and the explorer's degrade-to-raw-GLTFs behavior may well hinge on which it sees. Not covered by the new spec either.

## P3 — the proxy makes a deliberately loopback-only service LAN-reachable

#1498 binds the sidecar to `127.0.0.1` on purpose; the preview server binds `0.0.0.0` (`start/index.ts:180`). Mounting one behind the other exposes abgen — every method, including POST — to anyone on the network. It's partly inherent to single-origin (mobile preview needs the LAN path) and the preview server already serves scene content there, but abgen is a much larger surface than static files: a separate binary that writes to disk and does heavy CPU work on demand. Worth an explicit decision rather than a side effect — document it, or restrict the proxy to GET/HEAD if the registry POSTs don't actually need to reach abgen.

## Worth splitting out

The `endpoints.ts` hunk is a real bug fix that has nothing to do with this feature: `res.headers.delete()` on a `fetch()` response throws `TypeError: immutable` (verified in Node), so the `/content/entities` POST proxy currently 500s on **`main`** — that code predates #1498. Since main is affected, that one-line fix deserves its own PR with a test rather than riding a stacked draft.

nicoecheza (approved):

PR #1505, opened 2026-07-29
[DO NOT MERGE][PoC] feat: Refactor multiplayer network layer

cyaiox:
# Layer A refactor: authoritative multiplayer network layer

Refactors `@dcl/sdk`'s network layer around the centralized authoritative-server topology, plus a reliability pass on the local dev server. **Public API and wire format are frozen** — the migration is invisible to scenes, and that claim is machine-checked (see below).

## Why

The `auth-server` branch's network layer carried structural debt from the P2P era: a 240-line god-closure holding hydration state in loose booleans, a 3-module import cycle, two competing `isServer` sources, a hot path that parsed each message ~3×, default-allow validation, and an acknowledged-broken correction path. An evaluation catalogued 15 correctness gaps; the user-visible symptom is the #1 community complaint — *"some players see an entity and others don't."* Separately, creators reported the local hammurabi dev server "stops working during development."

## How it's proven safe (TDD structure)

1. **Characterization baseline first** (`1e16e307`): before any refactor, new specs pinned the current behavior — the public API surface, golden-byte wire format (translation round-trips, 12KB chunking, `engineToCrdt` dumps, comms envelope), and 3-engine message flow. These pass unchanged on every commit in this PR; the wire goldens are **byte-identical** across the whole refactor.
2. **Red defect suite**: 13 `it.failing` tests, one per defect, each asserting the *correct* behavior and verified failing before its fix. Phases flip them green; a flip-check gate verifies the direction on every phase.
3. **Convergence oracle**: every E2E scenario ends by asserting all engines' networked state is semantically identical — the users' #1 complaint as a CI regression gate.

## What changed, by commit

- `fb021e69` — pre-existing WIP carried from the working tree (world-storage namespacing by scene coordinates). **Not part of the refactor**; drop or keep per review.
- `5a0afb4f` — **local dev-server reliability** (sdk-commands): supervise/respawn the hammurabi child with backoff; kill the whole process tree (no more orphans holding the port); pin `@dcl/hammurabi-server` to an exact version (`@next` re-resolution was the "wait a few minutes" symptom); crash-proof the preview server (ws `error` handler, debounce rejection handling, process-level handlers); explicit-`--port` pre-check with an actionable error; atomic `server-storage.json` writes + loud reset warning; guarded mid-save `scene.json` parse; restart the server child on scene rebuild (no more stale bundles).
- `1e16e307` — characterization baseline + red suite + 3-engine harness + convergence oracle (1,540 lines of test infrastructure).
- `cb370c8f` — **foundations**: `constants.ts` leaf module dissolves all four import cycles; one `isServer` atom per transport (was two atoms / two API calls); error logs un-gated from DEBUG; `throw 1`, `debugger`, dead code removed; `ecs-adapter.ts` consolidates 22 deep `@dcl/ecs/dist` imports behind one seam; `observables.ts` switch fall-through fixed (duplicate subscriber registrations).
- `c2e0f2fa` — **hydration FSM**: pure state machine (`DISCONNECTED → CONNECTED → REQUESTING_STATE → SYNCED`) replacing the loose booleans; stray hydration replies are unrepresentable as FSM events (retry-halt bug structurally impossible); server-gated state responder; ghost reconciliation on re-hydration (multi-chunk-safe, local-only removal); `SERVER_ANNOUNCE` generation signal → clients re-hydrate on server restart; comms buffered until the role resolves; `RealmInfo` observed on the injected engine. Defects #1 #2 #6 #7 #9 #14 fixed.
- `a9334d80` — **validator hardening**: discriminated switch with default-deny (a client-sent `AUTHORITATIVE_PUT_COMPONENT` was previously accepted by the fallthrough); `DELETE_ENTITY` requires creator ownership; the correction path always corrects — a rejected *first* write gets a `DELETE_COMPONENT_NETWORK` at `rejectedTimestamp+1` that wins LWW on merit; O(1) validated-hit entity index; `docs/network-timestamp-trust.md`. Defects #3 #4 fixed.
- `7f87d8c7` — **codec unification** (highest-risk, wire goldens byte-identical): `codec.ts` is the sole owner of read/translate/pack; server ingest parses each buffer exactly once (spy-asserted) and re-broadcasts retained byte slices; one `packChunks` with O(1) size accounting; one loud oversize error path naming component/size/limit; fixed a latent ~28B/message undercount in dump sizing. Defect #8 fixed.
- `f9f4bb1e` — **topology**: clients send CRDT/state-requests/room-events targeted `[authoritative-server]` (client traffic O(N²)→O(N)); `~sdk/` reserved message prefix + collision reporting; players helper deduped (one per-frame system per engine) and its diff loop rewritten mark-and-sweep — fixing a real join-miss when a join and a leave landed in the same frame. Defects #10 #12 fixed.
- `bf9a51a6` — **boot cleanup + host contract**: `transportInitializedTicks` param replaces NODE_ENV sniffing (tests opt in explicitly; no environment detection left in the package); renderer transport deliberately *not* gated server-side (the readable headless host ingests CRDT through `crdtSendToRenderer` — evidence in doc); `docs/host-conformance.md` — the `~system` checklist a new host (e.g. Bevy) validates against.

## Deferred with measured evidence (not silently dropped)

Two defects need layers below this repo; both stay pinned as `it.failing` with probe evidence in `docs/network-peer-visibility.md`:

- **#11 echo-to-sender**: the headless server provably has no peer roster (avatar components are renderer-written; seen-senders can't reach silent listeners; `~system/Players` is unverified on hammurabi and deprecation-marked — a *partial* roster would cause silent divergence, worse than the echo, whose re-apply is an LWW no-op). Needs a host roster or, better, a wire-level **broadcast-except-X** comms primitive.
- **#13 same-frame join+leave**: the ECS collapses local same-frame create+delete to net state at flush; the *wire* path notifies per message, so the actual gap is a **component-wide change subscription** in `@dcl/ecs` (per-entity `onChange` can't be pre-subscribed for entities that don't exist yet).

Also explicitly out of scope (recorded in the plan): `@dcl/sdk/server` framework (Layer B), AOI/interest management (the Strategy seam for it now exists), `@dcl/ecs` internal-API promotion (its 100%-coverage gate makes that a separate PR), legacy message-bus/observables deprecation.

## Verification

- `make lint` ✅  `make build` ✅
- `test/sdk/network`: 14 suites / 104 tests / 7 wire golden snapshots ✅ (goldens byte-identical to pre-refactor)
- Red suite: 11/13 plain-`it` green; flip-check confirms exactly #11/#13 remain, each with a deferral note
- QuickJS snapshots: 17/17, `grep -rn "ERR!" test/snapshots/` clean; only size/memory metric lines drifted
- `test/sdk-commands`: 24 suites / 176 tests ✅
- Full `make test`: **green** — 163/164 suites (1 pre-existing skip), 1190 tests, 0 failures. (Two quirks observed only under scoped `--testPathPattern` runs, both pre-existing: `test/sdk/message-bus.spec.ts` hits TS2591 when compiled in a narrow file set, and `test/sdk-commands/utils/commands.spec.ts` occasionally exceeds its 5s timeout. Neither manifests in the full suite.)
- Not yet run (needs a real scene + explorer): live E2E against a spawned hammurabi (late-join + server-restart convergence) — the Workstream D acceptance list has the manual steps

https://claude.ai/code/session_01Kzo6zwrn1CA79S4RN1dgsU


gonpombo8 (changes requested):
Thanks for the effort, but I'm rejecting this one.

This is the most critical part of the SDK for creators and it's a 4k line change — I honestly can't review something this big in this layer, and I don't get why we need most of these changes. If something breaks in prod we won't even know which part did it.

If some of these fixes are real, let's do them as small separate PRs, each one explaining what problem it solves. The dev-server fixes and the test harness could be a good start, those stand alone.

cyaiox:
Superseded by the stacked split: network layer #1518→#1525 (test harness first), sdk-commands #1526→#1527. Same commits, rebased onto current auth-server (5ffe873b). Keeping this PR as the unified view until the stack is reviewed — close whenever.

PR #1503, opened 2026-07-28
feat: expose openExplorerUi in ~system/RestrictedActions

popuz:
## What
Bumps `@dcl/protocol` to the **main** build `1.0.0-30373157983.commit-2c475cb` ([decentraland/protocol#451](https://github.com/decentraland/protocol/pull/451), merged) so `~system/RestrictedActions` regenerates with:
- `openExplorerUi(body: OpenExplorerUiRequest): Promise<OpenExplorerUiResponse>`
- `ExplorerUi` enum (7 sections: `EU_MAP`, `EU_SETTINGS`, `EU_BACKPACK`, `EU_CAMERA_REEL`, `EU_COMMUNITIES`, `EU_PLACES`, `EU_EVENTS`)
- `OpenExplorerUiResult` enum + `OpenExplorerUiResponse`

Main-based counterpart of #1485 (experimental, pinned a protocol branch build): once this lands and main syncs into experimental, #1485 will be closed.

## Why
Lets SDK7 scenes open native Explorer fullscreen panels from a user gesture and receive a typed verdict. Client implementation: [decentraland/unity-explorer#9472](https://github.com/decentraland/unity-explorer/pull/9472). Test world: `flutterecho.dcl.eth` + `sdk7-test-scenes` parcel `80,-5`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1511, opened 2026-07-30
feat: add ExplorerUiEventsResult component and share the ExplorerUi enum

popuz:
Round 2 of scene-triggered explorer UI. Round 1 (#1503) gave scenes `openExplorerUi`; this adds the result channel, so a scene can observe when the panel it opened is actually shown or dismissed.

Depends on decentraland/protocol#455 — **do not merge before it**.

## What changes

- `ExplorerUiEventsResult` is registered in `GROWN_ONLY_COMPONENTS`, so it generates as a grow-only value set (`APPEND`) rather than LWW, matching `PointerEventsResult` and friends. The generated `maxElements: 100` lines up with the explorer's `MAX_APPEND_COMPONENTS_COUNT`.
- `~system/RestrictedActions` no longer re-declares `ExplorerUi`. The enum moved to `sdk/components/common/explorer_ui.proto` upstream, so `apis.d.ts` now reads `ui: import('../ecs').ExplorerUi` and scene code sees a single nominal type:

  ```ts
  for (const e of ExplorerUiEventsResult.get(engine.RootEntity)) {
    if (e.ui === ExplorerUi.EU_MAP) { /* no cast needed */ }
  }
  ```
- Regenerated: `playground-assets.api.md`, 12 CRDT snapshots, three lockfiles.
- New: `test/ecs/components/ExplorerUiEventsResult.spec.ts` (3 tests). The accumulation test is a deliberate guard — the serialization tests alone would still pass if the `GROWN_ONLY_COMPONENTS` entry were dropped.

## Before merging

- [ ] decentraland/protocol#455 merged
- [ ] `@dcl/protocol` repinned from the PR tarball to the published version, lockfiles refreshed, snapshots and `api.md` regenerated

The pin is currently a tarball URL in both `package.json` and `packages/@dcl/sdk-commands/package.json`, because PR builds are not published to npm. It carries one unrelated delta into `api.md` — `AvatarEmoteMask`, from `avatar_shape.proto` — which disappears on repin.

## Verification

Locally: `make build` exit 0 (38/38, including the `check no conflict in types` and `ae-forgotten-export` gates); snapshot comparison 17/17.

`make test` on this Windows checkout reports 12 failures in `test/ecs/composite.spec.ts` plus 3 jest timeouts in sdk-commands. The composite failures are a CRLF artifact: the fixtures are `i/lf w/crlf` under `core.autocrlf=true`, and `composite.spec.ts` compares without the `\r\n` normalization that `snapshots.spec.ts` applies. No `ExplorerUi` line appears in any of those diffs and this branch touches no fixture. CI on Linux is the real signal.

## Follow-up

`openExplorerUiAndWaitClose` — a promise-based helper wrapping the call and its matching result event — is a separate PR.


pravusjif (approved):
It only adds the `ExplorerUiEventsResult`, nothing else.

PR #1512, opened 2026-07-31
feat: consolidate player tracking in @dcl/sdk/players

LautaroPetaccio:
> **Scope changed during review.** This PR originally added a separate session-tracking API under `@dcl/sdk/server`. Review showed almost nothing in it was server-specific, and that it would have been a *third* per-frame scan of `PlayerIdentityData` on the same engine. It now reworks `@dcl/sdk/players` instead; `@dcl/sdk/server` is untouched.

## The problem

`onEnterScene` gates on `PlayerIdentityData` **and** `AvatarBase`. The avatar profile can replicate seconds after the identity, or never for a given peer — so a player can be present, addressable and interacting while `onEnterScene` has never fired for them.

That is unusable for anything that needs to know a player exists: presence counts, per-player storage, authorization. So every scene running an authoritative server hand-rolled the same tracker — flagtag, towerofmadness, dead-surge, cozy-farm, My-Dear-Pet. Across those five: **42** `getEntitiesWith(PlayerIdentityData…)` call sites and **384** `.toLowerCase()` calls, most normalizing an address the SDK could have handed over usable. Three also poll `AvatarBase` on a 2–3s timer for names, and one detects arrival from a client-announced message because there is no hook.

## The change

One options object, one diff pass, both thresholds:

```ts
onEnterScene(cb)                             // unchanged: identity + profile, name populated
onEnterScene(cb, { requireProfile: false })  // as soon as the identity exists
```

`requireProfile` defaults to `true`, so existing scenes are unaffected. `onLeaveScene` takes the same option and mirrors whichever threshold you chose.

Also new:

- **`onPlayerNameChanged(cb)`** — fires when the profile name resolves or changes, replacing those polling loops.
- **`getPlayers()` / `getPlayerCount()`** — replacing the hand-rolled scan, duplicate addresses already collapsed. Same element shape as `getPlayer`, from one private builder, so they cannot drift.
- **`displayName`** — never empty: resolved profile name, else a shortened address. `name` keeps its exact meaning (raw profile, `''` until replicated) and `nameResolved` distinguishes them. Sticky, so a profile that momentarily reports empty does not downgrade a known name.
- **`joinedAtMs`**, and a **second argument on `onLeaveScene`** carrying last-known state — `getPlayer()` legitimately returns null once the entity is gone.
- **Unsubscribers** from every subscription.
- **`async` handlers are failure-isolated.** TypeScript's void-return rule lets an `async` callback satisfy `=> void`, so a rejection after an `await` would otherwise escape — under Node's default policy that ends the process, which is how it showed up while measuring this. The callback types stay `=> void` deliberately; the code comments carry why.

Two smaller things travel with it. `definePlayerHelper(engine)` was called twice with the global engine, so two systems scanned the same component every frame; it is now memoized per engine. And `message-bus-sync.ts` bootstraps CRDT state off `onEnterScene`, which had two pre-existing bugs in three lines: it waited for the avatar profile when only the local identity is needed, and compared `myProfile.userId` against `PlayerIdentityData.address` case-sensitively, so any casing difference meant the bootstrap never fired. Both fixed: it now subscribes at the identity threshold with a case-insensitive compare, deferred until the local profile resolves, which `replayPresent` makes safe.

## Bugs fixed

All of these affected client scenes too, independent of the server use case:

- **`players.length === playerEntities.size` early return** compared a count of entities-with-both-components against a map cleaned elsewhere. The two can coincide while the sets differ, silently dropping a join.
- **Leaves ran off `AvatarBase.onChange`**, so a peer that lost identity without a profile change never fired `onLeaveScene`. Narrower than it sounds: entity *deletion* does reach that hook, since `crdtSceneSystem.sendMessages` reports locally-deleted entities even with no transport attached. What the old path misses is the profile-unchanged case. Separately, `onChange` is push-only with no removal path, so those per-entity callbacks accumulated for the life of the process. Leaves are now detected by absence and the hook is gone.
- **An identity with an empty `address` crashed the handler** — `getPlayer({userId: ''})` fell back to the local player entity, returned `null`, and was dereferenced with `!`. Any handler that reads the payload triggers it, and the throw aborts the delivery loop, so a valid player announced in the same pass is lost with it.
- **`getPlayer` pinned to the first entity by insertion order** for a duplicated address, and `onEnterScene` **double-fired** for it, because the key was the entity rather than the address. Not intermittently: `getEntitiesWith` iterates its first component's backing `Map`, which preserves insertion order, and `PlayerIdentityData` is PUT once per peer per join — so iteration order is join order, a stale entity always precedes its replacement, and first-match returned the stale one on every call. Resolution is now keyed on the address and ranked, on both the tracked path and the live-scan fallback.
- **`getPlayer` compared addresses case-sensitively**, so a lowercased address silently returned `null` — a real contributor to all that `.toLowerCase()` at call sites.
- **No unsubscribe and no error isolation** — one throwing handler killed the tracker for the rest of the run.
- **The payload aliased live component data.** `avatar`, `wearables`, `emotes` and `position` were handed straight out. `deepReadonly` is `Object.freeze({ ...value })` — a shallow freeze of a shallow copy — so nested references still point at the component, and a handler could write into it with no dirty-marking: a push into `player.wearables`, or `player.avatar.skinColor.r = 5`. All four are now copied.
- **The tracking system ran at the default priority**, so its position among scene systems depended on insertion order, which `SystemContainer`'s own TODO says is not guaranteed. Since the helper instantiates at module import, a scene registering systems before its first `import('@dcl/sdk/players')` got a tracker running *after* them: lookups read the previous tick's state, and arrivals fired after that frame's scene logic.

## Measured in two scenes

Two standalone scenes run the same 25-check matrix — same `src/checks.ts`, differing by one line of `package.json`:

| Scene | `@dcl/sdk` | Result |
| --- | --- | --- |
| [`players-helper-tracking`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/players-helper-tracking) | `7.26.0` (last release) | `BUG REPRODUCED: 11 of 25 checks fail, 8 not applicable` |
| [`players-helper-tracking-fixed`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/players-helper-tracking-fixed) | `7.26.1-32893350833.commit-c0528a4` (this branch) | `FIXED: all 25 checks pass` |

Each check drives its own isolated `Engine()` and synthesizes players by creating `PlayerIdentityData` / `AvatarBase` directly, so nothing depends on a second peer, a renderer or frame timing. The 8 not-applicable rows need API the old helper lacks; the 11 failures are behaviour present in both versions. Writing it corrected three claims that had been in this description.

## Compatibility

Only what changes behaviour or types:

- Subscription return values changed from `void` to an unsubscribe function, and `onLeaveScene` gained a second parameter — both additive.
- `GetPlayerDataRes` gained `displayName`, `nameResolved`, `joinedAtMs`, and **was not exported before this PR**, so it is newly nameable. The three fields are required, so code constructing the shape *structurally* — typically a hand-built fake player in a test — needs them.
- **Deliberate narrowing:** `getPlayer` returns `null` for an entity with avatar or wearable data but no `PlayerIdentityData`, where it previously returned a partial object with `userId: ''`. Also covers the local player in the window where a host writes `AvatarBase` first — near-theoretical, since unity-explorer and godot-explorer write identity and avatar base together.
- **A subscription made from inside a handler takes effect next tick.** The subscriber list is snapshotted once per pass.
- **A same-tick reconnect now emits nothing** (previously leave + enter). The tracker follows the new entity, but a payload captured at arrival keeps the dead one — re-read via `getPlayer(userId)` rather than caching `player.entity`.
- **`getPlayerCount()` and `getPlayers()` can disagree for one frame** when a backing entity disappears between ticks. Its JSDoc carries the rule: the count for cheap per-frame observation, `getPlayers().length` wherever the number must agree with the roster being acted on.
- **Not purely additive at the type level.** Three shapes now fail to compile: hand-building a `GetPlayerDataRes`; implementing or mocking the helper type; and `export *` from a barrel that already exports a `getPlayers`. Calling code is unaffected, including one-argument `onLeaveScene` handlers and `async` handlers.

## Deliberately not included

- **`PlayerIdentityData` is not server-authoritative.** There is no component allow-list on the inbound CRDT path and no `validateBeforeChange` for it, so a crafted frame can inject phantom players — inflating counts, faking arrival/departure pairs, shadowing a real player's name. Pre-existing, and it affects the hand-rolled trackers above identically; consolidating here at least reduces it to one place to guard. The likely fix is filtering on the reserved entity range, but that is only correct if the headless server really receives avatars in that band, which needs verifying against a live server.
- **`getPlayers()` includes the local player on a client**, documented in its JSDoc with the filtering idiom. Whether that deserves an `excludeSelf` option is left open rather than adding a flag pre-emptively.


gonpombo8:
Hey laucha, we already have something similar in the @dcl/sdk/players entrypoint.
It already resolves the name also, so i dont think we need this session layer 🤔 

```ts
import { getPlayer, onEnterScene, onLeaveScene } from '@dcl/sdk/players'
```

PR #1514, opened 2026-07-31
Default the preview multiplayer server to the bevy engine

gonpombo8:
## Summary

`sdk-commands start` previously hardcoded `npx @dcl/hammurabi-server@next` as the multiplayer server. This makes the Rust bevy-headless engine (`@dcl-regenesislabs/bevy-headless-server`, decentraland/bevy-explorer#974) serve that role, and — since it is what production multiplayer now runs — makes it the **default**, so previews exercise the same engine:

- `DCL_SERVER_ENGINE=bevy|hammurabi` picks the implementation; unset now means `bevy`. Hammurabi remains fully supported as an explicit opt-out.
- `DCL_SERVER_PACKAGE=<dir|tarball|spec>` overrides the npx package spec — npx accepts a local directory, so an unpublished engine build is testable through the real spawn path.
- If the bevy launcher exits with code 78 (its "permanently unavailable here" signal: unsupported platform, missing binary), the preview **aborts loudly** — no silent hammurabi fallback that would hide broken bevy installs. The error states the reason (platform/arch) and the exact command to opt into hammurabi (`DCL_SERVER_ENGINE=hammurabi npm start`), and the `Multiplayer server exited` analytics event is flushed before exiting so unavailability shows up in telemetry.
- The module is renamed `hammurabi-server.ts` → `multiplayer-server.ts` (`startHammurabiServer` → `startMultiplayerServer`) to match, and the never-populated `PreviewComponents.hammurabiServer` field is dropped.
- **Readable preview output.** The bevy child's stdout/stderr are piped through a line filter: every engine line gets a green `[Server]` tag and a compact local `HH:MM:SS` time; the Rust tracing prefix, ANSI codes, the `[headless] alive` heartbeat, and known headless-build noise (dot-prefixed gltf lookups, `$ipfs` path misses, GizmoPlugin/embedded-asset renderer warnings) are stripped; level is conveyed by color (yellow warnings, red errors) instead of `LOG`/`WARN`/`ERROR` words. The engine also gets a default `RUST_LOG=warn,scene_runner::renderer_context=info` (an explicit `RUST_LOG` wins), a child killed by an unexpected signal (SIGSEGV/OOM) is now surfaced with a warning, and the preview CLI's own logger defaults to `LOG_LEVEL=INFO` so the embedded comms server's per-connection DEBUG chatter stays out (`LOG_LEVEL=DEBUG` re-enables it). Hammurabi's output path is untouched.

- **Hot-reload ws contract test.** The preview file watcher broadcasts both the protobuf binary frame and the legacy text frames (`"update"` + `SCENE_UPDATE` JSON); bevy explorers — the headless server included — speak only the legacy text protocol (see decentraland/bevy-explorer#1083). A jest spec now pins both protocols and their binary-first ordering so the legacy frames can't be dropped silently.

Validated end to end against towerofmadness: `npm start` spawned the Rust engine, the scene ran as SERVER, a client synced (`[Game] Connected to server`), and a full 420s round cycle completed.

## What could break

- **The default engine changes for everyone**: a plain `npm start` in an authoritative scene now installs and runs `@dcl-regenesislabs/bevy-headless-server@next` instead of hammurabi. Anyone who needs the old behavior sets `DCL_SERVER_ENGINE=hammurabi`.
- **On platforms where the bevy binary can't run, the preview now exits with code 78** instead of continuing with a hammurabi fallback. This is deliberate (fail loud, get reports), but it means such users must explicitly set `DCL_SERVER_ENGINE=hammurabi` to keep working.
- The engine is consumed via the npm `latest` dist-tag, which only moves when someone dispatches bevy-explorer's publish-headless workflow with `dist_tag=latest` (decentraland/bevy-explorer#1082) — engine pushes to main land on `next` and can't change previews. Note `latest` currently points at an older snapshot; run a `latest` release after that PR lands.

## How to test

```bash
# in an authoritative scene project:
npm start
# expect:
#   Starting Multiplayer Server (bevy) with realm: http://localhost:8000
#   [Game] Running as SERVER
#   [Server] Ready
pgrep -fl "headless --realm"   # child is the Rust engine, not node

# explicit opt-out spawns hammurabi exactly as before:
DCL_SERVER_ENGINE=hammurabi npm start

# a local unpublished engine build through the real spawn path:
DCL_SERVER_PACKAGE=/path/to/bevy-explorer/deploy/headless/launcher npm start

# unavailability path: point DCL_SERVER_PACKAGE at a build that exits 78 and
# check the preview aborts, printing the platform reason and the hammurabi command
```

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/start/hammurabi-server.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/hammurabi-server.ts
@@ -9,7 +9,7 @@ import { isElectronEnvironment, getSpawnEnv, findNpxCliJs, getNpxBin } from './u
 const HAMMURABI_PACKAGE = '@dcl/hammurabi-server'
 const HAMMURABI_VERSION = 'next'
 
-const BEVY_PACKAGE = '@dcl/bevy-headless-server'
+const BEVY_PACKAGE = '@dcl-regenesislabs/bevy-headless-server'
 const BEVY_VERSION = 'next'
 
 // The bevy server exits with this when it can never run here (unsupported platform,
@@ -21,6 +21,16 @@ type ServerEngine = 'bevy' | 'hammurabi'
 function selectedEngine(): ServerEngine {
   const requested = process.env.DCL_SERVER_ENGINE
   return requested === 'bevy' || requested === 'hammurabi' ? requested : 'hammurabi'
+}
+
+/**
+ * npx accepts a directory or tarball as well as a registry spec, so pointing
+ * DCL_SERVER_PACKAGE at a local build exercises this spawn path without publishing.
+ */
+function packageSpec(engine: ServerEngine): string {
+  const override = process.env.DCL_SERVER_PACKAGE
+  if (override) return override
+  return engine === 'bevy' ? `${BEVY_PACKAGE}@${BEVY_VERSION}` : `${HAMMURABI_PACKAGE}@${HAMMURABI_VERSION}`
 }
 
 /**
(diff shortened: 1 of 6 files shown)


pravusjif (approved):
LGTM, also tested it in several scenes and they keep working as expected, including storage persistence

PR #1529, opened 2026-08-06
feat: content-versioned preview ids for local scene development

dalkia:
## What does this PR change?

In preview (`npm run start` / Creator Hub), a file's content id is derived from its absolute path only (`b64-<base64(path-machineId)>`). The id never changes when the file changes, so an explorer cannot tell from the content map whether anything is stale — its only correct option is to drop its caches wholesale on every scene reload and re-download/re-process the entire scene, even for a one-file edit.

This PR embeds the file's **mtime** into the preview id:

```
b64-<base64(path \u0000 truncatedMtimeMs -machineId)>
```

- Every URL and client cache key derived from the id becomes **self-invalidating**: an edited file gets a new id (cache miss → fresh bytes), unchanged files keep their warm caches.
- The separator is a NUL byte, which cannot occur in file paths or hostnames, so decoding is unambiguous (paths with spaces, hyphens or digits are safe) — and clients can detect a version-capable server by its presence.
- The **project directory's own id (the scene/entity id) stays path-only**, so scene identity remains stable across edits and reloads.
- The decoder accepts both plain and versioned ids, so nothing that still mints plain ids breaks.

## Why

Explorer-side scene reload currently spends most of its time re-downloading and re-parsing unchanged assets because of the forced cache drain. With versioned ids the explorer can skip the drain entirely and only re-fetch what actually changed (unity-explorer counterpart: [decentraland/unity-explorer#9819](https://github.com/decentraland/unity-explorer/pull/9819)). Old explorers are unaffected: they treat ids as opaque strings and keep draining as before.

## Notes / follow-ups

- A `.gltf`/`.glb` with *external* textures keeps its own id when only a texture changes; a client that caches the composed model must still handle that dependency (or a follow-up could fold dependency mtimes into the model's id). GLBs with embedded textures — the common export — are fully covered.
- Wearable preview projects keep plain ids for the wearable id derivation (`b64HashingFunction(workingDirectory)`), which is unaffected.

## Test it

```bash
npm run start # in any scene project
curl -s -X POST http://127.0.0.1:8000/content/entities/active -d '{"pointers":["0,0"]}' -H 'Content-Type: application/json'
# content[].hash changes when you touch a file; entity id does not
```


kuruk-mm (approved):
great work!

PR #1532, opened 2026-08-10
chore: warning-free, vulnerability-free install; jest 30 without ts-jest

eordano:
## What

`npm install` on this repo greeted every contributor with **22 vulnerabilities (2 critical, 10 high)**, 5 engine warnings, and (on a fresh resolve) 13 deprecation warnings. After this PR: **0 vulnerabilities, 0 engine warnings, and a single residual deprecation** that cannot be fixed from this repo (see below).

| | before | after |
|---|---|---|
| audit vulnerabilities | 22 (2 critical) | **0** |
| EBADENGINE warnings | 5 | **0** |
| deprecation warnings | 13 | **1** (`string-similarity`, see below) |

## How

- **Security**: `overrides` pin patched transitives — `protobufjs@^7.6.3` (critical prototype pollution, reached via `@dcl/protocol`/`ts-proto`), `semver`, `minimatch`, `zod`, `ajv`, `undici`, `test-exclude`; plus `npm audit fix` for the rest. Lockfile regenerated.
- **Engines**: `"node": ">=20.0.0"` instead of exactly `24.16.0` — the exact pin warned on every node that wasn't byte-identical (CI keeps 24.16.0 via `.node-version`).
- **Deprecations**: jest 30 (glob@7/inflight chain), glob 13 (only non-deprecated line), rimraf 6, syncpack 13, `@dcl/eslint-config` 3.
- **ts-jest removed**: tests compile through babel-jest (bundled with jest) + `@babel/preset-typescript`. Jest 30's removed matcher aliases are codemodded (`toBeCalledWith` → `toHaveBeenCalledWith` etc., 25 files). Type-only imports that ts-jest elided through type-checking are now explicit `import type` (they pulled raw `.ts` from `@dcl/protocol/out-ts` into the runtime under babel).
- **eslint 9**: `.eslintrc.json` + `.eslintignore` fold into `eslint.config.mjs` via FlatCompat (the shared config still publishes eslintrc shape).
- **Snapshots regenerated**: protobufjs 7.2.4→7.6.3 legitimately shifts bundle sizes and VM stats.

## Verification

Full suite collects **1097 tests (943 before** — suites that previously died at import now run). The only failures are the pre-existing ones that fail identically on `main`: the `init` blackbox network flake, the `help over all commands` timeout, and the ethereum-fixture cross-suite interference (each passes in isolation). `make build` green end to end, including proto generation.

## The one residual warning

[...]

PR #1534, opened 2026-08-11
fix: address CLI regressions from dependency slimming (#1532)

gonpombo8:
## Summary

Post-merge review of #1532 (dependency slimming) surfaced several regressions in the published CLI. This PR fixes them:

- **Node runtime floor was raised silently.** `fs.globSync` (used in `bundle.ts`, `composite.ts`, `code-commenter.ts`) requires Node ≥22, but the package declared no `engines` field, so Node 20 users got a hard runtime crash on every build with no install-time warning. Now declared: `"engines": { "node": ">=22.0.0" }`.
- **Phantom `glob` imports.** `glob` was removed from dependencies but `src/logic/project-files.ts` still imported it at runtime — `deploy`/`export-static` only worked because a transitive copy (via archiver) happened to hoist, which breaks under pnpm/strict resolution. `getPublishableFiles` is now migrated to `fs.globSync`, finishing the removal #1532 started. The old `dot: false` / `ignore` glob options were belt-and-braces: the `ignore`-package filter right after it is the authoritative `.dclignore` gate (gitignore semantics, applies `.*`/`node_modules`/etc. at any depth), so the glob call only needs to enumerate files — directories are filtered with a stat check (the old `nodir: true`), and dot-dirs/node_modules are pruned during the walk for performance. Old and new implementations produce byte-identical file lists on a fixture covering dotfiles, nested dot-dirs, node_modules, empty dirs, and every default-ignored extension. The repo-root scripts/tests (`scripts/helpers.ts`, `test/snapshots.spec.ts`, `test/ecs/composite.spec.ts`, `test/ecs/add-entity-from-composite.spec.ts`) had the same phantom import and were migrated the same way (root already pins Node ≥24.16).
- **Preview server lost its stable port.** `tryListen(0)` hands out a random OS ephemeral port on every run; portfinder used to search upward from 8000, so `sdk-commands start` without `--port` always served on a predictable URL. Restored the 8000-based upward scan.
- **Colors decided by the wrong stream.** All logger output goes to stderr, but `util.styleText`'s default validation consults `process.stdout` — piping stdout (`sdk-commands start | tee log`) or CI silently stripped all styling despite `FORCE_COLOR`. Now passes `validateStream: false`; the env-var contract (`FORCE_COLOR`/`NO_COLOR`) is the single source of truth again, as it was with colorette.
- **Ctrl+C during prompts hung on Node <24.** The readline-based `prompts` shim registered no `SIGINT` handling, and on Node 20–23 readline just pauses stdin — the prompt hung forever and `onCancel` never ran. Now aborts the pending question via `AbortController` on `SIGINT`.
- **Windows never opened the browser.** `spawn('cmd', ['/c', 'start', '""', url])` gets mangled by libuv's CRT-style quoting (the `'""'` title arg becomes `\"\"`, which cmd doesn't unescape). Now passes the command line verbatim (`windowsVerbatimArguments: true`) with the URL quoted.
- **The protobufjs security fix never reached sdk-commands' own install.** The root `overrides` block doesn't apply to sdk-commands' standalone install (this monorepo is not npm workspaces), so its lockfile still resolved protobufjs 7.2.4 — the exact critical advisory #1532 set out to eliminate. The override is now declared in sdk-commands' own `package.json`, so `make build`, CI, and development installs in the package dir resolve 7.6.x (the nested 6.11.6 copies collapse too). Caveat: npm only honors `overrides` at the install root, so scene creators installing the published CLI still resolve @dcl/protocol's exact `protobufjs@7.2.4` pin — fully fixing the consumer side needs a @dcl/protocol release that stops exact-pinning 7.2.4 (follow-up, tracked below).

Known issues from the same review that this PR does **not** address:

- Scene bundles grew ~38KB (+6.7%) because protobufjs 7.6.x inlines the `buffer`/`long` shims into every compiled scene (the `REQUIRE: buffer`/`REQUIRE: long` lines vanished from the snapshots). Recovering the old size likely needs esbuild `alias`/`external` configuration — worth its own PR.
- The ts-jest → babel-jest swap means the test suite is no longer type-checked (babel strips annotations without checking). A `tsc --noEmit` pass over `test/` would need the test tsconfig cleaned up first.
- End users installing the published CLI still get protobufjs 7.2.4 via @dcl/protocol's exact pin (see caveat above) — needs a @dcl/protocol bump.

## What could break

- The port scan starts at 8000 again: anyone who started relying on the random-port behavior from #1532 (unlikely in the ~1 day it was on main) gets 8000 back.
- `engines` produces an npm `EBADENGINE` warning for consumers on Node <22 (npm default is warn, not fail). That's the intent — today those installs crash at runtime instead.
- sdk-commands' `package-lock.json` was regenerated to pick up the restored `glob` dep and the protobufjs override, so the lockfile diff is large but mechanical.
- Colored output now appears in logs captured from stderr in CI (that was colorette's behavior before #1532; `NO_COLOR=1` opts out).

## How to test

1. `make install && make build && make test`
2. Preview port: `cd <any scene> && npx sdk-commands start` → serves on `http://127.0.0.1:8000` again (run twice; port is stable).
3. Deploy globbing: `npx sdk-commands deploy --skip-validations` in a scene with a `.dclignore` → file list matches pre-#1532 (the export-static blackbox test exercises the same path).
4. Prompts cancel (Node 20/22): run `npx sdk-commands init` in a non-empty directory and hit Ctrl+C at the confirm prompt → exits via onCancel instead of hanging.
5. protobufjs: `cd packages/@dcl/sdk-commands && npm ls protobufjs` → 7.6.x, and `npm audit` no longer reports the critical advisory.
6. Windows: `npx sdk-commands start` opens the default browser.

### DIFF
--- a/packages/@dcl/sdk-commands/src/logic/get-free-port.ts
+++ b/packages/@dcl/sdk-commands/src/logic/get-free-port.ts
@@ -1,5 +1,9 @@
 import * as net from 'net'
 
+// search upward from 8000 (portfinder's contract) so the preview URL stays stable across runs
+const BASE_PORT = 8000
+const HIGHEST_PORT = 65535
+
 function tryListen(port: number): Promise<number> {
   return new Promise((resolve, reject) => {
     const server = net.createServer()
@@ -16,11 +20,14 @@ export async function getPort(port: number, failoverPort = 2044) {
   const resolvedPort = port && Number.isInteger(port) ? +port : 0
 
   if (!resolvedPort) {
-    try {
-      return await tryListen(0)
-    } catch (e) {
-      return failoverPort
+    for (let candidate = BASE_PORT; candidate <= HIGHEST_PORT; candidate++) {
+      try {
+        return await tryListen(candidate)
+      } catch {
+        // busy, try the next one
+      }
     }
+    return failoverPort
   }
 
   return resolvedPort
--- a/packages/@dcl/sdk-commands/src/logic/open.ts
+++ b/packages/@dcl/sdk-commands/src/logic/open.ts
@@ -1,4 +1,4 @@
-import { spawn } from 'child_process'
+import { spawn, SpawnOptions } from 'child_process'
 import { release } from 'os'
 
 function isWsl(): boolean {
@@ -10,14 +10,25 @@ export default async function open(target: string): Promise<void> {
   if (protocol !== 'http:' && protocol !== 'https:') {
     throw new Error(`Refusing to open non-http(s) URL: ${target}`)
   }
-  const [command, args] =
-    process.platform === 'darwin'
-      ? ['open', [target]]
-      : process.platform === 'win32' || isWsl()
-        ? // `start` treats the first quoted arg as a window title; `&` must be escaped for cmd
-          [isWsl() ? 'cmd.exe' : 'cmd', ['/c', 'start', '""', target.replace(/&/g, '^&')]]
-        : ['xdg-open', [target]]
-  const child = spawn(command, args, { stdio: 'ignore', detached: true })
+  const spawnOptions: SpawnOptions = { stdio: 'ignore', detached: true }
+  let command: string
+  let args: string[]
+  if (process.platform === 'darwin') {
+    command = 'open'
+    args = [target]
+  } else if (process.platform === 'win32') {
+    // libuv's CRT-style quoting mangles the '""' title arg, so hand cmd the exact line
+    command = 'cmd'
+    args = ['/c', `start "" "${target}"`]
+    spawnOptions.windowsVerbatimArguments = true
+  } else if (isWsl()) {
+    command = 'cmd.exe'
+    args = ['/c', 'start', '""', target.replace(/&/g, '^&')]
+  } else {
+    command = 'xdg-open'
+    args = [target]
+  }
+  const child = spawn(command, args, spawnOptions)
   // opening the browser is best-effort: without this handler a missing
   // xdg-open/open/cmd (headless, containers) crashes the CLI
   child.on('error', () => {})
--- a/packages/@dcl/sdk-commands/src/logic/project-files.ts
+++ b/packages/@dcl/sdk-commands/src/logic/project-files.ts
@@ -1,7 +1,7 @@
 import { ContentMapping } from '@dcl/schemas/dist/misc/content-mapping'
 import { CliComponents } from '../components'
 import { getDCLIgnorePatterns } from './dcl-ignore'
-import { sync as globSync } from 'glob'
+import { globSync, statSync, Dirent } from 'fs'
 import ignore from 'ignore'
 import i18next from 'i18next'
 import os from 'os'
@@ -27,13 +27,15 @@ export async function getPublishableFiles(
   const ig = ignore().add(ignorePatterns)
   const allFiles = globSync('**/*', {
     cwd: projectRoot,
-    absolute: false,
-    dot: false,
-    ignore: ignorePatterns,
-    nodir: true
+    // prune walking into trees ig.filter below always excludes anyway (perf only;
+    // exclude receives a string on some Node versions and a Dirent on others)
+    exclude: (entry: string | Dirent) => {
+      const name = typeof entry === 'string' ? path.basename(entry) : entry.name
+      return name.startsWith('.') || name === 'node_modules'
+    }
   })
 
-  return ig.filter(allFiles)
+  return ig.filter(allFiles).filter((file) => statSync(resolve(projectRoot, file)).isFile())
 }
 
 /**
--- a/packages/@dcl/sdk-commands/package-lock.json
+++ b/packages/@dcl/sdk-commands/package-lock.json
@@ -47,6 +47,9 @@
         "@types/archiver": "^5.3.2",
         "@types/qrcode": "^1.5.6",
         "@types/ws": "^8.5.4"
+      },
+      "engines": {
+        "node": ">=22.0.0"
       }
     },
     "../ecs": {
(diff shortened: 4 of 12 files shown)


eordano (approved):

packages/@dcl/sdk-commands/src/logic/project-files.ts
```diff
@@ -27,13 +27,15 @@ export async function getPublishableFiles(
...
+    // prune walking into trees ig.filter below always excludes anyway (perf only;
+    // exclude receives a string on some Node versions and a Dirent on others)
+    exclude: (entry: string | Dirent) => {
+      const name = typeof entry === 'string' ? path.basename(entry) : entry.name
+      return name.startsWith('.') || name === 'node_modules'
+    }
   })
 
-  return ig.filter(allFiles)
+  return ig.filter(allFiles).filter((file) => statSync(resolve(projectRoot, file)).isFile())
```
regenesis-claw, line 38: [P2] `statSync` follows symlinks and throws, so a dangling symlink anywhere under the project root (or a file removed between the glob walk and the stat) aborts `deploy`/`export-static` with an unhandled `ENOENT` instead of a CLI error message. Reproduced on Node 24 with a broken symlink in the scene root.

`throwIfNoEntry: false` turns that into a silent skip, which matches the intent ("only keep real files"):

```suggestion
  return ig.filter(allFiles).filter((file) => statSync(resolve(projectRoot, file), { throwIfNoEntry: false })?.isFile() ?? false)
```

Alternatively `globSync('**/*', { withFileTypes: true, ... })` gives you `isFile()` straight off the `Dirent` and drops one stat syscall per entry — though the `exclude` callback typing gets a bit noisier, so the suggestion above is probably the better trade here.

packages/@dcl/sdk-commands/src/logic/get-free-port.ts
```diff
@@ -16,11 +20,14 @@ export async function getPort(port: number, failoverPort = 2044) {
...
 
   if (!resolvedPort) {
-    try {
-      return await tryListen(0)
-    } catch (e) {
-      return failoverPort
+    for (let candidate = BASE_PORT; candidate <= HIGHEST_PORT; candidate++) {
+      try {
+        return await tryListen(candidate)
+      } catch {
```
regenesis-claw, line 26: [P2] The bare `catch` treats every failure as "port busy". If `createServer`/`listen` fails for a reason that won't change on the next port (`EACCES` under a sandbox, `EMFILE`, a mocked `net` module), this walks all 57,536 candidates one `await` at a time before finally returning `failoverPort`. The existing `should return the fail-over port when probing fails` spec exercises exactly that path today.

Bailing out on anything that isn't an address conflict keeps the happy path identical and makes the failure fast:

```ts
} catch (e) {
  const code = (e as NodeJS.ErrnoException)?.code
  if (code !== 'EADDRINUSE') break
}
```

packages/@dcl/sdk-commands/src/logic/open.ts
```diff
@@ -10,14 +10,25 @@ export default async function open(target: string): Promise<void> {
...
+  const spawnOptions: SpawnOptions = { stdio: 'ignore', detached: true }
+  let command: string
+  let args: string[]
+  if (process.platform === 'darwin') {
+    command = 'open'
+    args = [target]
+  } else if (process.platform === 'win32') {
+    // libuv's CRT-style quoting mangles the '""' title arg, so hand cmd the exact line
+    command = 'cmd'
+    args = ['/c', `start "" "${target}"`]
```
regenesis-claw, line 22: [P2] With `windowsVerbatimArguments` the raw `target` string is interpolated straight into a cmd command line, so a `"` in the URL closes the quoting and leaves the remainder exposed to cmd metacharacter parsing. Not reachable today — both call sites build localhost/preview URLs — but you already parse the URL two lines above for the protocol check, and `new URL(target).href` percent-encodes `"` for free:

```ts
const { protocol, href } = new URL(target)
// ...
args = ['/c', `start "" "${href}"`]
```

Separately: the WSL branch keeps the old `['/c', 'start', '""', ...]` form. I assume that's deliberate (libuv doesn't apply CRT-style quoting when the host is POSIX, so the mangling you're fixing doesn't happen there) — a one-line comment saying so would stop the next reader from "fixing" it to match the win32 branch.

regenesis-claw (approved):
## Review — CLI regressions from dependency slimming

Followed the repo's agent/contributor conventions (`CLAUDE.md` → `AGENTS.md`; no `REVIEW.md` in this repo). Reviewed the full non-lockfile diff line by line and skimmed the lockfile churn (mechanical: nested `glob@9.3.5` copy dropped/hoisted, protobufjs 7.2.4 → 7.6.x, nested `protobufjs@6.11.6` copies collapsed — no new packages introduced).

**Verdict: approve.** Every fix here is a genuine regression repair, the reasoning in the description matches what the code does, and CI is green across build/test/lint/docs/E2E. Findings below are all P2 — none block the merge.

### Consumer impact

Two consumer-visible surface changes, both intentional and documented:
- `engines: >=22.0.0` on `@dcl/sdk-commands` → npm `EBADENGINE` **warning** (not failure) for scene creators on Node <22. Strictly better than today's runtime crash on `fs.globSync`.
- Preview server default port returns to the 8000 upward scan (pre-#1532 contract).

No exported symbol changes: `getPort`, `open`, `getPublishableFiles` and `prompts` are all internal to the CLI binary (`bin` entry only, not part of any package's public exports), so no cross-repo consumer check applies beyond the two items above.

### Security review

No secrets, no auth/authz surface, no unsafe deserialization, no new network I/O. The protobufjs override genuinely removes the critical advisory from this repo's own installs, and the "npm only honors overrides at the install root" caveat is honestly stated rather than glossed over. One low-severity quoting note on the Windows `open` path is inline below (not reachable from current call sites — both callers build localhost/preview URLs).

### P2 findings

1. `statSync().isFile()` throws on dangling symlinks — inline comment (reproduced locally on Node 24).
2. Port scan swallows every error class, not just "port busy" — inline comment.
3. Windows command-line interpolation of the raw `target` — inline comment.
4. **Engines floor is slightly optimistic.** `util.styleText`'s options object (`validateStream`) only exists from Node 22.8; on 22.0–22.7 the third argument is ignored (harmless — those versions don't validate the stream either, so colors still work), and `fs.globSync` was experimental for most of the 22.x line. `>=22.8.0` would describe the real floor more accurately. Related: `@dcl/sdk` (the package scenes actually depend on) carries no `engines` field, so the `EBADENGINE` warning only surfaces via the transitive CLI — worth mirroring it there so the signal is unmissable at scene-install time.
5. **No regression tests for the two widest-blast-radius fixes.** The description says the old/new publishable-file lists were diffed byte-for-byte on a fixture covering dotfiles, nested dot-dirs, node_modules and empty dirs — that fixture is exactly the test that would have caught #1532 before release, and landing it (plus a `getPort(0) >= 8000` assertion) is cheap insurance against the same class of regression recurring.
6. **Description nit:** it says the lockfile "was regenerated to pick up the restored `glob` dep", but `glob` is correctly *not* back in `dependencies` — the lockfile only drops the nested copy and hoists archiver's transitive `9.3.5`. Worth rewording so a future reader doesn't go looking for a dependency that isn't there.

Also worth a sanity check before release: `fs.globSync` and npm `glob` may differ in path-separator output on Windows, and both the `ignore` filter and `ContentMapping` keys expect POSIX separators. I believe both return native separators (so no regression), but a Windows `deploy` smoke test would confirm — you're already touching Windows behavior in this PR.

Requested by Gon via Slack

PR #1543, opened 2026-08-13
feat: add openExplorerUiAndWait helper

popuz:
Follow-up to #1511, as announced there. `openExplorerUiAndWait` opens an explorer panel and resolves once the session that call started ends.

This is a rewrite of what this PR originally proposed (`openExplorerUiAndWaitClose`). decentraland/protocol#464 added a second result component rather than new fields, which showed that this surface grows by **adding whole components** — the first version could not absorb that, so the helper was redesigned around correlation instead of guessing. The design notes live in the commit messages of `ae7413b` and `6c60912`.

## What changes

- New `@dcl/sdk/explorer-ui` subpath module (file-layout entry, same mechanism as `platform/`), exporting `openExplorerUiAndWait` plus re-exports of `ExplorerUi` / `OpenExplorerUiResult`, so one import covers the whole flow. Like `platform/`, it is not part of the playground-assets rollup, so none of the helper's own symbols reach `api.md`.
- **Sessions are correlated, not guessed.** Every call mints a `request_id` (a counter starting at 1; the wire's `0` means uncorrelated) and the explorer echoes it on the events that call produced. Listeners are armed *before* the RPC, so an event that lands while it is in flight is already observed. That deletes the anchor, the per-`ui` FIFO, the consumption high-water mark, the pre-RPC timestamp snapshot and the replay loop of the first version.
- **Graceful on an explorer that does not echo yet.** An uncorrelated lifecycle event falls back to matching on `ui`, which is unambiguous while panels cannot coexist (a second call for the same panel is answered `WAS_ALREADY_OPEN`). Where that fallback cannot be trusted — a panel that can coexist with another, or a channel with no `ui` field — the call rejects with an explicit message instead of attributing the event to the wrong session.
- **Channels are the extension point.** `channel(name, component)` wraps any grow-only result component, so a component added later needs no SDK release. `ExplorerUiEvents` and `ItemPurchase` ship ready-made. `collect` names the channels whose events are gathered, each tagged with its channel name; `until` (a channel, or `variant(channel, 'case')`) stops the wait early.
- **The panel closing is the terminal**, and it wins over `until`. `WaitOutcome` is `closed` | `matched` | `notOpened` | `timedOut` — the outcome is the outer axis, so `notOpened`/`timedOut` are written once and never change as channels are added. `$case` says *why the wait stopped*; `events` always holds everything collected, the stopping event included.
- **Non-`OPENED` verdicts resolve immediately** as `notOpened`: event delivery is owner-only, so a scene that did not open the panel never receives its `closed` — waiting would hang deterministically.
- **`timeoutMs` (no default) resolves with `$case: 'timedOut'`.** Implemented as a lazily added engine system (the scene runtime has no timers) that removes itself once no timed session remains. It stays optional because panels legitimately stay open for minutes; what omitting it costs is documented on the option.
- The promise never rejects on lifecycle grounds — only when an event cannot be attributed to any call, or when the RPC itself fails.
- **`ExplorerItemPurchaseResult` (1221) is declared grow-only** in `generateIndex.ts`. Without that allowlist entry it silently generates as LWW and same-tick events overwrite each other.
- **`code-to-composite` mock completed.** Its auto-mock returned an async function for any unknown property, which turned `OpenExplorerUiResult` into a function whose members are `undefined` — a scene using the helper hung forever while the command reported success. The mock now carries the real enum and answers `REJECTED_FEATURE_DISABLED`, matching the file's refuse-gracefully convention. Also fills in `copyToClipboard` and `stopEmote`.

## Protocol bump

`@dcl/protocol` moves to `1.0.0-34643020503.commit-6402953`, which carries decentraland/protocol#464. That is what adds the 69 lines to `api.md` — the new `PBExplorerItemPurchaseResult` message, the `EU_ITEM_PURCHASE` enum member and the `core::ExplorerItemPurchaseResult` entry in `componentDefinitionByName`. The regenerated `.crdt` snapshots in the diff come from the same bump.

## Verification

- `test/sdk/explorer-ui/openExplorerUiAndWait.spec.ts` — 24 tests, all green. Events are driven through the real incoming CRDT path (a transport feeding `APPEND_VALUE`), never `addValue`: the outgoing path hands `onChange` the whole set, which a scene never sees for a renderer-owned component. Tests read the minted id back from the mocked RPC rather than guessing it, and use a synthetic grow-only component to cover channel behaviour independently of 1221.
- Covered: id minting and RPC forwarding; foreign-id events ignored; two concurrent sessions with interleaved events; the uncorrelated `ui` fallback and both rejection paths; every non-`OPENED` verdict; the collected chain including its terminator; `until` by channel and by variant, including on a channel that is not collected; close beating `until`; `timedOut` with partial events and system self-removal; close beating a timeout that expires in the same tick; `timeoutMs: 0`; RPC failure; a re-wrapped channel not double-collecting; the name-conflict throw; events arriving while the RPC is in flight; the `undefined` delivered on `DELETE_ENTITY`.

## Trying it end to end

The helper needs an explorer that echoes `request_id`, which is decentraland/unity-explorer#10096. Against a client without it, the helper still works through the uncorrelated `ui` fallback, and only `EU_ITEM_PURCHASE` rejects.

decentraland/sdk7-test-scenes#98 exercises it live, next to a hand-rolled drain of the same component so both consumption styles stay covered. It is deployed, so no local build is needed: [jump in](http://decentraland.zone/jump/?dclenv=zone&realm=sdk7testscenes.dcl.eth&position=80,-6) and use the **Wait helper** row — the five cubes cover close, a collected chain, an early `until` match, a timeout, and a second call answered `WAS_ALREADY_OPEN`. Each outcome is printed to the scene's own HUD on the **right**; the **left** HUD shows the raw events underneath, and every event the helper consumes must still appear there.

## Merge order

This PR first, then decentraland/unity-explorer#10096 (independent of it), then decentraland/sdk7-test-scenes#98 — that scene is pinned to a branch build of `@dcl/sdk` and gets repinned to a published version once this lands.


packages/@dcl/sdk/src/explorer-ui/index.ts
```diff
@@ -0,0 +1,317 @@
+import {
+  engine,
+  ExplorerUi,
+  ExplorerUiEventsResult,
+  GrowOnlyValueSetComponentDefinition,
+  IEngine,
+  PBExplorerUiEventsResult
+} from '@dcl/ecs'
+import { openExplorerUi, OpenExplorerUiResult } from '~system/RestrictedActions'
```
gonpombo8, line 11: OpenExplorerUiResult
is this value availabe at runtime ? becuase all the ~systems modules are injected via explorer, is not a code pakage.

popuz: Yes, it is available at runtime.

You are right that `~system/*` is host injected and `@dcl/js-runtime` ships types only, so the real question is whether the explorer exposes the enum as a *value* and not just as a type. It does: `Explorer/Assets/StreamingAssets/Js/Modules/RestrictedActions.js` builds both enums through a `makeEnum` helper and exports them, with forward and reverse mappings. That shim landed in decentraland/unity-explorer#9472 (`8e1d2b62e`) and is on both `dev` and `main`.

Two caveats your question is right to surface:

1. It puts a floor under the client version. This is the only place in the SDK that imports a `~system` enum as a value rather than a type, so a scene using this helper needs an explorer that has #9472.

2. `code-to-composite` does not stub it. The mock at `packages/@dcl/sdk-commands/src/commands/code-to-composite/scene-executor.ts:205-214` stops at `triggerSceneEmote` and has no `openExplorerUi` or `OpenExplorerUiResult`, so a scene using the helper throws there while the command still reports success. I will fix that.

Heads up on the PR itself: I am moving it to draft and rewriting the helper. decentraland/protocol#464 adds a `request_id` that correlates a request with the events it produces, which removes the timestamp window and FIFO matching this version relies on. Your question stays relevant either way, since the new version still reads the verdict as a runtime value.

popuz:
Thanks — going through the list.

**P0, build failure on `PBExplorerItemPurchaseResult`.** Correct and known: decentraland/protocol#464 is still open. This PR stays in draft until it lands, and the pin will move to a published `@dcl/protocol@experimental` before merge, never a branch tarball.

### Applied (`6c60912`)

- **`for...in` → `Object.keys` in `caseOf()`.** Taken. The objects are plain today so this changes nothing at runtime, but the intent reads better and it costs nothing.
- **`timeoutMs` and the risk of hanging.** Documented rather than enforced. The "mandatory escape hatch" wording you quoted is from the old PR body, which described the previous helper — the body has now been rewritten. Making `timeoutMs` required would be wrong: a panel legitimately stays open for minutes, so a required timeout forces every caller to invent a number and turns a normal long session into a `timedOut`. What omitting it costs is now stated on the option itself.
- **Same-tick timeout + close, and `timeoutMs: 0`.** Both added as tests. One correction to the reasoning: it is not the `MIN_SAFE_INTEGER` priority that makes close win. `Engine.update` awaits `crdtSystem.receiveMessages()` before it runs *any* system, so the close settles the session in the same tick no matter what the timeout system's priority is. The lowest priority is there for a different reason — it lets the system remove itself mid-tick by splicing the last element of the array the engine is iterating. The test now pins the behaviour and the comment states the real cause.
- **PR title.** Fixed, along with the body — both still described `openExplorerUiAndWaitClose`, which the redesign in `ae7413b` deleted.

### Not applied

- **Channel/Variant as premature abstraction.** This is the one thing the redesign exists for. The first version of this helper was written against one result component and could not absorb a second one; protocol#464 added `PBExplorerItemPurchaseResult` as a whole new component rather than new fields, and that is the shape future additions will take. `channel()` is public precisely so the next component does not need an SDK release to be usable — the two shipped constants are one line each on top of it, not the reason it exists.

  On `openPurchaseAndWait(urn)` specifically: per-panel convenience wrappers were considered and dropped. They multiply with the panels, each one re-declares the same options, and a caller with a dynamically computed `ui` cannot use them at all — which is what the second overload of `openExplorerUiAndWait` covers instead.

- **`// SAFETY:` comments on each `as never[]`.** This repo's house style is to comment only where the code cannot carry the meaning itself. The invariant behind the casts is stated once, at the single point where the public signature and the implementation part ways; repeating it at four call sites adds volume without adding information, and comments that restate an invariant tend to drift out of sync with it.

pravusjif (approved):
LGTM

PR #1545, opened 2026-08-17
fix(sdk-commands): serialize local storage writes to prevent server-storage.json corruption

cyaiox:
## Bug Description

The local preview/storage dev server (`@dcl/sdk-commands start`) persists all runtime state — `env`, `world`, and `players` — to a single `server-storage.json` via a whole-file read-modify-write. Concurrent writes were unguarded, so the file could be silently corrupted or lose data.

**Expected:** concurrent storage upserts (e.g. a scene issuing several `set()` calls on load) all persist, and the file always stays valid JSON.

**Actual:** overlapping requests interleaved on Node's event loop — later saves clobbered earlier ones (lost updates), overlapping writes could interleave into invalid JSON (which the loader then discarded, resetting everything to defaults), and default (no-file) loads aliased shared objects so state leaked between unrelated reads.

## Root Cause

Every mutator did `loadServerStorage()` → mutate in memory → `saveServerStorage()` with no serialization. Because each step `await`s, two in-flight requests interleave at those yield points:

- **Lost updates:** both handlers load the same snapshot; the later save overwrites the other's change.
- **Byte corruption:** `saveServerStorage` wrote directly to the real file (no temp+rename), so two concurrent writes could interleave, and a crash mid-write left a truncated file.
- **State leak:** `DEFAULT_STORAGE` was a shared const; `{ ...DEFAULT_STORAGE }` shallow-copied it, so every default load shared the same nested `env`/`world`/`players` objects.

## Type of Change

- [x] Bug fix

## Fix Description

- Route every `load → mutate → save` cycle through a single in-process FIFO queue (`serialize()`), so writes apply in issue order and never race.
- Make `saveServerStorage` atomic: write a temp file, then `rename` over the target — a crash mid-write can no longer leave an unparseable file.
- Replace the shared `DEFAULT_STORAGE` const with a `createDefaultStorage()` factory so default loads no longer alias each other.

## How to Reproduce (Before Fix)

1. `sdk-commands start` a scene that issues several `player`/`world`/`env` `set()` calls on load (or drive concurrent `storage ... set --target http://localhost:<port>`).
2. Inspect `<@dcl/sdk-commands>/.runtime-data/server-storage.json`.
3. Observe missing keys (lost updates) or, on unlucky interleavings, invalid JSON that gets reset to defaults.

## How to Verify (After Fix)

1. `node_modules/.bin/jest --forceExit --testPathPatterns='test/sdk-commands/commands/start/runtime-env'` — the new concurrency/atomicity/isolation tests pass.
2. Repeat the reproduction: all values persist and the file stays valid JSON.
3. `node_modules/.bin/jest --forceExit --testPathPatterns='test/sdk-commands/commands/start'` — full start suite green (33/33).

## Impact Assessment

- **Severity:** Medium
- **Users affected:** Subset — developers using the local preview/storage server with concurrent or bursty writes.
- **Duration:** Present since the local storage server was introduced.

## Regression Risk

- Behavior/API unchanged — same functions, return types, and HTTP endpoints; purely a robustness change.
- Writes are now serialized in-process; a single hung executor would delay subsequent writes (bounded to local dev use).

-

## Checklist

- [x] Root cause identified and documented above
- [x] Fix addresses the root cause (not just symptoms)
- [x] Added test to prevent regression
- [x] Existing tests pass locally
- [x] Tested the specific reproduction steps
- [x] No new warnings or console errors introduced

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,6 +89,7 @@ Asset-packs injection is gated behind `isEditorScene` (requires `assets/scene/ma
 
 - A newer local npm rewrites committed lockfiles with `"peer": true` / `"dev": true` metadata churn during installs (root and per-package lockfiles). Revert lockfile changes unrelated to your dependency change before committing.
 - `make format` runs prettier over the whole repo, including paths CI's `make lint` does not check (`test/`, `scripts/`, dot-files), where HEAD may carry drift — a blind `make format` can dirty dozens of unrelated files. Check your own files, revert the rest.
+- **PR base branch:** this repo runs long-lived integration branches (e.g. `auth-server`) that are far ahead of `main`. Open a PR against the branch your feature branch was cut from, not `main` — targeting `main` sweeps in hundreds of unrelated files. Find the real base with `git branch -a --contains HEAD~` or by checking which branch gives a clean `git diff <base>...HEAD`; when working from `auth-server`, point the PR at `auth-server`.
 
 ## Code conventions
 
--- a/packages/@dcl/sdk-commands/src/commands/start/server/runtime-env.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/server/runtime-env.ts
@@ -16,10 +16,29 @@ export interface ServerStorage {
   players: Record<string, Record<string, unknown>>
 }
 
-const DEFAULT_STORAGE: ServerStorage = {
-  env: {},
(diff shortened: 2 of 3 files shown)


pravusjif (approved):
Tested with 2 local test scenes for multiplayer server and seem to work OK

PR #1567, opened 2026-09-01
fix(ecs): reject CRDT messages shorter than their header

LautaroPetaccio:
## What / why

A CRDT chunk is a sequence of length-framed messages, and the parser trusted the frame in three ways it should not. Every one is reachable from a peer, because the chunk arrives straight from a transport's `onmessage`: in `@dcl/sdk` that is the sync transport `@dcl/sdk/network` installs, with nothing checking the bytes in between.

**A length below the header hung the parser.** When the reader meets a message type it does not know, it skips it by advancing the cursor by the header's declared length. Nothing checked that length against the 8-byte header the protocol documents as its minimum, so a message declaring `length = 0` never moved the cursor and `parseChunkMessage` re-read the same bytes forever. The scene thread is single-threaded, so the freeze is total.

**A known type was read without checking the frame held its fields.** Known types were dispatched straight to their readers, which call `readUint32()` per field and `readBuffer()` for the data. A header-only `DELETE_ENTITY`, or a `PUT_COMPONENT` announcing more data than its frame carries, made the byte buffer throw `Outside of the bounds of writen data.` out of `onmessage`, aborting the tick that received the chunk.

**A known type could over-declare its length.** Each typed reader consumes only its own fields, so a frame claiming more than it holds left the cursor short of the declared boundary and the loop read the trailing bytes as another header. An over-declared `DELETE_ENTITY` followed by a valid update dropped that update silently.

None of these needs a hostile peer. Zero padding on a send buffer *is* a zero-length header, and a truncated write *is* a short frame.

## How it works

The first guard goes in `CrdtMessageProtocol.validate`, which every framing reader funnels through: a declared length below the header size is refused.

The rest goes in `readMessage`, which now owns the type dispatch. A table built from each reader's existing `MESSAGE_HEADER_LENGTH` says how many bytes that reader consumes and whether a data length follows, and a frame is accepted only if the header declares it **exactly** — header plus fixed fields, plus the announced data length for a data message. `parseChunkMessage` skips anything refused by its declared length, exactly as it already skipped an unknown type, so a valid message behind a bad one still arrives.

Two consequences worth stating outright.

- **The legacy network delete is a deliberate exception.** Released SDKs declare `DELETE_ENTITY_NETWORK` four bytes shorter than the sixteen they write, so both lengths are accepted and the chunk is still required to hold the sixteen. The cost is real: a frame that declares twelve and holds only twelve is indistinguishable from a legacy peer's write, so the reader takes the legacy reading, consumes four bytes of what follows, and because the next length field is then a type value below eight, everything left in that chunk is dropped. Peers on released SDKs are real and their deletes have to keep working, while a twelve-byte frame is malformed either way. It retires itself as those peers pick up #1595, which is already on `main`.
- **Trusting the declared length cuts both ways.** A frame that over-declares and really carries the padding it claims now resynchronises correctly, where before the cursor landed mid-frame. A frame whose header lies, with a valid message sitting where the padding should be, is now skipped past. Both are malformed input; the trade favours the honest one.

This guards framing only. A well-formed frame whose payload is too short for its component is #1597, which is queued immediately after and verified to stack on this with no source conflict.

## How to test

```
node_modules/.bin/jest --colors --forceExit --testPathPatterns='test/ecs/crdt-malformed-message'
```

All 41 cases pass here. Against `main` the reader-level zero-length cases fail outright, and the two transport cases for it hang the worker instead of failing, which is the defect itself, so that run has to be killed rather than reporting a count.

## Proof

`test/ecs/crdt-malformed-message.spec.ts` covers the framing reader directly and through a transport: a zero length, a length below and exactly at the header, an unknown type, a header-only `DELETE_ENTITY`, `PUT_COMPONENT` and `APPEND_VALUE` shorter than their fixed fields, a `PUT_COMPONENT` announcing more data than its frame holds, over-declared fixed and data frames, the legacy `DELETE_ENTITY_NETWORK` at both lengths including the dropped remainder, and, through a transport, a malformed frame followed by a valid update that must still be applied. `message.ts` is at 100% across statements, branches, functions and lines.

A scene reproducing both in-world symptoms is at [`crdt-zero-length-hang`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-zero-length-hang): it hands the parser a control chunk, then a header-only entity delete inside a `try/catch`, then the zero-length chunk. On `@dcl/sdk@7.26.0` the readout records the error that escaped the transport and then never updates again; here it prints `FIXED` with a climbing frame counter. Its README carries headless measurements for all four chunks against 7.26.0 and against this branch.


LautaroPetaccio:
Addressed the P1 in b6a6038c.

`readMessage` now owns the type dispatch and refuses a frame that cannot hold the fields its type's reader will consume. The minimums come from each reader's existing `MESSAGE_HEADER_LENGTH`, plus the announced data length for `PUT_COMPONENT`, `PUT_COMPONENT_NETWORK` and `APPEND_VALUE`. `parseChunkMessage` skips a refused message by its declared length, the same way it skips an unknown type, so a valid message behind it is still applied. `DELETE_ENTITY_NETWORK` keeps accepting the shorter length released SDKs declare for it (writer corrected in #1595), with the chunk still required to hold the sixteen bytes.

New cases in `test/ecs/crdt-malformed-message.spec.ts`: header-only `DELETE_ENTITY`, `PUT_COMPONENT` and `APPEND_VALUE` below their fixed fields, `PUT_COMPONENT` announcing more data than its frame holds, the legacy-length network delete both ways, and a transport-level header-only delete followed by an update that must still land. 12 of the 27 cases fail against the previous commit; `message.ts` is at 100%.

One caveat, noted in the body: this guards framing only. A well-formed `PUT_COMPONENT` whose payload is too short for its component still throws from the component deserializer inside `receiveMessages`, so `engine.update` rejects. That is a separate pre-existing path and I'd rather guard `updateFromCrdt` in its own PR than widen this one.

LautaroPetaccio:
Fixed in 94869009. Confirmed the effect first: an over-declared `DELETE_ENTITY` (header claims 20, real frame 12) followed by a valid `PUT` at offset 20 dropped the PUT silently — the reader consumed 12, the loop read the 8 padding bytes as a header, hit the zero-length guard, and stopped.

I took the first of your two options — exact declared lengths — rather than advancing the loop to `offset + header.length`. It is the smaller change (only `holdsBody`, the parse loop is untouched) and it drops the malformed frame whole instead of applying it and stripping padding, which is safer for hostile input. Advancing to the declared end also collides with the one message that legitimately under-declares, `DELETE_ENTITY_NETWORK`, and forces a decision about whether padding rides along on re-broadcast; rejecting the frame avoids both.

`holdsBody` now requires `header.length` to equal the exact frame — header + fixed fields, plus the announced data length for data messages. A mismatch is skipped by the declared length, landing on the next message. The legacy `DELETE_ENTITY_NETWORK` exception is explicit in the reader table: it accepts its correct 16 and the legacy 12, named as such. New tests: an over-declared fixed frame, an over-declared data frame, and a transport-level over-declared delete followed by a valid update that still applies.

pravusjif (approved):
Nice find!

cyaiox (approved):

LautaroPetaccio:
Checked against `auth-server`: this patch no longer applies there (the engine's parse loop was rewritten and network message types are now skipped rather than read), but the defect is present and **more exposed** — `validate` still lacks the `messageLength < CRDT_MESSAGE_HEADER_LENGTH` check, and every network type now reaches the skip-by-declared-length branch that hangs. Running the repro there froze the test runner until I killed it.

Ported as **#1611**.

LautaroPetaccio:
Closing in favour of the `auth-server` port, **#1611**.

To be explicit about what that means, since this one was approved: the defect is real on `main` too — `validate` accepts a declared length of zero, and readers skip by that length, so the parse loop re-reads the same bytes forever and one 8-byte packet from any peer freezes the scene. Closing this does not fix that on `main`; it relies on `auth-server` merging before the next release, at which point #1611's guard arrives with it.

The port is not a straight copy: on that branch every network message type now reaches the skip-by-declared-length path, so it is more exposed there, and the reader table needed `AUTHORITATIVE_PUT_COMPONENT` added.

PR #1593, opened 2026-09-02
fix(ecs): read components defined from a primitive or array schema

LautaroPetaccio:
## What / why

A component can be defined straight from a schema that is not a `Map`:

```ts
const Counter = engine.defineComponentFromSchema('counter', Schemas.Int)
```

Those are broken through the public API in four ways, all from the same two habits.

**Truthiness instead of presence.** `get`, `getOrNull` and `create` test the stored value:

```ts
const component = data.get(entity)
if (!component) {
  throw new Error(`[getFrom] Component ${componentName} for entity #${entity} not found`)
}
```

so a component holding `0`, `false` or `''` reports itself missing, and `create` over such a value silently succeeds a second time instead of throwing.

**Spreading whatever it is given.** `deepReadonly` was constrained to an object and did `Object.freeze({ ...val })`, which discards anything that is not one.

| stored | read back on `main` |
| --- | --- |
| `5` | `{}` |
| `0` | `null`, reported missing |
| `false` | `null`, reported missing |
| `[1, 2, 3]` | `{"0": 1, "1": 2, "2": 3}` |

The fix asks the map whether it holds the entity, and lets `deepReadonly` pass primitives through and copy arrays as arrays. Its type parameter is no longer constrained to `Record<string, unknown>`, which was the assumption that caused this; it is `@internal` and the generated `playground-assets.api.md` does not change.

Reading an absent component still throws from `get` and still returns `null` from `getOrNull`, which the new spec pins down.

## How to test

```
node_modules/.bin/jest --colors --forceExit --testPathPatterns='test/ecs|test/react-ecs'
```

On `main` seven of the nine cases in the new file fail, one per row of the table above plus the double `create`. On this branch all 125 suites and 872 tests pass.

## Proof

`test/ecs/primitive-schema-components.spec.ts` is new and covers a number component with a zero default and a set value, a boolean holding `false`, an array component, and the absent case that must keep behaving as it does.

`readonly.ts` is at 100% statements, branches, functions and lines. Snapshots were regenerated, sizes and allocation counters only.

`readonly.ts` fails `prettier --check` and reports the same six eslint problems on `main` and on this branch alike, over the conditional-type indentation neither touches.

No scene: every symptom is a value read back in scene code, which a spec states more precisely than an in-world readout could.


pravusjif (changes requested):
~with this PR these built-in fields now flip from mutable to readonly typing:~
~- AvatarShape.wearables, AvatarShape.emotes~
~- AvatarEquippedData.wearableUrns, AvatarEquippedData.emoteUrns~
~- UiDropdown.options~
~- AvatarModifierArea.excludeIds~
~- AssetLoad.assets~
~- MeshRenderer box and plane uvs, UiBackground.uvs~

~and except for `AvatarEquippedData` all the other components should have those fields as mutable, not readonly so they would all break.~

~That needs to be addressed.~

Edit: it would only be for `get` or `getOrNull` so it's fine.

pravusjif (approved):

PR #1597, opened 2026-09-02
fix(ecs): drop a component update whose payload the schema cannot read

LautaroPetaccio:
## What / why

A CRDT `PUT_COMPONENT` or `APPEND_VALUE` carries a component's serialized value as its payload. The frame around it can be well-formed — header, fields and the announced data length all agree — while the payload itself is shorter than the component's schema needs. The message reads fine; the **deserialize** the engine runs next walks off the end of the buffer and throws `Outside of the bounds of writen data.`

That deserialize happens in `crdtSceneSystem.receiveMessages`, which runs at the very start of every `engine.update`, so one malformed payload aborted the whole tick: every other message in the same batch was lost, and the runtime driving the scene saw its `update` reject.

Worse, the damage was not limited to the dropped message. For a network message announcing a `networkId` this engine had not seen, `receiveMessages` allocated a local entity and wrote its `NetworkEntity` mapping **before** the payload was validated. When the payload then failed, the entity and mapping stayed behind, so a peer varying the `networkId` accumulated ghost entities one per malformed message.

The chunk arrives straight from a transport's `onmessage`. In `@dcl/sdk` that is the sync transport `@dcl/sdk/network` installs, so the bytes come from another player in the room. A peer whose schema for a shared component differs, or that simply truncates a write, could take down the tick for everyone who received it.

## How it works

The payload is checked at the top of the loop, before anything mutates. For `PUT_COMPONENT`, `PUT_COMPONENT_NETWORK` and `APPEND_VALUE` — the three types that carry one — the component's own `schema` deserializes the buffer, and a payload it cannot read is dropped there. No entity is allocated, no component is touched, and nothing is relayed to other transports.

Because the payload is validated up front, the apply path can trust its input, so there is no broad `try/catch` around `updateFromCrdt`. `fixTransformParent` and `updateFromCrdt` only throw on a payload that fails the same check, so anything else that surfaces is a real bug rather than something swallowed.

The last-write-wins component also deserializes into a local before mutating its timestamp or data, so a failure leaves its state untouched rather than half-updated. The grow-only component already deserialized before appending.

A dropped message advances nothing, so a later well-formed update for the same entity at the same lamport timestamp is still a first write and still wins.

## Merge order

This is queued to merge **second, straight after #1567**. The two were verified to stack: with #1567 applied first, rebasing this on top produces no source conflict — they edit the same file in different places, the parse loop there and the receive path here — and the combined tree typechecks with both suites passing together.

The guards are complementary, not redundant: #1567 rejects a malformed *frame*, this rejects a payload the component's schema cannot read inside a well-formed frame. Because both regenerate the QuickJS goldens, this PR will conflict on the twelve `.crdt` files the moment #1567 lands. That is a single `UPDATE_SNAPSHOTS=true` regeneration with no code change.

## How to test

```
node_modules/.bin/jest --colors --forceExit --testPathPatterns='test/ecs/crdt-malformed-payload'
```

All 14 cases pass here; against `main` the tick rejects instead.

## Proof

`test/ecs/crdt-malformed-payload.spec.ts` drives everything through a transport and a real `engine.update`:

- a too-short `PUT_COMPONENT` for a Transform — `update` resolves and the entity ends with no value;
- a malformed message followed in the same chunk by a well-formed one — the well-formed one still applies;
- a well-formed update after a dropped one for the same entity — it wins, proving the timestamp was not advanced;
- a malformed message on one of two transports — nothing is relayed to the other;
- a too-short `APPEND_VALUE` on a grow-only component — `update` resolves and the set stays empty;
- a malformed `PUT_COMPONENT_NETWORK` for an unseen `networkId` — no entity and no mapping are created and nothing is relayed, repeated ones accumulate no ghosts, and a well-formed network put still maps exactly one entity and applies its component.

A scene reproducing it in-world is at [`crdt-short-component-payload`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-short-component-payload): it hands a private engine a chunk holding a 4-byte Transform payload followed by a valid one, then calls `engine.update`. On `@dcl/sdk@7.26.0` the update rejects and the sign reads `BUG REPRODUCED`; here it resolves, the malformed entity has no Transform, the healthy one does, and the sign reads `FIXED`. Its README carries the headless before/after table.

### DIFF
--- a/packages/@dcl/ecs/src/engine/lww-element-set-component-definition.ts
+++ b/packages/@dcl/ecs/src/engine/lww-element-set-component-definition.ts
@@ -130,13 +130,17 @@ export function createUpdateLwwFromCrdt(
     switch (action) {
       case ProcessMessageResultType.StateUpdatedData:
       case ProcessMessageResultType.StateUpdatedTimestamp: {
-        timestamps.set(entity, msg.timestamp)
-
         if (msg.type === CrdtMessageType.PUT_COMPONENT || msg.type === CrdtMessageType.PUT_COMPONENT_NETWORK) {
-          const buf = new ReadWriteByteBuffer(msg.data!)
-          data.set(entity, schema.deserialize(buf))
+          // Deserialize before touching any state. A peer can send a payload this
+          // schema cannot read (a truncated buffer, say), and deserialize then throws.
+          // Reading it first leaves timestamps and data untouched on failure, so the
+          // caller can drop the message and a later well-formed update still wins.
+          const value = schema.deserialize(new ReadWriteByteBuffer(msg.data!))
+          timestamps.set(entity, msg.timestamp)
+          data.set(entity, value)
           lastSentData.set(entity, new Uint8Array(msg.data!))
         } else {
+          timestamps.set(entity, msg.timestamp)
           data.delete(entity)
           lastSentData.delete(entity)
         }
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -22,6 +22,22 @@ import {
 import { INetowrkEntityType } from '../../components/types'
 import * as networkUtils from '../../serialization/crdt/network/utils'
 
+/**
+ * Whether a component's schema can read this payload without running off the end. Used as
+ * a trust boundary for CRDT input from peers: a payload that fails here is dropped before
+ * it can allocate an entity or mutate a component. Deserialize has no side effects, so the
+ * value it produces is discarded and read again when the message is actually applied.
+ */
+function payloadIsReadable(component: ComponentDefinition<unknown>, data: Uint8Array | undefined): boolean {
+  if (!data) return false
+  try {
+    component.schema.deserialize(new ReadWriteByteBuffer(data))
+    return true
+  } catch {
+    return false
+  }
+}
+
 // NetworkMessages can only have a MAX_SIZE of 12kb. So we need to send it in chunks.
 export const LIVEKIT_MAX_SIZE = 12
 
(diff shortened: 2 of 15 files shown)


LautaroPetaccio:
Both fixed in 971442ce.

**P1.** Confirmed the leak first — a malformed `PUT_COMPONENT_NETWORK` for an unseen `networkId` left a ghost `NetworkEntity` mapping (entity 512 → `{networkId:123, entityId:7}`), one per message. I took the "defer allocation" option rather than rolling back in the catch, because a rollback via `removeEntity` would still burn the entity id permanently (removed ids are never reused). The payload is now validated at the top of the loop, before `addEntity`/`NetworkEntity.createOrReplace`, so a dropped message allocates nothing.

**P2.** With the payload validated up front by the component's own `schema.deserialize`, the broad `try/catch` is removed. `fixTransformParent` and `updateFromCrdt` only deserialize the same bytes I just checked, so they no longer throw on malformed input, and anything else that throws is a genuine bug that now surfaces instead of being swallowed. This narrows the handled path to schema/ByteBuffer failures without brittle error-message matching.

The check covers `PUT_COMPONENT`, `PUT_COMPONENT_NETWORK` and `APPEND_VALUE` — the three message types that carry a payload. New tests assert no entity/mapping/relay for a malformed network put, no ghost accumulation across several, and that a well-formed network put still maps exactly one entity and applies; three assertions fail against the previous commit.

pravusjif (approved):

cyaiox (approved):

LautaroPetaccio:
Checked against `auth-server`. Partly superseded there: the engine wraps `updateFromCrdt` in a try/catch, so the tick-abort symptom is contained. The concern moved to the server's validation path, which has three unguarded spots — `engine.getComponent` on a peer-chosen id, `schema.deserialize` on a peer-chosen payload, and a `null` from `convertNetworkToRegularMessage` passed straight into `validateMessagePermissions`. All three land in a DEBUG-gated catch, so the message is dropped silently *and* the sender never gets the correction an invalid message is meant to produce.

Ported as **#1612**.

LautaroPetaccio:
Closing in favour of the `auth-server` port, **#1612**.

Worth recording how the halves land, since the port is not a copy of this one. On `auth-server` the engine already wraps `updateFromCrdt` in a `try/catch`, so the tick-abort this PR prevents is covered there; `main` has no such catch (0 occurrences), so closing this leaves the abort live on `main` until `auth-server` merges and brings that catch with it.

#1612 covers the other half, which is new code that only exists on that branch: the server validator's unguarded `getComponent`, unguarded `deserialize`, and a `null` dereference — all swallowed by a DEBUG-gated catch, so the message was dropped silently *and* the sender never got the correction it is owed.

PR #1598, opened 2026-09-03
fix(ecs): survive running out of entities for network peers

LautaroPetaccio:
## What / why

A peer decides how many distinct `(networkId, entityId)` pairs it announces. `receiveMessages` allocates a local entity for every pair it has not seen before, and nothing bounds that:

```ts
if (networkUtils.isNetworkMessage(msg) && !network) {
  entityId = engine.addEntity()          // <- throws once the range is used up
  network = { entityId: msg.entityId, networkId: msg.networkId }
  NetworkEntity.createOrReplace(entityId, network)
}
```

Once the entity container's range is exhausted it throws `It fails trying to generate an entity out of range 65535.`, and that throw left `receiveMessages` and rejected `engine.update`. So a peer could stop the scene for everyone using **nothing but well-formed messages** — no malformed frame, no protocol violation, just variety.

The allocation is now attempted, and a failure drops that one message. An entity that cannot be represented locally has nothing to apply an update to, so there is nothing useful to do with the message. The report is latched, because a peer flooding unseen ids would otherwise flood the console too.

This was found while reviewing #1567's parser hardening for edge cases, and it is deliberately a separate PR: #1567 is about frames the reader cannot trust, this is about well-formed frames arriving in unbounded variety. It stands alone against `main` and does not depend on #1567.

## Two things this does not try to fix

- **The rate, only the crash.** A peer can still burn through the range and, once exhausted, its later entities are ignored for the session. Choosing a per-peer budget is a policy decision that belongs with whoever owns the comms trust model, and it needs somewhere to put the policy. What this PR guarantees is that exhaustion degrades instead of stopping the tick.
- **The cost of getting there.** `findNetworkId` scans every `NetworkEntity` per message, so the approach to exhaustion is O(n²) and in practice degrades into a freeze before the throw. Indexing that lookup is already open as #1475, and fixing it here would collide.

## How to test

```
node_modules/.bin/jest --colors --forceExit --testPathPatterns='network-entity-allocation-limit'
```

All four cases fail against `main`, each rejecting with `It fails trying to generate an entity out of range 65535.`; all four pass here.

## Proof

`test/ecs/network-entity-allocation-limit.spec.ts` drives it through a real transport: forty `PUT_COMPONENT_NETWORK` messages with distinct pairs into an engine whose container is built with `reservedStaticEntities: 65530`, so exhaustion arrives in forty messages instead of the ~65k (~4.7 MB) a real peer would have to send. It asserts `engine.update` resolves, that the engine mapped as many entities as it had room for and no more, that it says so exactly once, and that the next tick still runs.

**No repro scene for this one.** Reaching it in-world needs a second participant deliberately announcing tens of thousands of distinct network ids, which a scene cannot do to itself: the sync transport derives the pair from the local entity, so a scene has no way to forge the variety. The regression drives the same `receiveMessages` path a peer's messages arrive on.

### DIFF
--- a/packages/@dcl/ecs/src/engine/entity.ts
+++ b/packages/@dcl/ecs/src/engine/entity.ts
@@ -108,6 +108,13 @@ export type IEntityContainer = {
  * read one number; this runs on every inbound CRDT message. `entity & MAX_U16` already lands
  * in [0, 65535], so the `>>> 0` fromEntityId applies is a no-op here.
  */
+/**
+ * Thrown when the entity range is used up and no number can be recycled. Distinguishable
+ * from an allocator bug so callers that handle exhaustion do not also swallow one.
+ * @public
+ */
+export class EntityRangeExhaustedError extends Error {}
+
 function isReservedEntity(entity: Entity, reservedStaticEntities: number): boolean {
   return (entity & MAX_U16) < reservedStaticEntities
 }
@@ -128,7 +135,7 @@ export function createEntityContainer(opts?: { reservedStaticEntities: number })
 
   function generateNewEntity(): Entity {
     if (entityCounter > MAX_ENTITY_NUMBER - 1) {
-      throw new Error(`It fails trying to generate an entity out of range ${MAX_ENTITY_NUMBER}.`)
+      throw new EntityRangeExhaustedError(`It fails trying to generate an entity out of range ${MAX_ENTITY_NUMBER}.`)
     }
 
     const entityNumber = entityCounter++
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -1,4 +1,4 @@
-import { Entity, EntityState } from '../../engine/entity'
+import { Entity, EntityRangeExhaustedError, EntityState } from '../../engine/entity'
 import type { ComponentDefinition } from '../../engine'
 import type { PreEngine } from '../../engine/types'
 import { ReadWriteByteBuffer } from '../../serialization/ByteBuffer'
@@ -50,7 +50,8 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
   const receivedMessages: ReceiveMessage[] = []
   // Messages already processed by the engine but that we need to broadcast to other transports.
   const broadcastMessages: ReceiveMessage[] = []
-  // Latched so a peer flooding unseen network ids cannot also flood the console.
+  // Latched so a peer flooding unseen network ids cannot also flood the console. Cleared
+  // once an allocation succeeds again, so a later exhaustion is still reported.
   let reportedEntityExhaustion = false
 
   /**
@@ -144,20 +145,19 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
       let { entityId, network } = findNetworkId(msg)
       // We receive a new Entity. Create the localEntity and map it to the NetworkEntity component
       if (networkUtils.isNetworkMessage(msg) && !network) {
-        // A peer decides how many distinct (networkId, entityId) pairs it announces, and
-        // each unseen pair takes a local entity here. The container throws once its range
-        // is exhausted, and that throw used to leave `receiveMessages` and reject
-        // `engine.update`, so a peer could stop the scene with nothing but well-formed
-        // messages. Run out quietly instead and drop the message: the entity cannot be
-        // represented locally, so there is nothing to apply it to.
+        // A peer chooses how many distinct pairs it announces, so this allocation is
+        // driven by remote input and can exhaust the range.
         try {
           entityId = engine.addEntity()
+          reportedEntityExhaustion = false
         } catch (err) {
+          if (!(err instanceof EntityRangeExhaustedError)) throw err
+
           if (!reportedEntityExhaustion) {
             reportedEntityExhaustion = true
             console.error(
               'Ran out of local entities while mapping entities announced over the network. ' +
-                'Further network entities will be ignored for the rest of this session.',
+                'These entities are ignored until entities are freed and numbers can be recycled.',
               err
             )
           }
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -50,6 +50,8 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
   const receivedMessages: ReceiveMessage[] = []
   // Messages already processed by the engine but that we need to broadcast to other transports.
   const broadcastMessages: ReceiveMessage[] = []
+  // Latched so a peer flooding unseen network ids cannot also flood the console.
+  let reportedEntityExhaustion = false
 
   /**
    *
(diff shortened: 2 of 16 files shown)


LautaroPetaccio:
Addressed in 3e3d51be. Seven of the P2s are fixed; three are deliberately left.

**Generic catch.** `generateNewEntity` now throws `EntityRangeExhaustedError` and `receiveMessages` matches on it, rethrowing anything else. Your point about the latch was the sharper half: a *different* error arriving later would have been reported as "ran out of entities" and then hidden. That class is exported, so the API report is regenerated in the same commit.

**The log message was wrong, and you were right about why.** I initially believed exhaustion was permanent, because `generateNewEntity` advances a counter that removals never lower. That reasoning was incomplete: `generateEntity` recycles a freed number with a bumped version first, so allocation resumes once entities are released. I verified it — after freeing three entities and draining, `addEntity()` returned a recycled id. So the message now says "until entities are freed and numbers can be recycled", and the latch clears on a successful allocation so a later exhaustion is reported instead of swallowed. There is a regression for exactly that cycle.

**Test gap for already-mapped entities.** Added. It sends a fresh update for a pair that *was* mapped and asserts the value applies, which pins that `continue` only fires inside `if (!network)`.

**Loose count assertion.** Tightened. Rather than hardcoding 5, the spec derives it: `SPARE_ENTITIES = 5`, `RESERVED = MAX_ENTITY_NUMBER - SPARE_ENTITIES`, and asserts `toBe(SPARE_ENTITIES)`, so the relationship is visible instead of a magic number.

**Global `console.error` count.** Now asserts content, `expect.stringContaining('Ran out of local entities')`, with the count assertion kept alongside as its own case so the latch is still proven.

**`DELETE_ENTITY_NETWORK` for an unseen pair after exhaustion.** Added, asserting the tick resolves and the report does not fire twice.

**Comment block.** Trimmed to one line. You are right that it read as change history; per REVIEW.md §1 the DoS story belongs in the description, where it already is.

Left as follow-ups, as you suggested:

- **Programmatic observability.** A callback alongside `onProcessEntityComponentChange` is the right shape, but it is public API surface and deserves its own decision rather than riding along here.
- **O(n²) `findNetworkId`.** #1475. Your residual note is the part worth carrying over: the scan cost outlives the attack, because mapped rows are never reclaimed even after the attacker disconnects. That is a stronger argument for prioritising it than the crash was.
- **Per-peer quota.** Needs somewhere to put the policy and an owner for the comms trust model.

Ten cases now, all failing against `main`.

LautaroPetaccio:
## Confirmed, and the practical threshold is far lower than exhaustion

Measured on `main`, driving the same `receiveMessages` path peer traffic arrives on.

**Any of the three network message types allocates.** The allocation happens before the message type is examined and before any validation, so `PUT_COMPONENT_NETWORK`, `DELETE_COMPONENT_NETWORK` and `DELETE_ENTITY_NETWORK` all create a local entity plus a `NetworkEntity` mapping for a pair the engine has not seen.

**Only two of them consume permanently.**

| Vector | Bytes/msg | Permanently consumes |
| --- | --- | --- |
| `DELETE_ENTITY_NETWORK` | 16 | No, cleaned and the number recycled |
| `DELETE_COMPONENT_NETWORK` | 24 | **Yes**, no payload required |
| `PUT_COMPONENT_NETWORK` | 72 | Yes, with a Transform payload |

600 entity-deletes left 0 mappings behind. The cheapest permanent vector is `DELETE_COMPONENT_NETWORK` at 24 bytes, which needs no payload at all.

**Exhaustion costs ~1.5 MB, but the scene dies at ~96 KB.** Because resolution scans every mapping, per-batch cost grows linearly with the count, so total cost is quadratic:

| Mapped entities | Cost of one 250-message batch |
| --- | --- |
| 250 | 21.4 ms |
| 1 000 | 83.8 ms |
| 2 500 | 227.7 ms |
| 4 000 | **374.8 ms** |

At 4 000 mapped entities a single batch costs 375 ms, roughly 22 frames of a 60 fps budget, and 4 000 entities is 96 KB of traffic and about 6% of the way to exhaustion. The scene is unplayable long before the crash this PR fixes.

**Exhaustion also disables the scene's own allocation.** After the range is used up, the scene's own `addEntity()` fails: 0 of 20 succeeded in my run. Mappings are never reclaimed, so this outlives the attacker's disconnection.

## The gap that keeps it open

`networkId` is derived from the sender's address, but on the wire it is just a field in the message and the receiver never checks it against who actually sent it. The comms layer does decode the sender: `decodeCommsMessage` returns it and `BinaryMessageBus`'s callback signature is `(value, sender)`. But `message-bus-sync.ts` drops it:

```ts
binaryMessageBus.on(CommsMessage.CRDT, (value) => {
  transport.onmessage!(value)   // sender decoded, then discarded
})
```

So the engine cannot attribute an inbound mapping to a peer. Any per-`networkId` budget is bypassed by minting new `networkId`s, and `players.onLeaveScene`, which already exists in this file and currently only logs, cannot be used to reclaim.

## Suggested direction

1. **Thread the sender through to the CRDT layer.** Nothing else is enforceable without it. It is already decoded one function away.
2. **Budget inbound-created entities per sender**, not per `networkId`.
3. **Reclaim on leave** using the existing `onLeaveScene` hook, so cost is bounded by concurrent peers rather than lifetime traffic.

This PR and #1475 remain the right floor: this one stops the crash, #1475 removes the quadratic term that makes 96 KB enough. Neither bounds what one peer may allocate, which is the item that still needs an owner for the comms trust model.

LautaroPetaccio:
Checked against `auth-server`: this patch can't apply — the engine branch it guards (`engine.addEntity()` on an unmapped network message) was deleted with the rest of the network-entity subsystem.

The concern relocated rather than disappeared. `findOrCreateNetworkEntity` in `packages/@dcl/sdk/src/network/server/index.ts` still calls `engine.addEntity()` for every unseen `(networkId, entityId)` from a peer, unbounded, and `entity.ts` is unchanged so it still throws on exhaustion. Two differences from `main`: `processServerMessages` catches per message, so the throw no longer aborts anything — but the log is `DEBUG_NETWORK_MESSAGES`-gated, so exhaustion is **silent** — and it's now the *server* that runs out, which takes the scene down for everyone rather than one client.

**I did not port it**, because the useful part here isn't the typed error — it's bounding the allocation, and that's a per-sender budget, i.e. a policy decision rather than a mechanical port. `CreatedBy` already records which address caused each entity, so the attribution a budget needs is in place.

Happy to implement once you've picked the policy.

LautaroPetaccio:
Correction to my note above, after testing the client path properly.

The situation on `auth-server` is split, and the **unguarded half is the one a peer can reach** — the opposite of what I implied:

```
processClientMessages  -> THREW OUT: It fails trying to generate an entity out of range 65535.
processServerMessages  -> survived
```

The server survives only because of a generic per-message `try/catch`, and its log is `DEBUG_NETWORK_MESSAGES`-gated, so it degrades silently. The client has no `try/catch` at all and rejects `engine.update` outright, confirmed end to end. That path is peer-reachable: a flood at the server is allocated *and rebroadcast* to every client, which then allocates too.

Ported as **#1614**, which guards the allocation and reports exhaustion once at error level. Bounding the pool itself is still the policy decision and is not in it.

LautaroPetaccio:
Closing in favour of the `auth-server` port, **#1614**.

As with the others, stating the gap plainly: the exhaustion abort is real on `main` and closing this does not fix it there. It relies on `auth-server` merging before the next release — that branch deletes the engine path this PR guards entirely, and #1614 guards the allocation where it moved to instead.

The port also corrects what I said in my earlier comment here: on `auth-server` the **client** path is the unguarded one, and it is the one a peer can reach, since a flood at the server is allocated *and rebroadcast* to every client. Bounding the pool itself is still an open policy decision in neither PR.

PR #1607, opened 2026-09-10
fix: stacking order between UI renderers + zIndex

### LINKED ISSUE #1196: zIndex ignored for cross-renderer stacking when using ReactEcsRenderer.addUiRenderer()
## Summary

When multiple UI trees are registered via `ReactEcsRenderer.addUiRenderer()`, the `zIndex` property on `UiTransform` has no effect on the stacking order between renderers. Stacking is determined solely by registration order — the last renderer registered always renders on top, regardless of `zIndex` values.

## Steps to Reproduce

```typescript
// Register Feature A first
ReactEcsRenderer.addUiRenderer(() => (
  <UiEntity
    uiTransform={{ width: 200, height: 200, positionType: 'absolute', zIndex: 300 }}
    uiBackground={{ color: Color4.Blue() }}
  />
))

// Register Feature B second
ReactEcsRenderer.addUiRenderer(() => (
  <UiEntity
    uiTransform={{ width: 200, height: 200, positionType: 'absolute', zIndex: 50 }}
    uiBackground={{ color: Color4.Red() }}
  />
))
```

## Expected Behavior

The element with `zIndex: 300` (Feature A) should render in front of the element with `zIndex: 50` (Feature B).

## Actual Behavior

The element with `zIndex: 50` (Feature B) renders on top because it was registered later. The `zIndex` values are completely ignored for cross-renderer ordering.

## Impact

This makes it impossible to control the visual stacking order of UI elements from independent features that each use `addUiRenderer()`. The only workaround is to collapse all features into a single `setUiRenderer()` call, which breaks the modular design that `addUiRenderer()` was intended to enable.

## Suggested Fix

Either:
- Respect root-level `zIndex` values when compositing multiple renderer outputs, or
- Add an explicit `priority` / `zIndex` parameter to `addUiRenderer()` so callers can declare their intended stacking layer independently of registration order

---

Reported by Omar Bobrow
Requested by Omar Bobrow (<@U04F2UJ9726>)

pravusjif:
Fixes decentraland/sdk#1196.

## What / why

UI renderers registered with `addUiRenderer()` stacked by registration order only, and `zIndex` had no effect between them. Two causes on the SDK side:

- The reconciler never ordered top-level elements: its React container was a stub, so every renderer root was written with `rightOf: 0`. The Explorer sorts siblings by their `rightOf` chain and keeps only one unordered root, dropping the others from its sort.
- Each renderer is wrapped in its inset container, so an author's root `zIndex` sat one level below the entities that actually compete.

Changes:

- The reconciler container is a real instance for the canvas root, so top-level roots get the same `parent`/`rightOf` bookkeeping as any other siblings (append, insert-before and removal). A missing left sibling is now written as `rightOf: 0` everywhere instead of `undefined` — same wire meaning, canonical encoding.
- Renderers are ordered by the tick they first render, the main UI first within a tick — the order their entities were already created in, so scenes that set no `zIndex` keep the stacking they had.
- New `UiRendererOptions.zIndex`, applied to the renderer's root container. With `screenInset: 'none'` a whole-screen wrapper is added only when the option is given.

## Proof

Test scene `sdk7-test-scenes/scenes/0,6-ui-zindex-and-opacity`: three rectangles, one per `addUiRenderer()`, registered A → B → C with renderer `zIndex` 20 / 10 / 0.

- Before: C (registered last) is in front; the buttons that change `zIndex` do nothing.
- After: A is in front; the buttons reorder the rectangles by `zIndex`, and `0` falls back to registration order.

### QA TEST STEPS

Enter the ZONE/SEPOLIA world `sdk7testscenes.dcl.eth` at position 0,6 and confirm: that the top 3 colored rectangles render according to their zIndex calue (higher value, renders on top)

https://decentraland.zone/jump/?realm=sdk7testscenes.dcl.eth&position=0%2C6&dclenv=zone

Screenshot 2026-09-11 at 12 58 45 AM


DafGreco (approved):
✔️  PR reviewed and approved by QA on both platforms following instructions playing both happy and un-happy path

Regressions for this ticket had been performed in order to verify that the normal flow is working as expected:

Looking good!

Image
Image

cyaiox (approved):

PR #1608, opened 2026-09-10
fix: only let a peer create network entities in its own id space

LautaroPetaccio:
## What

A network CRDT message names its target entity with the pair `(networkId, entityId)`. When the receiving engine has no local entity mapped to that pair, `crdtSceneSystem` calls `engine.addEntity()` and creates one. It does this for whatever pair arrives, from whoever sends it.

`networkId` is written by the sending peer. Nothing checked it against who actually sent the message, so one peer could walk the 16-bit `entityId` range under another peer's `networkId` and drain the local entity pool — which is the same pool the scene's own `addEntity()` draws from. `DELETE_COMPONENT_NETWORK` costs 24 bytes on the wire and consumes an entity permanently, with no payload needed.

The fix uses the one field a peer cannot choose for itself. The comms runtime stamps the sender's address onto a chunk before the scene sees it (`craftCommsMessage` writes only `[type][payload]`; `decodeCommsMessage` reads a sender prefix, so the prefix is added in between). `BinaryMessageBus` already decoded it and threw it away one line before the engine.

So: transports that know the origin now hand it to the engine as a `TransportSender`, and **creating** an entity requires the message's `networkId` to be the sender's own.

## What is deliberately *not* restricted

**Writes to entities that already exist stay open to every peer.** `networkId` identifies whose id space an entity belongs to, not who is allowed to touch it — shared mutable state is the entire point of `syncEntity`, and whoever placed the door is not the only one allowed to open it. Binding writes to the sender breaks that outright (`test/sdk/network/sync-engines.spec.ts` catches it immediately).

**The `networkId: 0` shared namespace can never be created from the wire.** Those entities are the ones each client registers for itself with `syncEntity(entity, components, enumId)` during `main()`, so an unknown one is never something to take on faith. Without this the namespace would remain an open allocation vector.

**Join snapshots (`RES_CRDT_STATE`) are not validated.** One peer relays the whole scene there, other peers' entities included, so nothing in it can be attributed to the sender. Same trust model as before this PR.

**Transports with no notion of an origin pass no sender and behave exactly as before** — the renderer transport, local test transports, the editor.

## Known trade-off

If a peer's write to an entity reaches you *before* the message that creates it — possible when the two come from different senders and comms reorders across them — the write is now dropped instead of implicitly creating the entity. The entity is created when its owner's message lands, and CRDT state converges on the next update. This is the cost of not letting arbitrary peers mint entities; it is narrow, and worth naming.

## Still open after this

This bounds *who* may spend from an id space, not *how much*. A peer can still create up to ~65k entities under its own legitimate `networkId`. Because allocation is now attributable to a stamped address, a per-sender budget and reclaim-on-leave become possible to write — `players.onLeaveScene` already receives the same address. That is a policy decision and is not in this PR.

CRC32 is also not collision-resistant, so `networkId` should stay a routing key: any future quota or attribution should key on the stamped address string, which is exact.

## How to test

```bash
node_modules/.bin/jest --testPathPatterns='network-sender-validation'
```

`test/ecs/network-sender-validation.spec.ts` (12 cases) drives a transport directly. Five of them fail on `main` and pass here:

- a new entity announced in another peer's id space does not consume a local entity
- a chunk mixing an owned and an unowned entity creates only the owned one
- 200 unowned entities in one chunk allocate zero
- `DELETE_COMPONENT_NETWORK` / `DELETE_ENTITY_NETWORK` naming a non-existent foreign entity spend nothing

The other seven pin behaviour that must *not* change: creation in the peer's own space, writes to another peer's existing entity, writes to a locally registered shared-namespace entity, plain non-network messages, and transports that report no sender at all.

Full regression run:

```bash
node_modules/.bin/jest --testPathPatterns='test/(ecs|sdk|react-ecs|snapshots)'
```

1291 pass. `test/sdk/network/sync-engines.spec.ts` is the meaningful one — it runs two real engines through `addSyncTransport` and exercises cross-peer writes and deletes end to end.

Golden snapshots are regenerated (bundled SDK changed) and carry no `ERR!` lines. `playground-assets.api.md` is regenerated for the new `TransportSender` export and the `onmessage` signature.

### DIFF
--- a/packages/@dcl/ecs/src/runtime/types.ts
+++ b/packages/@dcl/ecs/src/runtime/types.ts
@@ -1,5 +1,5 @@
 export type { SystemFn, SYSTEMS_REGULAR_PRIORITY } from '../engine/systems'
-export type { TransportMessage, ReceiveMessage, Transport } from '../systems/crdt/types'
+export type { TransportMessage, ReceiveMessage, Transport, TransportSender } from '../systems/crdt/types'
 export { TransformType, TransformComponent } from '../components/manual/Transform'
 export * from '../engine/component'
 export * from '../schemas/typing'
--- a/packages/@dcl/ecs/src/systems/crdt/index.ts
+++ b/packages/@dcl/ecs/src/systems/crdt/index.ts
@@ -12,7 +12,7 @@ import { DeleteComponent } from '../../serialization/crdt/deleteComponent'
 import { DeleteEntity } from '../../serialization/crdt/deleteEntity'
 import { PutComponentOperation } from '../../serialization/crdt/putComponent'
 import { CrdtMessageType, CrdtMessageHeader, CrdtMessage } from '../../serialization/crdt/types'
-import { ReceiveMessage, Transport } from './types'
+import { ReceiveMessage, Transport, TransportSender } from './types'
 import { PutNetworkComponentOperation } from '../../serialization/crdt/network/putComponentNetwork'
 import {
   NetworkEntity as defineNetworkEntity,
@@ -24,6 +24,10 @@ import * as networkUtils from '../../serialization/crdt/network/utils'
 
 // NetworkMessages can only have a MAX_SIZE of 12kb. So we need to send it in chunks.
 export const LIVEKIT_MAX_SIZE = 12
+
+// Entities that every client announces under the same id share this networkId, so it
+// names a namespace instead of an owner.
+const SHARED_NETWORK_ID = 0
 
 /**
  * @public
@@ -62,7 +66,7 @@ export function crdtSceneSystem(engine: PreEngine, onProcessEntityComponentChang
      * Component Operation Messages at messages queue
      * @param chunkMessage A chunk of binary messages
      */
-    return function parseChunkMessage(chunkMessage: Uint8Array) {
+    return function parseChunkMessage(chunkMessage: Uint8Array, sender?: TransportSender) {
       const buffer = new ReadWriteByteBuffer(chunkMessage)
 
       let header: CrdtMessageHeader | null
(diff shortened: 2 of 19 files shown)


LautaroPetaccio:
Checked against `auth-server`: this patch can't apply — the engine no longer parses network messages at all.

The exposure splits in two there. **On clients it's gone**: `binaryMessageBus.on(CommsMessage.CRDT, ...)` only processes messages where `sender === AUTH_SERVER_PEER_ID`, so a peer can't reach another peer's engine.

**On the server it's still open, and already acknowledged.** `processServerMessages` creates entities from any peer's messages and records `CreatedBy { address: sender }` — exactly the attribution this PR adds — but `validateMessagePermissions` ends with:

```ts
// TODO: Check if sender owns the entity
// TODO: Check component sync mode ('all' | 'owner' | 'server')
```

So a peer can still claim another peer's `networkId`, and the server then rebroadcasts it authoritatively.

**I did not port it.** Implementing that TODO means choosing a permissions model — which components are owner-only, what a non-owner may do to a shared entity, how the sync mode is declared — and inventing one unilaterally seemed worse than leaving it visible. The mechanism is all there; it needs the policy.

Happy to implement once that's settled.

LautaroPetaccio:
Closing: superseded by the `auth-server` branch.

**What is superseded.** This PR's mechanism — thread the runtime-stamped sender into the engine and require a created network entity's `networkId` to match it — targets code that no longer exists on `auth-server`. The engine there does not parse network messages at all: `findNetworkId` is gone, and `PUT_COMPONENT_NETWORK` / `DELETE_COMPONENT_NETWORK` / `DELETE_ENTITY_NETWORK` fall through to be skipped. The client-side exposure is closed by construction rather than by validation, because `binaryMessageBus.on(CommsMessage.CRDT, ...)` only processes messages whose sender is `AUTH_SERVER_PEER_ID`. A peer can no longer reach another peer's engine, so there is nothing left here to check.

**What is not, and where it went.** The same concern is live on the server. `processServerMessages` creates entities from any peer's messages and records `CreatedBy { address: sender }` — the attribution this PR was adding — but `validateMessagePermissions` still ends with:

```ts
// TODO: Check if sender owns the entity
// TODO: Check component sync mode ('all' | 'owner' | 'server')
```

so a peer can still claim another peer's `networkId` and have the server rebroadcast it authoritatively. That is a different fix in a different place, and it needs a permissions policy rather than a port of this one. It is not carried by any open PR, so it should be tracked separately.

**One caveat on closing.** The vulnerability is real on `main` and this PR fixes it there. Closing is right if `auth-server` is what ships; if `main` continues to ship first, the hole stays open until that branch lands.

PR #1609, opened 2026-09-10
fix(sdk-commands): keep the editor's own autosaves out of hot-reload notifications

nearnshaw:
## The use case

You are editing a scene in the **Creator Hub with the Bevy renderer**. You drag an entity with the gizmo, or change a material colour in the inspector. The change shows up in the viewport immediately (the inspector forwards its own edits to the running scene live). Then, about two seconds later, **the whole scene sometimes reloads**: everything blinks out, `main()` runs again, animations restart, the scene lands frozen. The same edit will reload the scene one time and not the next.

Nothing about that edit needs a reload. The reload is the scene-code **hot-reload** path firing by mistake, and this PR stops it at the only place that can tell the two cases apart.

## Why it happens

The editor's realm is `sdk-commands start --data-layer`. That one process hosts both the inspector's **data layer** (which autosaves the scene) and the **bundler** (which rebuilds on file changes). After one inspector edit:

1. The data layer autosaves `assets/scene/main.composite` and `entity-names.ts`.
2. The bundler's watcher matches `.composite`/`.ts`, waits 100 ms and rebuilds, rewriting `bin/index.js` and `main.crdt` (`logic/bundle.ts`).
3. The file-watch notifier sees those outputs (none are dclignored), waits 800 ms after the last one, and broadcasts `SCENE_UPDATE` (`server/file-watch-notifier.ts`).

The inspector treats `SCENE_UPDATE` as "the scene code changed" and reloads the scene through its Stop/reset path. Because the message carries no file name, the inspector's only defence is a heuristic: ignore `SCENE_UPDATE` if one of its own ECS edits happened in the last 1.5 s. Measured against a real Creator Hub scene (6.4 MB bundle), the message lands **1.32–1.33 s after the save** on an idle machine. That leaves under 200 ms of slack, so any CPU contention (the Bevy wasm engine and Electron are both busy during a drag) or a bigger bundle pushes it past the window and the scene reloads. Hence "sometimes".

The same guard has a second cost: a **real** code save made within 1.5 s of an inspector edit is dropped outright, not deferred.

## The fix

The inspector cannot distinguish its own autosave from an IDE edit, but this process can: it performed the write, and it knows what triggered each rebuild. So track the origin here and skip the notification for editor-originated writes.

- **`logic/editor-write-tracker.ts`** (new) records which absolute paths were last written by the editor, with a 10 s TTL so a hand edit of the same file later is still reported. `markRebuildOutputs(triggers, outputs)` gives the rebuild outputs the editor origin only when **every** file that triggered the rebuild was an editor write; a rebuild caused by any other change (an IDE save) clears the mark, so a code change made right after an autosave is still delivered.
- **`data-layer/fs.ts`** marks every file (and every folder it has to create, since chokidar reports those as their own events) that the data layer writes or removes. **Exception: anything under `src/`.** Source the editor writes there (UI Designer roots) changes the compiled scene and must hot-reload exactly like a hand edit.
- **`logic/bundle.ts`** collects the paths that triggered a rebuild and marks `bin/index.js` + `main.crdt` *before* rebuilding, so the notifier's watcher can never observe the write first.
- **`server/file-watch-notifier.ts`** drops an event whose path is an editor write before it reaches the 800 ms debounce.
- **`start/index.ts`** creates the tracker only with `--data-layer` and threads it to the three places above. Plain `sdk-commands start` (the preview) is unchanged.

## What still reloads, on purpose

- Editing `src/**` in an IDE or through the AI assistant (the notifier sees the source change and/or the user-attributed rebuild output).
- Replacing a model or other asset from the file system (not a data-layer write).
- A UI Designer edit written under `src/ui/` (excluded from marking, see above).

## Tests

- `test/sdk-commands/logic/editor-write-tracker.spec.ts`: mark / lookup / ancestor-folder match / TTL expiry / rebuild-origin rules (100% coverage, the `logic/` threshold).
- `test/sdk-commands/utils/data-layer-fs.spec.ts`: the fs adapter marks scene data and created folders, does not mark `src/`, and works without a tracker.
- `test/sdk-commands/commands/start/file-watch-notifier.spec.ts`: real chokidar watcher on a temp project. An editor write plus its rebuild outputs produce **no** `SCENE_UPDATE` within 1.5 s; an unmarked write still does; without a tracker behaviour is as before. This test is what caught the folder-creation case.

## Verifying in the Creator Hub

Point the app at this build of `@dcl/sdk-commands`, open a scene with the Bevy renderer, and drag an entity or change a material colour repeatedly: the viewport updates live and the scene never reloads. Edit `src/index.ts` in an IDE: the scene still hot-reloads.

Related inspector code: `packages/inspector/src/lib/renderer/bevy/register.ts` (`LOCAL_EDIT_QUIET_MS`) in creator-hub. Once this ships, that heuristic can be relaxed or removed in a follow-up.

### DIFF
--- a/packages/@dcl/sdk-commands/src/commands/build/index.ts
+++ b/packages/@dcl/sdk-commands/src/commands/build/index.ts
@@ -4,7 +4,7 @@ import { declareArgs } from '../../logic/args'
 import { installDependencies, needsDependencies, SceneProject, WearableProject } from '../../logic/project-validations'
 import { getBaseCoords } from '../../logic/scene-validations'
 import { b64HashingFunction } from '../../logic/project-files'
-import { bundleProject } from '../../logic/bundle'
+import { bundleProject, CompileOptions } from '../../logic/bundle'
 import { printCurrentProjectStarting } from '../../logic/beautiful-logs'
 import { getValidWorkspace } from '../../logic/workspace-validations'
 import { Result } from 'arg'
@@ -56,7 +56,11 @@ export async function main(options: Options) {
   }
 }
 
-export async function buildScene(options: Options, project: SceneProject | WearableProject) {
+export async function buildScene(
+  options: Options,
+  project: SceneProject | WearableProject,
+  bundleOptions: Pick<CompileOptions, 'writeTracker'> = {}
+) {
   const canInstall = !options.args['--skip-install']
 
   if (canInstall) {
@@ -76,7 +80,8 @@ export async function buildScene(options: Options, project: SceneProject | Weara
       production: !!options.args['--production'],
       emitDeclaration: !!options.args['--emitDeclaration'],
       ignoreComposite: !!options.args['--ignoreComposite'],
-      customEntryPoint: !!options.args['--customEntryPoint']
+      customEntryPoint: !!options.args['--customEntryPoint'],
+      ...bundleOptions
     },
     project.scene
   )
--- a/packages/@dcl/sdk-commands/src/commands/start/data-layer/fs.ts
+++ b/packages/@dcl/sdk-commands/src/commands/start/data-layer/fs.ts
@@ -1,6 +1,7 @@
 import { FileSystemInterface } from '@dcl/inspector'
 import path from 'path'
 import { CliComponents } from '../../../components'
+import { EditorWriteTracker } from '../../../logic/editor-write-tracker'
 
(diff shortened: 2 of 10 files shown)


nearnshaw:
We're fixing this in a Creator Hub PR

PR #1610, opened 2026-09-10
fix: only accept a CRDT state snapshot that answers an outstanding request

LautaroPetaccio:
## The problem

`RES_CRDT_STATE` hands a joining client the scene state from a peer. It was accepted from anyone, at any time, whether or not this client had asked for it.

The check that looks like authentication isn't one:

```ts
const { sender, data } = decodeCRDTState(value)
if (sender !== myProfile.userId) return
```

`encodeCRDTState(userId, chunk)` is called by the *responder* and fills that field with the address of whoever asked. So it means "is this addressed to me?", against a value that is public — the runtime stamps my address on every message I send, so anyone can write it into a payload.

The handler was also never gated. `stateIsSyncronized` is written inside it and read nowhere in it, so it stopped the retry loop and nothing else; the handler stayed live for the whole session. A peer who was never asked for anything could therefore inject arbitrary CRDT at any moment. Verified with an entity belonging to a *third party*, landing after the client's sync had already completed.

## What it does now

A snapshot is accepted only while a state request is outstanding. `requestState` opens the window before broadcasting and closes it when state arrives, when the retries give up, or when comms disconnects; a retry reopens it, so the three attempts behave as one window of roughly fifteen seconds. Outside it, nothing is applied.

`decodeCRDTState` returns `addressee`, because calling it `sender` is what made a recipient check look like authentication.

## Why every answer is still merged

Pinning to a single responder looks safer and isn't. LWW resolves on a sender-chosen `uint32`, so one peer writing `MAX_U32` beats every honest peer however many answer — narrowing the source buys nothing against that. It costs real robustness: a peer answering with an empty snapshot would decide the scene is empty, and the honest answer behind it would be refused. I built that version first and reverted it; the tests pin the behaviour so it doesn't come back by accident.

Bounding how far a timestamp may jump is what would make merged peer state genuinely safe, and that is a separate change.

## Scope

The window bounds *when* a snapshot is accepted, not what one may contain — whoever answers during it can still fabricate entities or send unbounded numbers of them. Contents cannot be validated per message: `engineToCrdt` dumps every `NetworkEntity` in the scene, so an answer legitimately carries other peers' entities and none of it is attributable to whoever sent it.

Two smaller notes. An answer arriving after the window closes is dropped; in practice chunks are emitted in one `sendBinary` batch and arrive together. And the join flow resolves `RealmInfo` and `PlayerIdentityData` from the engine it was handed rather than the global one, matching `definePlayerHelper` in this package — identical in production, and what makes the flow testable at all.

## How to test

```bash
node_modules/.bin/jest --testPathPatterns='crdt-state-request-window'
```

Seven cases driving the real join flow through `addSyncTransport`. Two fail on `main`: a snapshot arriving when nothing was requested, and one arriving after the window closed. The other five hold what must keep working — the first answer is applied, every chunk of one answer is applied, several peers' answers merge, an answer following an empty one still delivers the real state, and a snapshot addressed to a different player is ignored.

```bash
node_modules/.bin/jest --testPathPatterns='test/(ecs|sdk|react-ecs)/'   # 1064 pass
node_modules/.bin/jest --testPathPatterns='test/snapshots'              # 17 pass, no regeneration
```

`test/sdk/network/sync-engines.spec.ts` is the regression check that matters: two real engines through `addSyncTransport`, state handshake included. Goldens don't move (the sync transport is opt-in), there's no public API change, and `make lint` exits 0 with the same findings as `main`.

### DIFF
--- a/packages/@dcl/sdk/src/network/message-bus-sync.ts
+++ b/packages/@dcl/sdk/src/network/message-bus-sync.ts
@@ -1,4 +1,5 @@
-import { IEngine, Transport, RealmInfo, PlayerIdentityData } from '@dcl/ecs'
+import { IEngine, Transport } from '@dcl/ecs'
+import { RealmInfo as defineRealmInfo, PlayerIdentityData as definePlayerIdentityData } from '@dcl/ecs/dist/components'
 import { type SendBinaryRequest, type SendBinaryResponse } from '~system/CommunicationsController'
 
 import { syncFilter } from './filter'
@@ -22,6 +23,11 @@ export function addSyncTransport(
   const myProfile: IProfile = {} as IProfile
   fetchProfile(myProfile!, getUserData)
 
+  // Resolved from the engine we were handed rather than the global one, so the
+  // join flow belongs to that engine.
+  const RealmInfo = defineRealmInfo(engine)
+  const PlayerIdentityData = definePlayerIdentityData(engine)
+
   // Entity utils
   const entityDefinitions = entityUtils(engine, myProfile)
 
@@ -41,6 +47,9 @@ export function addSyncTransport(
 
   let stateIsSyncronized = false
   let transportInitialzed = false
+  // Whether a state request is outstanding. A snapshot only means anything as an
+  // answer to one; outside that window nobody asked for it.
+  let awaitingState = false
 
   // Add Sync Transport
   const transport: Transport = {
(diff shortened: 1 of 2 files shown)


LautaroPetaccio:
Checked against `auth-server`: **superseded there**, and by something stronger. That branch's handler pins acceptance to the authoritative server using the authenticated sender —

```ts
if (isServerAtom.getOrNull() || sender !== AUTH_SERVER_PEER_ID) return
```

— so the unrequested-injection channel this PR closes doesn't exist on it. No port needed.

While checking, I did find an adjacent bug: the two lines that clear the in-flight request run *before* that guard, so any peer can cancel a joining client's retry and leave it without state. Fixed in **#1613**.

PR #1611, opened 2026-09-11
fix(ecs): reject CRDT messages shorter than their header

LautaroPetaccio:
Reject CRDT frames shorter than their header and validate each known message type before reading its fields. This prevents zero-length frames from hanging the engine and truncated network frames from throwing during client/server message processing. Port of #1567 to `auth-server`; includes regression tests and updated snapshot goldens.

Legacy network deletes that declare 12 bytes but write 16 remain supported. A malformed 12-byte delete followed by another frame is indistinguishable from that legacy format; the reader may consume bytes from the following frame. Tests cover this compatibility limit.

## How to test

After `make install` and `make build`, run:

```bash
node_modules/.bin/jest --runInBand --testPathPatterns='crdt-malformed-message|crdtMessage.spec|malformed-network-frame'
```

The regression tests cover zero/undersized headers, truncated bodies, valid messages following malformed frames, and legacy network deletes. The zero/undersized-header assertions fail with the base implementation; all 51 focused tests pass with this change.

## Repro scene

[`crdt-zero-length-hang`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-zero-length-hang) — the scene for #1567, which this replaces.

It still reproduces here unchanged: the frame it injects is not a network message, so it takes the same parse loop on this branch, and `validate` lacks the same guard.


LautaroPetaccio:
Thanks — both P1s were real. Verified each before acting, and one of them lands on my test rather than the code.

**P1.1 — fixed in the latest push.** Confirmed exactly as described:

```
readMessages(8-byte DELETE_ENTITY_NETWORK) -> THREW: Outside of the bounds of writen data.
```

`readMessages` now goes through the same checked reader as the engine, so a frame is validated against the type it claims and skipped by its declared length when it does not hold. With that, nothing downstream can throw, which covers both ingresses rather than adding a catch to each.

One deliberate omission: `APPEND_VALUE` stays unhandled there even though the reader understands it, because `chunkCrdtMessages` re-emits whatever `readMessages` returns and would start carrying appends it currently drops. That is a behaviour change worth its own PR, not a rider on this one.

New tests in `test/sdk/network/server/malformed-network-frame.spec.ts`: header-only network frames through `readMessages`, `processClientMessages` and `processServerMessages`, a valid update surviving behind a malformed one, and a real `DeleteEntityNetwork.write()` frame followed by an update. Four of the five fail without the change.

**P1.2 — you are right about the writer, and right about my test.** `DeleteEntityNetwork.write` declares `CRDT_MESSAGE_HEADER_LENGTH + 4` and emits sixteen, and skipping a real one by its declaration loses the message behind it. My regression test used a crafted twelve-byte frame, where skipping twelve does land correctly, and then generalised from it — so its conclusion did not hold for writer output. The comment is corrected to say exactly what it covers, and the real-writer case is now pinned in the new SDK test.

The writer declaration itself is deliberately not changed here: that fix is coming across with the `main` changes, and doing it in both places would conflict. Worth noting the practical impact on this branch is limited, since network frames reach `readMessages` rather than the engine skip path — and `readMessages` reads them, so it advances the real sixteen either way. I confirmed the same loss occurs on pristine `auth-server`, so it is not a regression from this PR.

**CI — you were right and my description was wrong.** The twelve snapshot failures are real and caused by this branch's ECS change. My local run had `@dcl/protocol 1.0.0-34239802138.commit-e89d7fa` installed, pulled in while working on `main`, against a branch pinning `1.0.0-32863591307.commit-0b3d285`. That is what made a pristine checkout look drifted. With the pinned version installed, a plain build produces no churn at all, the goldens regenerate cleanly with no `ERR!` lines, and the full suite is green — including `Generated OnPointerDown ProtoBuf`, which I had also written off as environmental. Goldens are in this push and the description is corrected.

PR #1612, opened 2026-09-11
fix(sdk): refuse a peer component update the server cannot read

LautaroPetaccio:
Reject unreadable peer component updates and use the existing correction path instead of silently dropping them. Unknown component IDs, deserialization failures, and failed message conversion now return a validation refusal; a correction is sent when the server has component state to restore.

## How to test

After building the packages, run:

```bash
node_modules/.bin/jest --runInBand --testPathPatterns='unreadable-payload'
```

`test/sdk/network/server/unreadable-payload.spec.ts` accepts a valid Transform update, then sends a truncated payload and checks for an authoritative reply. That assertion fails against the original `auth-server` implementation and passes here. The suite also checks that malformed or unknown components are not relayed and valid updates still are.

Validation: all four focused tests pass.

## Repro scene

[`crdt-short-component-payload`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-short-component-payload) — the scene for #1597, which this replaces.

**It does not exercise this branch as written.** It drives a truncated `PUT_COMPONENT` into a private engine, and the engine here already wraps `updateFromCrdt` in a try/catch, so the scene reports FIXED:

```
truncated PUT_COMPONENT -> engine.update aborted: no
```

The live defect on this branch is one level up, in the server's validation path, so the scene needs retargeting at `createServerValidator` to demonstrate it.

**An `auth-server` variant now exists alongside it:** [`crdt-short-component-payload-authserver`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-short-component-payload-authserver), which drives `createServerValidator` rather than the engine and reproduces against the published head `7.28.1-34624845556.commit-7d0808b`. The original [`crdt-short-component-payload`](https://github.com/LautaroPetaccio/js-sdk-bug-repros/tree/main/crdt-short-component-payload) stays as the `main` reproduction.


LautaroPetaccio:
Both points were right, and the P2 one is the more uncomfortable of the two.

**P1 — fixed.** What a message can be judged on alone (is the component defined, does its payload deserialize) now runs *before* `findOrCreateNetworkEntity`, so a peer sending unique `(networkId, entityId)` pairs it knows will be refused no longer drains the range one refusal at a time.

One nuance in how it is gated: the pre-check only skips allocation when nothing is mapped for that pair yet. A pair the server already knows still takes the normal path, so its sender keeps getting the authoritative correction — refusing those outright would have traded this bug for the silent-drop one this PR set out to fix. New test sends fifty refused updates naming entities the server has never seen and asserts no mappings are left behind.

**P2 — correct, and my test was passing for the wrong reason.** `networkMessageToLocal` calls `fixTransformParent` only when `componentId === TransformComponentId`, and that deserializes, so a truncated Transform throws inside `convertNetworkToRegularMessage` and comes back as `null`. The case was exercising the null-conversion guard; the `schema.deserialize` guard it was written for was never reached. The A/B looked convincing because the null guard is also part of this PR.

It uses `GltfContainer` now — LWW, synchronised, no transform fixing — so the guard is actually covered.

Both new cases fail without the change.

PR #1616, opened 2026-09-11
fix(sdk): only the authoritative server answers a state request

LautaroPetaccio:
Part of the state-handshake issue. The response side is pinned to the authoritative server; the **request** side is not gated at all.

## The problem

```ts
binaryMessageBus.on(CommsMessage.REQ_CRDT_STATE, async (data, sender) => {
  const chunks = engineToCrdt(engine)
  ...
})
```

Every participant answers, clients included. But a client's answer is never consumed — the RES handler drops anything whose sender isn't `AUTH_SERVER_PEER_ID`, so the recipient throws the whole thing away.

That makes it an amplifier. `engineToCrdt` serialises the entire scene, and one 1-byte broadcast makes **every client in the room** do it simultaneously:

| synced entities | chunks | bytes | time |
| --- | --- | --- | --- |
| 200 | 3 | 30 KB | 4 ms |
| 1,000 | 13 | 148 KB | 21 ms |
| 3,000 | 38 | 445 KB | 44 ms |

Per client, per request, repeatable at will. Twenty clients in a 3,000-entity scene turn one request into roughly 9 MB emitted and ~0.9 s of aggregate CPU — for output nobody reads.

## What it does

Answers only when this participant is the authoritative server.

Nothing a recipient can observe changes, because those client answers were already being discarded. This only stops the work being done and the bytes being sent.

## One timing note

`isServerAtom` is populated asynchronously from `isServerFn`, so a request arriving before the runtime has answered is dropped. The requester retries every 2 s, which is the same path that already covers a server that hasn't finished starting — and in practice the atom resolves well before comms connects.

## How to test

```bash
node_modules/.bin/jest --testPathPatterns='state-request-responder'
```

Two cases, driving the real join flow through `addSyncTransport` on a connected `RealmInfo` with a synced entity present. One fails on `auth-server`: a client answers a peer's request. The other pins the behaviour that must survive — the server still answers with its state.

```bash
node_modules/.bin/jest --testPathPatterns='test/(ecs|sdk|react-ecs)/'   # 1089 pass
```

`Generated OnPointerDown ProtoBuf › should serialize/deserialize OnPointerUp` fails on a pristine `auth-server` checkout here too — the installed `@dcl/protocol` is newer than the committed artifacts. Unrelated, untouched.

### DIFF
--- a/packages/@dcl/sdk/src/network/message-bus-sync.ts
+++ b/packages/@dcl/sdk/src/network/message-bus-sync.ts
@@ -115,9 +115,7 @@ export function addSyncTransport(
 
   // Receive & Process CRDT_STATE
   binaryMessageBus.on(CommsMessage.REQ_CRDT_STATE, async (data, sender) => {
-    // The RES handler drops anything not from the authoritative server, so a client
-    // answering dumps its whole scene for a recipient that discards it. The atom
-    // resolves asynchronously; a request that beats it is covered by the retry.
+    // Requests arriving before role resolution are recovered by the requester's retry.
     if (!isServerAtom.getOrNull()) return
 
     DEBUG_NETWORK_MESSAGES() && console.log('[REQ_CRDT_STATE]', sender, Date.now())
--- a/packages/@dcl/sdk/src/network/message-bus-sync.ts
+++ b/packages/@dcl/sdk/src/network/message-bus-sync.ts
@@ -115,6 +115,11 @@ export function addSyncTransport(
 
   // Receive & Process CRDT_STATE
   binaryMessageBus.on(CommsMessage.REQ_CRDT_STATE, async (data, sender) => {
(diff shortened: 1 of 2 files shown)


packages/@dcl/sdk/src/network/message-bus-sync.ts
```diff
@@ -115,6 +115,9 @@ export function addSyncTransport(
 
   // Receive & Process CRDT_STATE
   binaryMessageBus.on(CommsMessage.REQ_CRDT_STATE, async (data, sender) => {
+    // Requests arriving before role resolution are recovered by the requester's retry.
+    if (!isServerAtom.getOrNull()) return
```
gonpombo8, line 119: doesnt this check happen on the CommsMessage.RES_CRDT_STATE ?  if not it should happen there too !
if (sender === server_addresss

PR #1617, opened 2026-09-11
fix(sdk): refuse components the scene never synchronises

LautaroPetaccio:
## The problem

`NOT_SYNC_COMPONENTS` names components that describe one client's *own* situation rather than shared world state — UI, engine and realm info, tween and media state, another player's input and raycast results.

`shouldSyncComponent` is consulted in exactly two places, `engineToCrdt` and `syncFilter`, and both of them run when **sending**. Nothing checked it on the way in.

So a peer could push any of those as a network message and the server would apply it and relay it to the whole room. Verified with `UiTransform`, which is on the list:

```
UiTransform on NOT_SYNC list: true
(a) UiTransform applied on server: true | relayed to peers: true
```

## What it does

Runs the same check on the server's validation path — the one place peer state enters the authoritative engine.

Nothing legitimate is refused. Honest clients can't send these components in the first place, because the same list already stops them leaving, so this is the receiving half of a rule that was only ever half applied.

## Scope, measured rather than assumed

The grow-only variant is **not reachable on this branch**. A plain `APPEND_VALUE` returns no bytes from `processServerMessages` and is never relayed — only network message types are handled and there is no `APPEND_VALUE_NETWORK`:

```
(b) APPEND_VALUE bytes returned by the server: 0 | relayed: 0
```

That matters because the sharpest version of this bug is a grow-only one: on `main`, where the engine parses `APPEND_VALUE` straight off the transport, a peer can append to another client's `PointerEventsResult` and make their `inputSystem.isTriggered` return true — driving someone else's input. This branch's architecture already prevents that, so what's left here is the LWW half: injected UI and local-state confusion.

A refused message still costs an entity, since `findOrCreateNetworkEntity` runs before validation. That's the unbounded-pool issue and isn't changed here.

## How to test

```bash
node_modules/.bin/jest --testPathPatterns='inbound-component-allowlist'
```

Four cases. Two fail on `auth-server` — a peer's `UiTransform` is applied on the server and relayed to peers. The other two pin what must not change: `UiTransform` really is on the never-sync list, and a component the scene *does* synchronise is still relayed.

```bash
node_modules/.bin/jest --testPathPatterns='test/(ecs|sdk|react-ecs)/'   # 1091 pass
```

`Generated OnPointerDown ProtoBuf › should serialize/deserialize OnPointerUp` fails on a pristine `auth-server` checkout here too — the installed `@dcl/protocol` is newer than the committed artifacts. Unrelated, untouched.


LautaroPetaccio:
Fixed, and **retargeted onto #1612** rather than `auth-server`.

Both reviews asked for the same structural change — check before allocating — and it lands in the same function, so doing it twice would have meant resolving the same conflict twice. #1612 now carries the pre-allocation check; this PR adds the allowlist to it and shows only its own two-file delta. It should merge after #1612.

**P1 — fixed.** `shouldSyncComponent` now gates allocation as well as the write. Those are genuinely two places rather than a duplicate: the pre-check stops an entity being spent on a forbidden component for a pair the server has never seen, and the check in `validateMessagePermissions` still refuses the write when the pair is already mapped. Without the second one, a forbidden component aimed at an existing entity would be applied.

New test sends twenty-five never-sync updates naming unseen entities and asserts no mappings are created.

PR #1624, opened 2026-09-15
feat(ecs): expose renderer playback-position reports for audio

LautaroPetaccio:
## Summary

`audioEventsSystem` gains a way to read the playback-position reports renderers write into `AudioEvent` (see the protocol change below), and to resolve them against the scene's own clock:

- `registerAudioPlaybackEntity(entity, callback)` / `removeAudioPlaybackEntity(entity)`: the callback runs once per scene frame with the newest position report for that entity, already resolved against the scene clock at the tick the renderer sampled it in, as `{ report, sceneTime, offset }`. It is skipped on frames where no new position arrived. A renderer sampling faster than the scene ticks appends several reports per frame; the callback receives the freshest, which is the one to align against. Reports carrying no position never reach it; those are media-state changes, which `registerAudioEventsEntity` already delivers.
- `getSceneTimeAtTick(tickNumber)`: the scene clock recorded in a given tick, exposed on its own for scenes that handle raw reports, and for resolving `PBVideoEvent` the same way.
- `getAudioPlayback(entity)`: the latest report carrying `currentOffset`, or `undefined`.

`registerAudioEventsEntity` keeps its state-change semantics, so existing scenes receive no extra callbacks from the position reports.

## Why

Renderers start an `AudioSource` some milliseconds after being asked to (100 to 250 ms measured on the Unity explorer, varying per start) and `AudioSource.currentTime` is a write-only seek, so scenes that need to align gameplay with what is heard had no signal to work from.

A report has to be compared against the scene clock **at the tick it was sampled in**, not at the moment the callback runs, or the result is off by however long the report spent in transit. That per-tick history is the one piece a scene can get wrong, and every scene aligning anything to audio or video needs the same one, so it lives here rather than in each scene ([review thread](https://github.com/decentraland/adr/pull/324#discussion_r4038301378)).

## Accuracy, and what it does not cover

Resolving at the sampling tick removes the transport delay. It does not make the result exact, and the docstrings now say so rather than implying otherwise. Two terms remain:

- The renderer reads its playhead once per frame, from a counter that advances in audio-buffer steps, so the reading is quantised to the coarser of the two.
- The reported position is the decoder's read position, not the moment a sample leaves the speaker. The mixer buffer, the driver and the device add tens of milliseconds on top, consistently signed and roughly constant for a given machine and output device. No field carries it.

The sum behaves as a per-session constant, so a scene needing alignment finer than a tick can measure it once and subtract it. ADR-318 has the detail.

## A note on the tick lookup

A renderer publishes `EngineInfo.tick_number` and the reports sampled in that tick in the same batch, so the lookup normally resolves inside the frame that receives the report. It still has to tolerate a tick it has not recorded, and it must not drop a report when that happens: resolving after marking a report as seen would lose it permanently, and a renderer stamping a tick the scene never records would silence the callback entirely rather than degrade. It resolves first and falls back to the current scene clock.

## Dependencies

- Protocol: https://github.com/decentraland/protocol/pull/488 adds the optional `tick_number`, `current_offset` and `clip_length` fields.
- ADR-318: https://github.com/decentraland/adr/pull/324

## Protocol pin (temporary, blocks merge by design)

CI regenerates the ecs components from the installed `@dcl/protocol`, so the second commit pins both protocol dependencies to the package published from the protocol PR's branch. The **Check for development dependencies** job fails on that pin on purpose: it rejects any `https:` dependency so a branch build can never ship. That failure is expected and will clear once the pin is replaced. **Before merging, replace the pin with the released protocol version** (`make update-protocol`) once decentraland/protocol#488 lands. The API report was regenerated against that pin, which also brings in unrelated additions that landed on protocol `main` this week (`ExplorerItemPurchaseResult`, `request_id` on `ExplorerUiEventsResult`); the latter needed its serialization spec updated to the zero value. The bundle-size snapshots were refreshed because the generated `AudioEvent` grew.

## Both branches

The audio sources are identical on `main` and `auth-server`, and the feature commit cherry-picks cleanly onto `auth-server`, so the regular sync carries it over.

## Testing

`test/ecs/events/audioEventsSystem.spec.ts` covers 20 cases, including the resolved-sample delivery, a late report resolving against the clock at its own tick, and reports naming a tick the history never recorded, which must all still be delivered rather than skipped. `test/ecs/components/AudioEvent.spec.ts` covers a state-only report, a full position report and a position without clip length. The whole `test/ecs` suite passes at 808 cases, and build, lint, docs and CLI E2E are green against the pin.

## Related pull requests

- Test scene exercising the API against this branch build: https://github.com/decentraland/sdk7-test-scenes/pull/99
- Documentation: https://github.com/decentraland/documentation/pull/609
- SDK skills: https://github.com/decentraland/sdk-skills/pull/97


pravusjif (approved):
LGTM, remember to update the protocol package before merging 👍

